PR #41, opened 2014-10-15
Added example of conditional_operator not working in scan.

andersonwolfe:


mhart:
This must have been on an old version of dynalite that supported the previous DynamoDB API, before the conditionals API changed.

This was updated in dynalite v0.4.0 with https://github.com/mhart/dynalite/commit/0d21dc811ac83a146d89200f17f7090db0aa8942

@andersonwolfe what version of dynalite were you using?

Issue #56, opened 2015-04-06
Bug with Model init when having a hash or range key of 0


pior:
Given a Thread Project with a numerical key:

```python
class Project(Model):
    class Meta:
        table_name = DB_TABLE_PREFIX + 'Device'

    project_id = NumberAttribute(hash_key=True)
```
```python
In [58]: Project(0).project_id is None
Out[58]: True

In [59]: Project(project_id=0).project_id == 0
Out[59]: True
```
See https://github.com/jlafon/PynamoDB/blob/devel/pynamodb/models.py#L241

pior:
There seems to be the same issue on the deserialization side : 

```
In [70]: list(Project.scan())
Out[70]: [Project<1>, Project<null>]
```

kyangmove:
I have the same issue for the numeric range key.

PR #57, opened 2015-04-06
Fix bug with falsy values (other than None) #56

pior:
See #56


jlafon:
Great catch, thanks! The failing build is not because of this change, just FYI.

PR #68, opened 2015-07-14
range_key in get fails with truthy Falses

khwilson:
Previously, if used a .get(hash_key, range_key) where bool(range_key)
was False, the range_key wouldn't actually show up in the query to
DynamoDB because the get_identifier_map function would simply use
the truthy evaluation `if range_key`. By changing this to an
explicit `if range_key is not None`, this allows gets when a Python
value would otherwise evaluate to False, e.g., 0.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -136,7 +136,7 @@ class MetaTable(object):
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


jlafon:
Looks good. Thanks.

PR #75, opened 2015-09-20
remove OrderedDict references

joshowen:


joshowen:
I'll fix the tests later today.

pynamodb/tests/deep_eq.py
```diff
@@ -0,0 +1,140 @@
...
+#LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+#OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+#THE SOFTWARE.
+
+import datetime, time, functools, operator, six
+
+default_fudge = datetime.timedelta(seconds=0, microseconds=0, days=0)
+
+def deep_eq(_v1, _v2, datetime_fudge=default_fudge, _assert=False):
+  """
```
danielhochman: re-indent to 4 spaces?

pynamodb/tests/test_model.py
```diff
@@ -515,7 +524,7 @@ def test_delete(self):
                 'TableName': 'UserModel'
             }
             args = req.call_args[0][1]
-            self.assertEqual(args, params)
+            deep_eq(args, params, _assert=True)
```
danielhochman: a couple of readability suggestions:
- since it's not a library and we can modify the code, why not change ```_assert``` to default to ```True```?
- i would rename it ```assert_deep_eq``` just so it's more obvious to the test reader that it's where the assertion is happening. you could also do ```assert deep_eq(args, params)```. when the assertion passes it will pass and if it fails the inner assertion will still throw the more helpful message

danielhochman:
looks good to me @joshowen

pynamodb/tests/test_model.py
joshowen: We can, it was someone else's gist, so I tried to modify it as little as possible.  (I had to use six for python3 compatibility)

PR #76, opened 2015-09-22
fix for unprocessed items in put/delete requests

joshowen:
Put and Delete use different keys than BatchGet.


joshowen:
Put and Delete use different keys than BatchGet.

PR #78, opened 2015-09-23
Use ujson if its available

joshowen:
If you have it, might as well monkey patch botocore.vendored.requests to use it too, but thats out of scope.

https://github.com/kennethreitz/requests/issues/1595

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -2,13 +2,19 @@
 PynamoDB attributes
 """
 import six
-import json
 from base64 import b64encode, b64decode
 from delorean import Delorean, parse
 from pynamodb.constants import (
     STRING, NUMBER, BINARY, UTC, DATETIME_FORMAT, BINARY_SET, STRING_SET, NUMBER_SET,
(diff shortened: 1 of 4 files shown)


danielhochman:
https://github.com/jlafon/PynamoDB/blob/33eb91a40ee252525871fa1ce0e47cf6b002ad17/pynamodb/connection/base.py#L238

joshowen:
That is where the monkey patching mentioned above comes in.

danielhochman:
we don't need to patch though, why not call json.loads(request.text)?

danielhochman:
+1 from me

joshowen:
@jlafon any thoughts

joshowen:
Thoughts on merging this in?  I'm happy to update the branch.  We've been using this in production for a while.

danielhochman:
@joshowen definitely. let's take advantage of the new global config introduced in 1.6.0 and allow you to globally patch in whatever json parser you want, defaulting to the built-in json.

let me know if you have any questions.

PR #84, opened 2015-09-30
Fix binary attributes

joshowen:


joshowen:
@jlafon @danielhochman this fixes binary attributes and sets on the devel branch.

danielhochman:
are there tests we can add?

PR #88, opened 2015-11-01
Support Streams

brln:
Botocore v1.1.2 supports new parameters for adding a stream to a dynamo
table on creation. It's relatively straightforward to add an interface
for doing this on Pynamodb table creation, so add that.


brln:
@jlafon I'd appreciate any feedback on this, it would be useful to me if this feature were available.

jlafon:
Sorry for the slow response. This looks good, thanks!

PR #91, opened 2015-11-19
Configurable Session Class

joshowen:
This is very useful for overriding the size of the Connection Pool and other Session fields.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -1,6 +1,7 @@
 """
 Lowest level connection
 """
+from base64 import b64decode
 import logging
 
 import six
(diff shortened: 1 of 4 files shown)


pynamodb/connection/base.py
```diff
@@ -181,6 +181,11 @@ def __init__(self, region=None, host=None):
         else:
             self.region = DEFAULT_REGION
 
+        if session_cls:
+            self.session_cls = session_cls
+        else:
+            self.session_cls = requests.Session
```
danielhochman: Any reason not to default to this in the param?

danielhochman:
I was going to do this myself! +1 from me

pynamodb/connection/base.py
danielhochman: Ah I see, None can be passed

joshowen:
Any other optimizations you can think of to the session other than increasing the connection pool size?

danielhochman:
I'd have to think on that a bit more. We're actually standing up an egress proxy local to each application server to handle connections to external services. I'll let you know what we learn from that in a couple of weeks.

joshowen:
@jlafon any thoughts

joshowen:
@danielhochman hows that egress proxy working?

PR #97, opened 2016-02-01
Fix for binary hash keys

joshowen:


danielhochman:
i'm going to take #114 for now. it's less of a change and this function is unfortunately not covered by unit tests.

when i get a chance i'm going to write some tests and probably try and get rid of some of the code repetition that this approach affords us. will leave this PR open in case you want to do the same. feel free to close if you want.

danielhochman:
actually i'll close. if you want to take this approach with tests, you can reopen or create a new PR. trying to get number of open PRs down.

Issue #92, opened 2015-11-23
broken proxy support


kapilt:
surprising since boto3 works great out of the box, its unclear why pynamodb is doing a lot of the lower level connection level mgmt, instead of just using the default from boto3. the issue is the use of the requests.Session in Connection directly doesn't respect standard proxy environment variables.

kapilt:
its a quick hack to make this work for me for a demo, but it works okay

on Connection
```
    @property
    def requests_session(self):
        """                                                                                                                                                                                                                                                                                                                   
        Return a requests session to execute prepared requests using the same pool                                                                                                                                                                                                                                            
        """
        if self._requests_session is None:
            self._requests_session = requests.Session()
            if 'https_proxy' in os.environ:
                self._requests_session.proxies['https'] = os.environ['https_proxy']
            if 'http_proxy' in os.environ:
                self._requests_session.proxies['http'] = os.environ['http_proxy'] 
        return self._requests_session
```

jlafon:
This was done for performance reasons, as botocore's request parsing is very slow.

PR #100, opened 2016-02-25
Use appropriate HTTP proxies from botocore.

ab:
Since PynamoDB breaks abstraction barriers to use botocore's vendored
requests module directly, we ought to thread through the proxies that
exist in the botocore Endpoint. Otherwise we will not respect proxies
that are set by system environment variables or Boto configuration.


ab:
Also fixes: #92 

@jlafon What would it take to get this merged?

jlafon:
Thanks!

ab:
Thanks, @jlafon, do you have any expected timeline for when this might make it into a new release?

PR #107, opened 2016-03-26
Correctly handle expected valuation generation for set types.

### LINKED ISSUE #102: On update_item when using "IN" Conditional Write Operator, AttributeValueList gets only the last item of the list.
I am trying to do an update_item with the following parameters

`model.update_item("status", "delivered", status__in=['queued', 'sending', 'sent'])`
The model's current status is 'queued'. So this should work. However, I get a UpdateError. When I look at the logs, this is what the AWS json is generating.

`{'status': {'ComparisonOperator': 'IN', 'AttributeValueList': [{'S': u'sent'}]}}`

Shouldn't the AttirbuteValueList have all three items in it? It seems to be only putting the last item in the list.

jpinner-lyft:
Resolves #102


jlafon:
Thanks!

PR #108, opened 2016-03-26
Return the cause of connection exceptions to the caller.

jpinner-lyft:
Extract the DynamoDB error code from non-200 responses and return it to the caller when throwing Connection Exceptions. This provides a mechanism for the caller to modify their error behavior the error type (e.g. ConditionalCheckFailedException vs InternalServerError).


jlafon:
Thanks! Would you be willing to add a little documentation for this?

jpinner-lyft:
@jlafon added brief documentation at https://github.com/jlafon/PynamoDB/pull/108

PR #112, opened 2016-03-29
Document conditional operation failure check.

jpinner-lyft:
Relates to https://github.com/jlafon/PynamoDB/pull/108


jpinner-lyft:
@jlafon is this what you had in mind?

PR #114, opened 2016-04-17
Treat UnprocessedItems as dict of list of dicts

nickgravgaard:
Previously, when a BatchWriteItem call returned UnprocessedItems, Pynamo
incorrectly assumed that it was a dict of dict of dicts, when it was
actually a dict of list of dicts.


tomhillable:
Awesome, think this will resolve the issue I'm currently having - `AttributeError: 'list' object has no attribute 'values'`

joshowen:
This is fixed on https://github.com/jlafon/PynamoDB/pull/97/files as well.

danielhochman:
+1

Issue #95, opened 2016-01-19
Limit is off by one on Model.query


SpacedOutScott:
Passing limit into Model.query returns the incorrect # of results. Passing limit=1 returns 0, limit=2 returns 1 etc. 

I believe this to be because the function subtracts from the limit and then does a comparison to 0 before yielding. If the function did the operations in the opposite order, i believe it would work correctly (but i'm unsure of implications it has for other bits of functionality). 

As of this writing, this is the line in question: https://github.com/jlafon/PynamoDB/blob/devel/pynamodb/models.py#L571

hadrien:
I am facing the same issue and reason given by @SpacedOutScott seems valid to me.

PR #117, opened 2016-04-29
Make limit inclusive. Add query limit tests.

danielcorreia:
Related to issue [#95](https://github.com/jlafon/PynamoDB/issues/95), which is already closed but doesn't seem to work.

So, this returns the correct number of results. Example: when limit is 5 and there are at
least 5 matches return 5 items.

Hopefully I wrote the tests properly, not very experienced with testing. Please look at it.

Thanks!

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -564,11 +564,12 @@ class Model(with_metaclass(MetaModel)):
         last_evaluated_key = data.get(LAST_EVALUATED_KEY, None)
 
         for item in data.get(ITEMS):
-            if limit is not None:
-                if limit == 0:
-                    return
-                limit -= 1
+            if limit == 0:
+                return
             yield cls.from_raw_data(item)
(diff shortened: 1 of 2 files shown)


LiteWait:
This really seems to be the wrong fix, this works for me,

     for item in data.get(ITEMS):
            yield cls.from_raw_data(item)
            if limit is not None:
                limit -= 1
                if not limit:
                    return

danielcorreia:
@LiteWait both solutions work if the `limit` is > 0 or `None`.
With that iterator, if the user uses 0 as limit (which doesn't really make sense, I know) the iterator will yield all the items.

danielhochman:
We also need tests for limit logic when `last_evaluated_key` is returned.

danielhochman:
closed in favor of #123 which also fixed limits when there were multiple pages.

thanks for the contribution though!

Issue #103, opened 2016-03-04
BatchWriteItem: UnprocessedItems is not handled correctly


Sazpaimon:
When a BatchWriteItem call returns UnprocessedItems, Pynamo incorrectly assumes that it's a dict of dict of dicts, when it's actually a dict of list of dicts. The current behavior causes an error along the lines of "list has no attribute iteritems". I think changing `for item in six.itervalues(item_mapping):` to a simple `for item in item_mapping` (as is the case with UnprocessedKeys) would resolve the problem.

beeradb:
#120 was opened to fix this

danielhochman:
Fixed with #114

PR #120, opened 2016-05-03
[no-itervalues] Do not call itervalues on unprocessed items

beeradb:


beeradb:
This fixes #103

danielhochman:
dupe of #97 and #114

PR #121, opened 2016-05-04
retry on RequestExceptions only

### LINKED ISSUE #98: Pynamo's session handling doesn't handle ConnectionError failures
When a ConnectionError is through from botocore, the error is raised all the way through. Pynamo should catch the error and re-make the connection, then retry the request.

danielhochman:
Fixes #98. Should alleviate most of the problems users are having in production, though some may still be more comfortable with legacy botocore behavior of retrying on everything. At any rate that will take some more time to implement and come in another PR and should be highly configurable.

This PR does not retry based on status code or error code from DynamoDB. RequestExceptions can be retried quickly without need for exponential backoff and more granular handling so I have implemented it first.

I included private variables with the retry parameters in case someone wants to get at them, but it should  be considered an interface that won't be maintained and might break between minor releases.

TODO:
- [x] tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ pip-log.txt
 
 # Unit test / coverage reports
 .coverage
+cover/
 .tox
 nosetests.xml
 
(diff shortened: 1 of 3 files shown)


ryan-lane:
Looks good to me.

PR #123, opened 2016-05-06
fix limits

danielhochman:
copied query limit code from @joshowen's fork after a discussion with him over email.

i wrote the tests.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -574,9 +574,9 @@ class Model(with_metaclass(MetaModel)):
             log.debug("Fetching query page with exclusive start key: %s", last_evaluated_key)
             # If the user provided a limit, we need to subtract the number of results returned for each page
             if limit is not None:
-                limit -= data.get("Count", 0)
                 if limit == 0:
                     return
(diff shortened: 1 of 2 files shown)


jlafon:
LGTM. Thanks!

Lifto:
Still having an issue with this, maybe it's not the same thing?

Invalid range for parameter Limit, value: 0, valid range: 1-inf

for code like this:
items = Item.user_index.query(user_key, limit=2)

if limit is 1 I don't get an exception, if limit is 3 or more I don't get an exception.

danielhochman:
@Lifto are you on v1.5.0 or later?

Lifto:
tried on both 1.5.0 and 1.5.1

danielhochman:
@Lifto in the absence of a limit, how many items would that query return?

Lifto:
I would have to check in depth to 100% sure, but my basic expectation is that it would return 1 or 0 items.  When I wanted to get single items from a query I would have to request limit=2, because limit=1 would return nothing, but limit=2 would return the one item. (that was on 1.4.4)

PR #126, opened 2016-05-11
Fix handling of binary attr unprocessed

### LINKED ISSUE #125: unprocessed items in batch write may be double encoded
now that we fixed the parser in 1.5.0 for UNPROCESSED_ITEMS in #114, it appears we are double encoding in this retry: https://github.com/jlafon/PynamoDB/blob/da6aa1068d5a496ed25bcb2b0864b4c894ffc0ce/pynamodb/models.py#L120

probably we need to decode in the parser in addition to _convert_binary_attr

danielhochman:
fixes #125 

- moved binary attribute handling to its own method and added a test for this specific error case
- fixed the test for re-batch_writing unprocessed items since it wasn't doing anything before

worth noting that #97 #114 #120 all got it wrong, at least i think #97 did, the approach is a little different

I read the spec and reimplemented correctly http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html#API_BatchWriteItem_ResponseSyntax

### DIFF
--- a/pynamodb/tests/test_model.py
+++ b/pynamodb/tests/test_model.py
@@ -1803,14 +1803,6 @@ class ModelTestCase(TestCase):
                     batch.save(item)
 
     def test_batch_write_with_unprocessed(self):
-        items = []
-        for idx in range(10):
-            item = copy.copy(GET_MODEL_ITEM_DATA.get(ITEM))
-            item['user_id'] = {STRING_SHORT: 'id-{0}'.format(idx)}
-            item['email'] = {STRING_SHORT: 'email-{0}'.format(random.randint(0, 65536))}
-            item['picture'] = {BINARY_SHORT: BINARY_ATTR_DATA}
-            items.append(item)
-
         picture_blob = b'FFD8FFD8'
 
         items = []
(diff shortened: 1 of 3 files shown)


jlafon:
It would be nice if we didn't have to 'fix' binary attributes when returned as unprocessed items.

pynamodb/tests/test_model.py
```diff
@@ -1800,24 +1802,57 @@ def test_batch_write(self):
...
                     batch.save(item)
 
-        def fake_unprocessed_keys(*args, **kwargs):
-            if pythonic(REQUEST_ITEMS) in kwargs:
-                batch_items = kwargs.get(pythonic(REQUEST_ITEMS)).get(UserModel.Meta.table_name)[1:]
-                unprocessed = {
-                    UNPROCESSED_KEYS: {
-                        UserModel.Meta.table_name: batch_items
+    def test_batch_write_with_unprocessed(self):
+        items = []
```
ewalker11: Looks like you're re-assigning this on line 1816, delete section?

danielhochman:
binary attributes are double encoded by default unfortunately see #127

it turns out that binary attributes written by the unprocessed items flow were n+2 times encoded, where n is the number of times the item was returned as an unprocessed item from subsequent retries.

PR #129, opened 2016-05-13
Prevent double encoding of binary

### LINKED ISSUE #127: Binary Attributes Double-base64 Encoded
The botocore client, called via [`_convert_to_request_dict`](https://github.com/jlafon/PynamoDB/blob/devel/pynamodb/connection/base.py#L247), is performing another round of base64 encoding on binary fields. This is on top of the existing round of encoding done by the BinaryAttribute's [`serialize()`](https://github.com/jlafon/PynamoDB/blob/621b90aab64e14d274289e4e5684926f340bc48c/pynamodb/attributes.py#L100) method.

This is resulting in **double base64 encoded** values stored in Dynamo, which is why we have to undo one of the decodes via [`_convert_binary`](https://github.com/jlafon/PynamoDB/blob/master/pynamodb/connection/base.py#L956). 

IOW, all binary data that has ever been persisted using the BinaryAttribute in any version of Pynamo is sitting double-base64-encoded in Dynamo somewhere.

danielhochman:
Fixes #127 

Unfortunately, all Binary data is double-encoded that has been written by Pynamo. This PR prevents that from happening going forward, but we need to have a way to deserialize the double-encoded data so that we don't break everyone going forward.

TODO:
- [ ] figure out how to support legacy double-encoded data


ewalker11:
Looks good to me so far. Regarding the backwards-compatibility question, how about we roll the double base64 encoding/decoding into the existing Binary attributes, deprecate them, and create a new set of binary attributes (e.g. `attributesv2.BinaryAttribute`) that do the right thing. Then we can eventually remove the legacy binary attributes.

jmphilli:
can we update this and merge?

jmphilli:
also add a test? 😸

ikonst:
Superseded by #1112 (oh wow, a lot of numbers gone by...).

PR #130, opened 2016-05-17
Raise on multiple uses of attrs in query or scan conditions

danielhochman:
See comment in code for thorough explanation

### DIFF
--- a/pynamodb/tests/test_model.py
+++ b/pynamodb/tests/test_model.py
@@ -1227,6 +1227,9 @@ class ModelTestCase(TestCase):
         # you cannot query a range key with multiple conditions
         self.assertRaises(ValueError, lambda: list(UserModel.query('foo', user_id__gt='id-1', user_id__le='id-2')))
 
+        # you cannot query a non-primary key with multiple conditions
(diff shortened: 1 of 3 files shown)


pynamodb/tests/test_model.py
```diff
@@ -1224,17 +1224,8 @@ def test_query(self):
...
-                item = copy.copy(GET_MODEL_ITEM_DATA.get(ITEM))
-                item['user_id'] = {STRING_SHORT: 'id-{0}'.format(idx)}
-                items.append(item)
-            req.return_value = {'Items': items}
-            queried = []
-            for item in UserModel.query('foo', user_id__gt='id-1', user_id__le='id-2'):
-                queried.append(item._serialize())
-            self.assertTrue(len(queried) == len(items))
+        # you cannot query a range key with multiple conditions
+        self.assertRaises(ValueError, lambda: list(UserModel.query('foo', user_id__gt='id-1', user_id__le='id-2')))
```
ewalker11: Add corresponding tests for key conditions / scans?

ewalker11:
LGTM +1

ewalker11:
Maybe we could add a note in the documentation to use between instead of gt and lt?

Issue #132, opened 2016-06-05
update_item delete item from set fails


eukaryote:
Here is a simple test case that attempts to remove an value from a UnicodeSet attribute:

``` python
from pynamodb.models import Model
from pynamodb.attributes import NumberAttribute, UnicodeSetAttribute


class UserModel(Model):

    class Meta:
        table_name = 'User'
        region = 'us-west-2'
        read_capacity_units = 1
        write_capacity_units = 1
        host = 'http://localhost:8000'

    id = NumberAttribute(hash_key=True)
    nicknames = UnicodeSetAttribute(null=True)


UserModel.create_table(wait=True)
try:
    original_nicknames = {'name1', 'name2', 'name3'}
    nicknames_to_remove = {'name2'}
    expected_result_nicknames = original_nicknames - nicknames_to_remove
    user = UserModel(id=1, nicknames=original_nicknames)
    user.save()
    user = UserModel.get(1)
    print('original nicknames:', user.nicknames)
    assert original_nicknames == user.nicknames
    print('nicknames to remove:', nicknames_to_remove)
    user.update_item('nicknames', value=nicknames_to_remove, action='delete')
    user = UserModel.get(1)
    print('expected result nicknames:', expected_result_nicknames)
    print('  actual result nicknames:', user.nicknames)
    print()
    assert expected_result_nicknames == user.nicknames
finally:
    UserModel.delete_table()
```

When running this test case, the output is:

```
original nicknames: {'name2', 'name3', 'name1'}
nicknames to remove: {'name2'}
expected result nicknames: {'name3', 'name1'}
  actual result nicknames: None

Traceback (most recent call last):
  File "mytest.py", line 34, in <module>
    assert expected_result_nicknames == user.nicknames
AssertionError
```

I expected 'name2' to be removed from the `{'name1', 'name2', 'name3'}` set, but instead all the values were removed.

This bug exists since 1.4.4 (1.4.3 works fine), and is still present in 1.6.0. The associated pull request has been updated to be be based on the post 1.6.0 devel branch, and it includes a unit test that fails with only the 'pynamodb/tests' commits applied but succeeds with the changes to connection/base.py and models.py (for update_item of connection and model, respectively).

eukaryote:
See pull request #133 for a tentative fix that may still need some work.

eukaryote:
A similar problem also occurs with the new update method, and the pull request has been updated to add a test for 'update' and to apply a similar fix for that case.

[...]

PR #133, opened 2016-06-05
Starting point for issue #132: update_item@action='delete' with value

eukaryote:
Changes to update_item on both the model and connection, as well as a unit test, to restore the correct behavior of allowing deletion of items from a set while still preserving the ability to delete without having to specify a value when the entire set should be removed.

This fixes a bug since 1.4.4 where if you had a user with a set-valued attribute like:

```
user = UserModel(1, nicknames={'nick1',  'nick2', 'DELETEME'})
user.save()
```

Then if you tried to remove a nickname from the set, with:

```
user.update_item('nicknames', value=['DELETEME'], action='delete')
```

The query sent to DynamoDB is as if no value were provided, and DynamoDB removes the nicknames attribute rather than just removing the item from the set.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -360,7 +360,7 @@ class Model(AttributeContainer):
                 ACTION: action.upper() if action else None,
             }
         }
-        if action is not None and action.upper() != DELETE:
+        if action is not None and (action.upper() != DELETE or (value is not None and not(isinstance(value, str) and value.upper() == 'NULL'))):
             kwargs[pythonic(ATTR_UPDATES)][attribute_cls.attr_name][VALUE] = {ATTR_TYPE_MAP[attribute_cls.attr_type]: value}
         kwargs[pythonic(RETURN_VALUES)] = ALL_NEW
         kwargs.update(conditional_operator=conditional_operator)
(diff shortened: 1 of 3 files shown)


eukaryote:
@jlafon This is an old pull request for #133 that has been updated and improved. Could you please review the bug report and pull request and let me know if there is anything else I can do to help move this along?

garrettheel:

pynamodb/connection/base.py
```diff
@@ -783,14 +783,14 @@ def update_item(self,
             value = update.get(VALUE)
             attr_type, value = self.parse_attribute(value, return_type=True)
             action = update.get(ACTION)
-            if attr_type is None and action is not None and action.upper() != DELETE:
+            if attr_type is None and action is not None and (action.upper() != DELETE or (value is not None and not(isinstance(value, str) and value.upper() == 'NULL'))):
```
behos: Can you extract this condition to a function? It's being repeated in multiple places.

eukaryote:
I've updated the changes related to update_item to work with the new expression syntax, and added a test for the new update method and made a similar fix for that case too.

Please review my changes carefully, as I'm not familiar enough with the code base to be sure my changes haven't broken something else that wasn't currently being tested in the test suite.

behos:
During my testing around this it seemed to me like the check done in my similar PR wasn't causing any regressions, and yours is much more conservative than that one :)

https://github.com/pynamodb/PynamoDB/pull/345/files

We've been running our own patched version of this code with no issues so far.

jl-DaDar:
i am not sure why but this does not work for me. I am on 3.0.1

jpinner-lyft:
Closing in favor of #374

PR #135, opened 2016-06-23
Retry on InternalServerError exceptions with backoff

danielhochman:
getting a little more retry logic in

want to do more around configurability, but will come later


ewalker11:
+1

PR #139, opened 2016-07-12
The limit for query and scan is modified incorrectly based on the items

anandswaminathan:
Today - The limit variable input is being used, as keep getting items until you have retrieved a value equal to X. 

This variable when passed to DynamoDB is interpreted as page size.

The bug involves subtracting item count from the Limit, twice, instead of once.

This PR separates the usage of limit from page size.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -510,6 +510,7 @@ class Model(with_metaclass(MetaModel)):
               limit=None,
               last_evaluated_key=None,
               attributes_to_get=None,
+              total_items_to_return=None,
               **filters):
         """
         Provides a high level query API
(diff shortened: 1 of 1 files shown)


anandswaminathan:
@danielhochman 

Note that, I am passing in the Limit as page_size, if the page_size is None. If not, all the callers which do not have page_size input (currently), will end up doing 1 MB scan/query.

pynamodb/models.py
```diff
@@ -572,13 +578,9 @@ def query(cls,
 
         while last_evaluated_key:
             log.debug("Fetching query page with exclusive start key: %s", last_evaluated_key)
```
danielhochman: this log message is a dupe of :584, can you delete it?

danielhochman:
+1 pending comment

pynamodb/models.py
anandswaminathan: Done.

PR #140, opened 2016-07-12
Fixing incorrect manipulation of limit value during pagination

anandswaminathan:
Today - The limit variable input is being used, as keep getting items until you have retrieved a value equal to X. 

This variable when passed to DynamoDB is interpreted as page size.

The bug involves subtracting item count from the Limit, twice, instead of once.

This PR separates does not fix the incorrect use of Limit, but merely fixes the bug where value of limit is decremented twice.


anandswaminathan:
DISCARDED

PR #144, opened 2016-07-26
Deserialize hash key while loading from json

krishna-lyft:
Fixes issue # 143

While trying to load a table from a json dump, https://github.com/jlafon/PynamoDB/blob/devel/pynamodb/models.py#L764 does not deserialize the hash key. Hence, we run into an exception if the hash key is anything other than a UnicodeAttribute.


jmphilli:
hey -- i think you're probably trying to deserialize something like a json attribute or binary one right? 

can you add a test or example of what you're trying to deserialize? i think i see what's going on but having a concrete example would help :)

PR #146, opened 2016-07-29
Add the ability for pynamo to read + write map and list attributes

jmphilli:
[Context](https://github.com/jlafon/PynamoDB/issues/38)
This allows pynamo to serialize + deserialize map and list types.


pynamodb/constants.py
```diff
@@ -81,13 +81,17 @@
...
 STRING = 'String'
 STRING_SET = 'StringSet'
 NUMBER = 'Number'
 NUMBER_SET = 'NumberSet'
 BINARY = 'Binary'
 BINARY_SET = 'BinarySet'
-SHORT_ATTR_TYPES = [STRING_SHORT, STRING_SET_SHORT, NUMBER_SHORT, NUMBER_SET_SHORT, BINARY_SHORT, BINARY_SET_SHORT]
+MAP = 'Map'
+LIST = 'List'
+SHORT_ATTR_TYPES = [STRING_SHORT, STRING_SET_SHORT, NUMBER_SHORT, NUMBER_SET_SHORT, BINARY_SHORT, BINARY_SET_SHORT, MAP_SHORT]
```
ld-git: LIST_SHORT missing here

pynamodb/attributes.py
```diff
@@ -257,3 +257,139 @@ def deserialize(self, value):
...
+            elif 'BOOL' in v:
+                rval[k] = bool(v['BOOL'])
+            elif 'M' in v:
+                rval[k] = MapAttribute().deserialize(v['M'])
+            else:
+                raise Exception('Unknown value: ' + str(v))
+        return rval
+
+
+class ListAttribute(Attribute):
```
ld-git: for list attributes that are composed of a list of maps (like timeWindows). how do we specify the model representation of the member maps?

pynamodb/attributes.py
```diff
@@ -257,3 +257,128 @@ def deserialize(self, value):
...
+    attr_type = LIST
+
+    def serialize(self, values):
+        """
+        Encode the given list of objects into a list of AttributeValue types.
+        """
+        rval = []
+        for v in values:
+            if v is None:
```
garrettheel: I'm sensing a pattern with all of these.. surely we can write this in a generic way such that we don't need to do the type checks every time?

pynamodb/attributes.py
```diff
@@ -257,3 +257,128 @@ def deserialize(self, value):
...
+                if v:
+                    value_to_dump = bool(1)
+
+                rval[k] = {'BOOL': value_to_dump}
+            else:
+                raise Exception('Unknown value: ' + str(v))
+
+        return rval
+
+    def deserialize(self, values):
```
ld-git: We need to check None here. When reading from records written from other sources a None may be passed in here.

pynamodb/attributes.py
ld-git: When a list is a list of maps, we can also let the list constructor take a model, and use that model to validate each member map.

pynamodb/attributes.py
```diff
@@ -257,3 +258,181 @@ def deserialize(self, value):
         Takes a UTC datetime string and returns a datetime object
         """
         return parse(value, dayfirst=False).datetime
+
+
+class MapAttribute(Attribute):
```
ld-git: so when I want to define an MapAttribute that requires verification, I need to create a subclass, otherwise I can use MapAttribute directly, right?

jmphilli: bingo. for our user case we would just want to do
```python
class CouponTemplate(MapAttribute):
    code = UnicodeAttribute()
    ...
```

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+    def _set_attributes(self, **attrs):
+        """
+        Sets the attributes for this object
+        """
+        for attr_name, attr in self._get_attributes().aliased_attrs():
+            if attr.attr_name in attrs:
+                if type(attrs.get(attr.attr_name)) is type(attr) or not isinstance(attrs.get(attr_name), collections.Mapping):
+                    setattr(self, attr_name, attrs.get(attr.attr_name))
+                else:
+                    thing = attrs.get(attr_name)
```
danielhochman: thing?

pynamodb/__init__.py
```diff
@@ -7,4 +7,4 @@
 """
 __author__ = 'Jharrod LaFon'
 __license__ = 'MIT'
-__version__ = '1.5.3'
+__version__ = '1.6.0'
```
johnliu: probably 2.0.0 haha.. pretty major

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+        """
+        Encode the given list of objects into a list of AttributeValue types.
+        """
+        rval = []
+        for v in values:
+            if v is None:
+                rval += [{'NULL': bool(1)}]
+            elif type(v) is dict:
+                rval += [{'M': MapAttribute().serialize(v)}]
+            elif type(v) is list or type(v) is set:  # no set embeded in List
```
danielhochman: spelling

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+
+    def serialize(self, values):
+        """
+        Encode the given list of objects into a list of AttributeValue types.
+        """
+        rval = []
+        for v in values:
+            if v is None:
+                rval += [{'NULL': bool(1)}]
+            elif type(v) is dict:
```
danielhochman: use isinstance instead of type is?

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+                rval += [{'M': MapAttribute().serialize(v)}]
```
danielhochman: use .append()

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
         return parse(value, dayfirst=False).datetime
+
+
+class MapAttributeMeta(type):
+    def __init__(cls, name, bases, attrs):
+        setattr(cls, '_attributes', None)
+
+
+class MapAttribute(with_metaclass(MapAttributeMeta, Attribute)):
+    _attributes = None
```
garrettheel: I think the metaclass will do this for you, right?

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+
+    def __init__(self, **attrs):
+
+        hash_key = attrs.get('hash_key', False)
```
garrettheel: Is it actually valid to have a `Map` as a hash or range key? Can't find anything in the docs explicitly excluding it but it seems slightly weird (even if it's possible).

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
         Takes a UTC datetime string and returns a datetime object
         """
         return parse(value, dayfirst=False).datetime
+
+
+class MapAttributeMeta(type):
+    def __init__(cls, name, bases, attrs):
```
johnliu: this should probably call super

pynamodb/attributes.py
```diff
@@ -35,8 +38,9 @@ def __init__(self,
             self.attr_name = attr_name
 
     def __set__(self, instance, value):
-        if isinstance(value, Attribute):
-            return self
+        # we may want to override in the child class
+        #if isinstance(value, Attribute):
+        #    return self
```
johnliu: seems like this used to be chainable, should this still return self?

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+        """
+        for attr_name, attr in self._get_attributes().aliased_attrs():
+            if attr.attr_name in attrs:
+                if type(attrs.get(attr.attr_name)) is type(attr) or not isinstance(attrs.get(attr_name), collections.Mapping):
+                    setattr(self, attr_name, attrs.get(attr.attr_name))
+                else:
+                    thing = attrs.get(attr_name)
+                    instance = type(attr)()
+                    aliased_attrs = type(attr)._get_attributes().aliased_attrs()
+                    for bad_k,v in thing.iteritems():
```
garrettheel: This for block (along with `bad_k` and `good_k`) is pretty confusing. Can you explain with comments what's going on and use different variable names?

pynamodb/attributes.py
johnliu: nevermind `__set__` should not return anything

pynamodb/attributes.py
```diff
@@ -35,8 +38,9 @@ def __init__(self,
             self.attr_name = attr_name
 
     def __set__(self, instance, value):
-        if isinstance(value, Attribute):
-            return self
+        # we may want to override in the child class
```
garrettheel: Remove comment

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+class MapAttribute(with_metaclass(MapAttributeMeta, Attribute)):
+    _attributes = None
+
+    def __init__(self, **attrs):
+
+        hash_key = attrs.get('hash_key', False)
+        range_key = attrs.get('range_key', False)
+        null = attrs.get('null', None)
+        default = attrs.get('default', None)
+        attr_name = attrs.get('attr_name', None)
```
johnliu: why not make the required attributes explict?

```python
def __init__(self, hash_key=False, range_key=False, null=None, default=None, attr_name=None, **attrs):
    super(MapAttribute, self).__init(...)
    ...
```

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+    @classmethod
+    def _get_attributes(cls):
+        """
+        Returns the list of attributes for this class
+        """
+        if cls._attributes is None:
+            cls._attributes = AttributeDict()
+            for item in dir(cls):
+                try:
+                    item_cls = getattr(getattr(cls, item), "__class__", None)
```
johnliu: when would this raise `AttributeError`? you're passing a default to the outer `getattr`, and the inner one is based on `dir(cls)`

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+                except AttributeError:
+                    continue
+                if item_cls is None:
+                    continue
+                if issubclass(item_cls, (Attribute,)):
```
johnliu: instead of getting the __class__, could you not just check if `isinstance(item, Attribute)` ?

pynamodb/attributes.py
johnliu: actually I'm a bit confused why this metaclass is necessary at all, can't we just set `_attributes` directly on MapAttribute as it's done so right now?

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+        return self.attribute_values[item]
+
+    @classmethod
+    def _get_attributes(cls):
+        """
+        Returns the list of attributes for this class
+        """
+        if cls._attributes is None:
+            cls._attributes = AttributeDict()
+            for item in dir(cls):
```
johnliu: It might be a good idea to ignore anything that starts with `__`. This will iterate through all of the internals of the object as well..

```python
>>> class A(object):
...     pass

>>> dirs(A)
['__class__',
 '__delattr__',
 '__dict__',
 '__doc__',
 '__format__',
 '__getattribute__',
 '__hash__',
 '__init__',
 '__module__',
 '__new__',
 '__reduce__',
 '__reduce_ex__',
 '__repr__',
 '__setattr__',
 '__sizeof__',
 '__str__',
 '__subclasshook__',
 '__weakref__']
```

or perhaps even better to just filter for subclasses of Attribute directly:

```python
for attribute in filter(lambda attr: isinstance(getattr(cls, attr), Attribute), dir(cls)):
    ...
```

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+                try:
+                    item_cls = getattr(getattr(cls, item), "__class__", None)
+                except AttributeError:
+                    continue
+                if item_cls is None:
+                    continue
+                if issubclass(item_cls, (Attribute,)):
+                    instance = getattr(cls, item)
+                    if instance.attr_name is None:
+                        instance.attr_name = item
```
johnliu: So the purpose of this is basically telling the attribute what it's named on the class?

pynamodb/attributes.py
johnliu: Ah I see.. you need this so that each subclass sets it's own _attribute

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+                    cls._attributes[item] = instance
```
johnliu: so is the key for `_attributes` ever different from `attr_name`?

pynamodb/attribute_dict.py
```diff
@@ -0,0 +1,33 @@
+import collections
+
+
+class AttributeDict(collections.MutableMapping):
+    """
+    A dictionary that stores attributes by two keys
+    """
+    def __init__(self, *args, **kwargs):
+        self._values = {}
+        self._alt_values = {}
```
johnliu: the names here are a bit unclear, one is basically for the key used by dynamo, the other is the key used by the object right?

I think it's a bit more clear if this was explicit instead why not just use two dictionaries? If the keys are the same you can store the same on both

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+        return cls._attributes
+
+    def _set_attributes(self, **attrs):
+        """
+        Sets the attributes for this object
+        """
+        for attr_name, attr in self._get_attributes().aliased_attrs():
```
johnliu: would it make more sense to iterate of attrs instead?

pynamodb/attributes.py
johnliu: edit 2: don't think you actually need to pass these down to _set_attributes so probably better to be explicit.

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+                    aliased_attrs = type(attr)._get_attributes().aliased_attrs()
+                    for bad_k,v in thing.iteritems():
+                        k = None
+                        for good_k, aliased_v in aliased_attrs:
+                            if aliased_v.attr_name == bad_k:
+                                k = good_k
+                        setattr(instance, k, v)
+                    setattr(self, attr_name, instance)
+
+            elif attr_name in attrs:
```
johnliu: why doesn't the above apply here?

pynamodb/attributes.py
```diff
@@ -257,3 +261,223 @@ def deserialize(self, value):
...
+        return getattr(self, key) and type(getattr(self, key)) is not type(
+            value)
+
+    def validate(self):
+        for key, value in self._get_attributes().iteritems():
+            if not self.is_type_safe(key, value):
+                return False
+        return True
+
+    attr_type = MAP
```
johnliu: put this up at the top

pynamodb/tests/test_attributes.py
```diff
@@ -452,3 +453,146 @@ def test_control_chars(self):
...
+                'name': 'Mandy',
+                'age': 29,
+                'height': 157.48
+            },
+        ]
+        serialized = ListAttribute().serialize(family_attributes)
+        self.assertEqual(ListAttribute().deserialize(serialized), family_attributes)
+
+    def test_list_of_map_of_list(self):
+        multiple_store_grocey_list = [
```
johnliu: grocery*

pynamodb/attributes.py
jmphilli: i don't think it's valid to have a `Map` as hash / range

garrettheel: Looks like it's not valid to use `Map` or `List` as hash or range, so let's take these out of here

pynamodb/attributes.py
```diff
@@ -325,3 +326,239 @@ def deserialize(self, value):
...
+        rval = dict()
+        for k in values:
+            v = values[k]
+            attr_class = _get_class_for_deserialize(v)
+            attr_value = _get_value_for_deserialize(v)
+            rval[k] = attr_class.deserialize(attr_value)
+        return rval
+
+
+def _get_value_for_deserialize(value):
```
jmphilli: is there a better way to write these functions? i want to return something different but the conditions are the same

pynamodb/attribute_dict.py
```diff
@@ -0,0 +1,33 @@
...
+    def __setitem__(self, key, value):
+        if value.attr_name is not None:
+            self._values[value.attr_name] = value
+        self._alt_values[key] = value
+
+    def __delitem__(self, key):
+        del self._values[key]
+
+    def __iter__(self):
+        return iter(self._alt_values)
```
garrettheel: Agree with @johnliu that it's somewhat confusing to have `_alt_values` and `_values`. In cases like line, it seems like we might be missing out on values stored in the other dict (`_values`).

Do we strictly need the second dict? I understand that you've probably done it that way to be consistent with how PynamoDB was originally written

pynamodb/attributes.py
```diff
@@ -253,6 +254,17 @@ def serialize(self, value):
     def deserialize(self, value):
         return bool(json.loads(value))
 
+    def get_value(self, value):
```
garrettheel: Why is this logic on the generic `Attribute`? I would've expected it to be on the more specific Boolean subclasses

pynamodb/attributes.py
```diff
@@ -325,3 +337,234 @@ def deserialize(self, value):
...
+
+    def __getitem__(self, item):
+        return self.attribute_values[item]
+
+    @classmethod
+    def _get_attributes(cls):
+        """
+        Returns the list of attributes for this class
+        """
+        if cls._attributes is None:
```
garrettheel: If this block is the same as what a `Model` does can we extract it to a generic function which just populates an `AttributeDict` given a class?

pynamodb/attributes.py
```diff
@@ -325,3 +337,234 @@ def deserialize(self, value):
...
+                cls._attributes[item] = instance
+        return cls._attributes
+
+    def _set_attributes(self, **attrs):
+        """
+        Sets the attributes for this object
+        """
+        for attr_name, attr in self._get_attributes().aliased_attrs():
+            if attr.attr_name in attrs:
+                if type(attrs.get(attr.attr_name)) is type(attr) or not isinstance(attrs.get(attr_name), collections.Mapping):
```
garrettheel: This function is probably the most complex and hardest to grasp. It would really appreciate some <3 with variable names and comments to show your intent

E.g what does `not isinstance(attrs.get(attr_name), collections.Mapping)` mean in terms of how we figure out where to set the value?

pynamodb/attributes.py
```diff
@@ -325,3 +337,234 @@ def deserialize(self, value):
...
+
+    def is_type_safe(self, key, value):
+        can_be_null = value.null
+        if can_be_null and getattr(self, key) is None:
+            return True
+        return getattr(self, key) and type(getattr(self, key)) is not type(
+            value)
+
+    def validate(self):
+        for key, value in self._get_attributes().iteritems():
```
garrettheel: `return all(self.is_type_safe(k, v) for k, v in self._get_attributes())`

pynamodb/attributes.py
```diff
@@ -325,3 +337,234 @@ def deserialize(self, value):
...
+            if not self.is_type_safe(key, value):
+                return False
+        return True
+
+    def serialize(self, values):
+        """
+        Encode the given list of numbers into a list of AttributeValue types.
```
garrettheel: List of numbers?

pynamodb/attributes.py
```diff
@@ -325,3 +337,234 @@ def deserialize(self, value):
...
+    LIST_SHORT: ListAttribute(),
+    NUMBER_SHORT: NumberAttribute(),
+    STRING_SHORT: UnicodeAttribute(),
+    BOOLEAN: BooleanAttribute(),
+    MAP_SHORT: MapAttribute(),
+}
+
+SERIALIZE_CLASS_MAP = {
+    'NoneType': None,
+    'dict': MapAttribute(),
```
garrettheel: Why can't we use the actual types as keys here?

pynamodb/attributes.py
jmphilli: it's on the `LegacyBooleanAttribute` class

pynamodb/attribute_dict.py
```diff
@@ -0,0 +1,33 @@
+import collections
+
+
+class AttributeDict(collections.MutableMapping):
```
jmphilli: i just pulled this class out, not sure how much refactoring of it i should really be doing

pynamodb/attribute_dict.py
jmphilli: yeeeeaaaahhhhh not super stoked on redoing all of this code, but if it's a requirement i can figure it out

pynamodb/attributes.py
jmphilli: it's almost the same, a little different

pynamodb/attributes.py
jmphilli: added a comment below (line 406)

pynamodb/attribute_dict.py
jmphilli: this has to do with the fact that pynamo lets you pass in an `attr_name` ... the name in the db might be different than the name in the code (e.g. couponCode in db, coupon_code in py code) 

garrett also suggested that i figure out a way to mask this complexity from other parts of the code. i think if i do that though it will incur an O(n) search through the keys when i need to look for the one that i don't index off of, but i may be wrong

pynamodb/attributes.py
```diff
@@ -325,3 +337,233 @@ def deserialize(self, value):
...
+        return self.attribute_values[item]
+
+    @classmethod
+    def _get_attributes(cls):
+        """
+        Returns the list of attributes for this class
+        """
+        if cls._attributes is None:
+            cls._attributes = AttributeDict()
+            for item in [attr for attr in dir(cls) if
```
danielhochman: hmm, dir shouldn't really be used except as a help function. can you use getattr? and rewrite this longform? mixing list comps and for loops is confusing

pynamodb/attributes.py
danielhochman: why use isinstance instead of type?

danielhochman:
in general you should avoid using type() as much as possible: http://docs.quantifiedcode.com/python-anti-patterns/readability/do_not_compare_types_use_isinstance.html

is there any reason you chose to use `type()` so extensively? sometimes you even check issubclass(type) which is the exact reason why type is not preferred over isinstance.

pynamodb/attributes.py
jmphilli: I changed it into a `vars` call (is that any better??) and broke it out into it's own line

pynamodb/attributes.py
```diff
@@ -325,3 +337,224 @@ def deserialize(self, value):
...
+        can_be_null = value.null
+        if can_be_null and getattr(self, key) is None:
+            return True
+        return getattr(self, key) and type(getattr(self, key)) is not type(value)
+
+    def validate(self):
+        return all(self.is_type_safe(k, v) for k, v in six.iteritems(self._get_attributes()))
+
+    def serialize(self, values):
+        rval = dict()
```
danielhochman: nit: consistency use literal {}

pynamodb/attribute_dict.py
jlafon: My two cents: I'd rather see a single dictionary too, but it's not a requirement from my perspective.

jlafon:
LGTM!

jmphilli:
Merged into `lyft` going to do an RC and then merge into `jlafon` from there

PR #147, opened 2016-08-08
Using Retries using headers set in settings file rather than manual retries

anandswaminathan:
The default headers point to envoy. 
Removing the earlier retries at PynamoDB layer.

The settings can be overridden by setting the environment variable.
Default max retries is 3. Will retry on 5xx, and connect-failure.

@danielhochman  @mattklein123


pynamodb/connection/base.py
```diff
@@ -178,9 +178,10 @@ class Connection(object):
     A higher level abstraction over botocore
     """
 
-    def __init__(self, region=None, host=None, session_cls=None):
+    def __init__(self, region=None, host=None, session_cls=None, retry_enabled=False):
         self._tables = {}
         self.host = host
+        self.retry_enabled = retry_enabled
```
danielhochman: we should allow the user to configure the parameters of retry, not enable/disable

to disable the user can pass max_attempts = 1

pynamodb/models.py
```diff
@@ -17,6 +17,7 @@
 from pynamodb.types import HASH, RANGE
 from pynamodb.compat import NullHandler
 from pynamodb.indexes import Index, GlobalSecondaryIndex
+from pynamodb.settings import RequestSessionWithHeaders, RETRY_ENABLED
```
danielhochman: i'm not sure what's happening here

to recap, the user should be able to put a settings file in a known location on the system, e.g. `/etc/pynamodb/settings.py`

we need to check for existence of that file. if it exists, we load it. it should have the same parameters as the Meta class. if the parameter exists in the config file and not in the Meta class, then we use the parameter from the config file as the default. if the parameter exists in the Meta class, and the config file, the value in the Meta class takes precedence over the value in the config file.

pynamodb/connection/base.py
```diff
@@ -178,7 +179,8 @@ class Connection(object):
     A higher level abstraction over botocore
     """
 
-    def __init__(self, region=None, host=None, session_cls=None):
+    def __init__(self, region=None, host=None, session_cls=None,
+                 request_timeout_seconds=None, max_retry_attempts=None, base_backoff_ms=None):
```
danielhochman: let's rename to max_attempts since i think this variable encompasses the initial attempt as well

anandswaminathan: Actually I have implemented it as max_retry_attempts.

Let's say you pass 
max_retry_attempts = 0. There will be no retries.
max_retry_attempts = 1. There will be one retry. So totally two calls.

pynamodb/settings.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+
+log = logging.getLogger(__name__)
+
+OVERRIDE_SETTINGS_PATH = getenv('PYNAMO_CONFIG', '/etc/pynamodb/settings_override.py')
+
+settings = {}
+
+DEFAULT_REQUEST_SESSION_HEADERS = dict()
+DEFAULT_REQUEST_SESSION_HEADERS['x-envoy-retry-on'] = '5xx,connect-failure'
```
danielhochman: we should not be upstreaming these custom headers

pynamodb/settings.py
```diff
@@ -0,0 +1,52 @@
...
+        request_Session_headers = settings['request_session_headers']
+        for header_key in request_Session_headers:
+            self.headers.update({header_key: request_Session_headers[header_key]})
+
+override_setting_dict = {}
+if os.path.isfile(OVERRIDE_SETTINGS_PATH):
+    log.info('Override settings for pynamo available {0}'.format(OVERRIDE_SETTINGS_PATH))
+    try:
+        with open(OVERRIDE_SETTINGS_PATH, 'r') as f:
+            exec(f.read(), override_setting_dict)
```
danielhochman: security alert!

try http://stackoverflow.com/a/67692 instead

pynamodb/settings.py
```diff
@@ -0,0 +1,52 @@
...
+    if key in override_setting_dict:
+        return override_setting_dict[key]
+
+    return default_settings_dict[key]
+
+settings['request_timeout_seconds'] = get_settings_value('REQUEST_TIMEOUT_SECONDS')
+settings['max_retry_attempts'] = get_settings_value('MAX_RETRY_ATTEMPTS')
+settings['base_backoff_ms'] = get_settings_value('BASE_BACKOFF_MS')
+settings['request_session_headers'] = get_settings_value('REQUEST_SESSION_HEADERS')
+settings['request_session_cls'] = RequestSessionWithHeaders
```
danielhochman: variable name in override should exactly match that of meta class

anyways, we should not need to enumerate the Meta class variables here

config merging could be handled in the Meta class itself so there is a single source of truth for configuration handling and field enumeration

pynamodb/models.py
```diff
@@ -141,7 +142,6 @@ class DefaultMeta(object):
     region = DEFAULT_REGION
     host = None
 
-
```
danielhochman: do not delete these lines, PEP8 is 2 blank lines between classes https://www.python.org/dev/peps/pep-0008/#blank-lines

pynamodb/settings.py
```diff
@@ -0,0 +1,41 @@
...
+
+DEFAULT_REQUEST_SESSION_HEADERS = dict()
+
+default_settings_dict = {}
+default_settings_dict['REQUEST_TIMEOUT_SECONDS'] = 25
+default_settings_dict['MAX_RETRY_ATTEMPTS'] = 3
+default_settings_dict['BASE_BACKOFF_MS'] = 25
+default_settings_dict['REQUEST_SESSION_HEADERS'] = DEFAULT_REQUEST_SESSION_HEADERS
+
+class RequestSessionWithHeaders(requests.Session):
```
danielhochman: we shouldn't need this customization in this file. we will ship that in our infrastructure only.

for example there is no need to have a settings value of REQUEST_SESSION_HEADERS. that we can pass a custom session_cls using the override file which has the headers that we want.

the responsibility of this file should be solely to load overrides

pynamodb/settings.py
```diff
@@ -0,0 +1,41 @@
...
+log = logging.getLogger(__name__)
+
+OVERRIDE_SETTINGS_PATH = getenv('PYNAMO_CONFIG', '/etc/pynamodb/settings_override.py')
+
+settings = {}
+
+DEFAULT_REQUEST_SESSION_HEADERS = dict()
+
+default_settings_dict = {}
+default_settings_dict['REQUEST_TIMEOUT_SECONDS'] = 25
```
danielhochman: i'd still really like to see all defaults in one place. currently they are in the Meta, the constructor, and here. also all should be overrideable from the settings file or Meta

pynamodb/connection/base.py
mikramulhaq: should we not initialize those with default values as below?
DEFAULT_TIMEOUT, DEFAULT_MAX_RETRY_ATTEMPTS_EXCEPTION, DEFAULT_BASE_BACKOFF_MS instead of None.

pynamodb/connection/base.py
```diff
@@ -254,8 +255,9 @@ def _make_api_call(self, operation_name, operation_kwargs):
         )
         prepared_request = self.client._endpoint.create_request(request_dict, operation_model)
 
-        for attempt_number in range(1, self._max_retry_attempts_exception + 1):
-            is_last_attempt_for_exceptions = attempt_number == self._max_retry_attempts_exception
+        for i in range(0, self._max_retry_attempts_exception + 1):
+            attempt_number = i + 1
```
mikramulhaq: nit: probably you can get rid of attempt_number completely?

anandswaminathan: The attempt_number is used for logging at multiple places. If I remove it, I would have to do i+1, everywhere. Not so clean/manageable/uniform

pynamodb/settings.py
anandswaminathan: Removed the default values.

All the default values are now in one place constants.py

pynamodb/settings.py
anandswaminathan: Yup. Done.

All the default values are now in one place constants.py

pynamodb/settings.py
```diff
@@ -0,0 +1,38 @@
...
+
+import logging
+
+log = logging.getLogger(__name__)
+
+OVERRIDE_SETTINGS_PATH = getenv('PYNAMO_CONFIG', '/etc/pynamodb/settings_override.py')
+
+settings = {}
+
+class RequestSessionWithHeaders(requests.Session):
```
danielhochman: i'm not sure why this still exists.

we will create a class with headers and pass it in settings_override.py

pynamodb/models.py
```diff
@@ -169,7 +170,13 @@ def __init__(cls, name, bases, attrs):
                     if not hasattr(attr_obj, HOST):
                         setattr(attr_obj, HOST, None)
                     if not hasattr(attr_obj, 'session_cls'):
-                        setattr(attr_obj, 'session_cls', None)
+                        setattr(attr_obj, 'session_cls', get_settings_value('session_cls'))
+                    if not hasattr(attr_obj, 'request_timeout_seconds'):
+                        setattr(attr_obj, 'request_timeout_seconds', get_settings_value('REQUEST_TIMEOUT_SECONDS'))
```
danielhochman: why is REQUEST_TIMEOUT_SECONDS capitalized and session_cls not?

pynamodb/constants.py
```diff
@@ -239,3 +239,8 @@
 AND = 'AND'
 OR = 'OR'
 CONDITIONAL_OPERATORS = [AND, OR]
+
+REQUEST_SESSION_HEADERS = dict()
```
danielhochman: this var is unused

pynamodb/models.py
```diff
@@ -138,8 +139,8 @@ def commit(self):
 
 class DefaultMeta(object):
```
danielhochman: this class can be changed to

```
class DefaultMeta(object):
    pass
```

pynamodb/settings.py
```diff
@@ -0,0 +1,36 @@
+from os import getenv
+from pynamodb.constants import DEFAULT_REGION
```
danielhochman: let's delete the `DEFAULT_REGION` constant entirely

pynamodb/settings.py
```diff
@@ -0,0 +1,36 @@
...
+import imp
+import os
+import logging
+
+log = logging.getLogger(__name__)
+
+default_settings_dict = {}
```
danielhochman: this should be defined as dict literal, e.g.

```
default_settings_dict = {
    'BASE_BACKOFF_MS': 10000,
    'REGION': 'us-east-1',
    ...
}
```

pynamodb/settings.py
```diff
@@ -0,0 +1,36 @@
...
+default_settings_dict['REQUEST_TIMEOUT_SECONDS'] = 25
+default_settings_dict['MAX_RETRY_ATTEMPTS'] = 3
+default_settings_dict['BASE_BACKOFF_MS'] = 25
+default_settings_dict['REGION'] = DEFAULT_REGION
+
+OVERRIDE_SETTINGS_PATH = getenv('PYNAMO_CONFIG', '/etc/pynamodb/settings_override.py')
```
danielhochman: let's make the env var `PYNAMODB_CONFIG`

danielhochman: also let's call the file `global_default_settings.py`

pynamodb/settings.py
```diff
@@ -0,0 +1,36 @@
+from os import getenv
```
danielhochman: https://www.python.org/dev/peps/pep-0008/#imports

```
Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports
You should put a blank line between each group of imports.
```

there is a tool called `isort` that can do it for you automatically too

pynamodb/settings.py
```diff
@@ -0,0 +1,36 @@
...
+default_settings_dict['REQUEST_TIMEOUT_SECONDS'] = 25
+default_settings_dict['MAX_RETRY_ATTEMPTS'] = 3
+default_settings_dict['BASE_BACKOFF_MS'] = 25
+default_settings_dict['REGION'] = DEFAULT_REGION
+
+OVERRIDE_SETTINGS_PATH = getenv('PYNAMO_CONFIG', '/etc/pynamodb/settings_override.py')
+
+override_settings = {}
+if os.path.isfile(OVERRIDE_SETTINGS_PATH):
+    override_settings= imp.load_source(OVERRIDE_SETTINGS_PATH, OVERRIDE_SETTINGS_PATH)
```
danielhochman: missing space after = sign

pynamodb/settings.py
```diff
@@ -0,0 +1,36 @@
...
+    log.info('Override settings for pynamo available {0}'.format(OVERRIDE_SETTINGS_PATH))
+else:
+    log.info('Override settings for pynamo not available {0}'.format(OVERRIDE_SETTINGS_PATH))
+    log.info('Using Default settings value')
+
```
danielhochman: two blank lines here

pynamodb/connection/base.py
```diff
@@ -189,16 +191,15 @@ def __init__(self, region=None, host=None, session_cls=None):
         else:
             self.region = DEFAULT_REGION
```
danielhochman: this should use get_settings_value mechanism and not default_region

pynamodb/connection/base.py
```diff
@@ -189,16 +191,15 @@ def __init__(self, region=None, host=None, session_cls=None):
...
-        self._request_timeout_seconds = DEFAULT_TIMEOUT
-        self._max_retry_attempts_exception = DEFAULT_MAX_RETRY_ATTEMPTS_EXCEPTION
-        self._base_backoff_ms = DEFAULT_BASE_BACKOFF_MS
-
         if session_cls:
             self.session_cls = session_cls
         else:
             self.session_cls = requests.Session
 
+        self._request_timeout_seconds = get_settings_value('REQUEST_TIMEOUT_SECONDS') if request_timeout_seconds is None else request_timeout_seconds
```
danielhochman: let's write them all out longform instead of shortened. lines start getting long

pynamodb/models.py
anandswaminathan: Done

pynamodb/settings.py
anandswaminathan: Done

pynamodb/settings.py
anandswaminathan: Done

pynamodb/settings.py
anandswaminathan: Done

pynamodb/settings.py
anandswaminathan: Done

pynamodb/settings.py
anandswaminathan: Done

pynamodb/settings.py
anandswaminathan: Done

pynamodb/connection/base.py
anandswaminathan: Done

pynamodb/connection/base.py
anandswaminathan: Done

danielhochman:
+1

PR #150, opened 2016-08-17
test the testsg

jmphilli:
Cannot deserialize / serialize a proper set of strings


jmphilli:
"clientIds": {
                "SS": [
                    "\"client1\"", 
                    "\"client2\""
                ]
            },

That's what pynamo saves

jmphilli:
But we have records saved without the extra quotes

PR #151, opened 2016-08-17
Serialize UnicodeSetAttributes correctly

jmphilli:
When writing strings to db do not json encode them.

If they were json encoded, decode them with json. Otherwise just return them.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -160,7 +160,7 @@ class UnicodeSetAttribute(SetMixin, Attribute):
                 result_set = set()
                 for str_val in value:
                     result_set.add(str(str_val))
-                return result_set
+                return list(result_set)
         return None
 
     def deserialize(self, value):
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -120,6 +120,22 @@ class BinarySetAttribute(SetMixin, Attribute):
     attr_type = BINARY_SET
     null = True
 
+    def serialize(self, value):
+        """
+        Returns a base64 encoded binary string
+        """
+        if value and len(value):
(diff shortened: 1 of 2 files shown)


pynamodb/attributes.py
```diff
@@ -66,44 +66,47 @@ def get_value(self, value):
 class SetMixin(object):
     """
     Adds (de)serialization methods for sets
+
+    def element_serialize(self, value):
+        raise NotImplemented()
+
+    def element_deserialize(self, value):
+        raise NotImplemented()
```
jmphilli: @danielhochman how do i do this? how do i 🐍 ?

jmphilli: oh i just switch the import order don't i?

danielhochman:
i think this could be cleaner by not requiring redefinition of serialization functions we already have on for the main attributes.

```
class MyMixin(object):
    def serialize_stuff(self):
        self.attr_class.serialize()

class BaseThing(object):
    @staticmethod
    def serialize():
        print 'ok'


class MyThing(MyMixin, BaseThing):
    attr_class = BaseThing

m = MyThing()

m.serialize_stuff()```

it requires the serialization method in the attributes to be static but i think that's an okay change. in this way we can call the original serializer rather than rewriting it.

seems no matter what we are going to break backwards compat if we try and fix this. i would vote we do the above or revert back to the other change where we override the Mixin serialize and leave the Mixin class for b, and can do the backwards compat breaking change in a later version.

danielhochman:
also i think there is a way to do this without breaking backwards compat at all but it would require pulling stuff out of `locals()` which is pretty hacky.

danielhochman:
+1

PR #156, opened 2016-09-05
Add the ability to treat Model instances as dictionaries

csimmons0:
This change allows Model instances to be treated as dictionaries mapping attribute names to values. There are two motivations for this change:
1. Make it easier to replace or mix boto3 with PynamoDB.
2. Make it easy to convert a model to a language-agnostic format (probably JSON).

Regarding the first motivation, boto3 operations that get table entries return the entries as dictionaries mapping attribute names to values. The following is a simplified code snippet from the product that I'm working on. It originally used boto3, but it now uses PynamoDB. Its continued use of "in" to check whether "optionalAttribute" has been set and its continued use of dictionary-style lookups of the "firstName," "lastName," and "email" attributes demonstrate how this change can make switching from boto3 to PynamoDB easier.

```
def set_optional_attribute_in_user(user_id, new_attribute_value):
    user = User.get(caller_id)  # This used to be the 'Item' from a boto3 'get_item' operation.
    if 'optionalAttribute' in user:
        raise ValueError('The user already has this attribute.')

    x = user['firstName']
    y = user['lastName']
    z = user['email']
    . . . do some other work . . .
    return dict(user)  # This used to be "return user"
```

Regarding the second motivation, I'm using Lambda and DynamoDB as the backend for an iOS app. The app, written in Swift, asks a Lambda function, written in Python, to get data from DynamoDB, and the Lambda function responds with JSON dictionaries mapping models' attributes to values. I believe that the existing functions to convert PynamoDB models to JSON, such as _serialize and _get_json, do not provide what I need. Those functions produce results that match the way models are stored inside DynamoDB, while what I need is an abstraction away from DynamoDB. 

With this change, one can call dict() on a model to get a Python dictionary mapping attribute names to values, and that Python dictionary can easily be converted to JSON. Lambda does it automatically if your function returns a Python dictionary. Any applications that use Python to access a DynamoDB database but need to manipulate the models in other programming languages may find this functionality useful.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -210,6 +210,9 @@ class AttributeDict(collections.MutableMapping):
     def __len__(self):
         return len(self._values)
 
+    def aliased_attr_names(self):
+        return self._alt_values.keys()
+
     def aliased_attrs(self):
         return self._alt_values.items()
 
@@ -754,6 +757,26 @@ class Model(with_metaclass(MetaModel)):
     def load(cls, filename):
         with open(filename, 'r') as inf:
             cls.loads(inf.read())
+
+    def keys(self):
+        # Get the names of this model's attributes, and filter out the
+        # attributes that are set to None in this instance of the model.
+        names = self._get_attributes().aliased_attr_names()
+        return [name for name in names if getattr(self, name) != None]
+
+    def __contains__(self, key):
+        return key in self.keys()
+
+    def __getitem__(self, key):
+        if not key in self:
+            raise KeyError(key)
+        return getattr(self, key)
+
+    def __setitem__(self, key, value):
+        # Require that the key is one of this model's attribute names.
+        if not key in self._get_attributes().aliased_attr_names():
+            raise KeyError(key)
+        setattr(self, key, value)
 
     # Private API below
     @classmethod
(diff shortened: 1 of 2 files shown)


ccverak:
+1

garrettheel:

inventionlabsSydney:
Is anyone prepared to review this?

garrettheel:
In principle I totally get the desire to be able to easily convert a model to a dictionary. I spoke with a few people about the best implementation and we settled on exposing an `as_dict` method on the model (similar to `MapAttribute`s) as a better approach. This would return a copy of the model data as a dictionary, but not expose any kind of additional set functionality. What do you think?

inventionlabsSydney:
@garrettheel If it's any consolation that's what I do at the moment. I attach to a parent class model of all my models a toolset of common methods, One of them is as_dict(). The code is:
```
    def as_dict(self):
        '''
        Takes the current model and reviews the attributes to then translate to a dict
        '''
        return {key: getattr(self, key) for key in self._get_attributes().keys()}
```

Obviously, this is super primitive and doesn't take into account maps of models, et-all.
Copies work fine in my opinion but if people want to be able to edit via dict instead of instance methods then I'd be inclined to run with that.
--Karl.

careless25:
Would this change also allow dictionaries to be converted to a model instance?

mariemargishvili:
👍 
This is a great change, can we get it in?

Vygo:
Is this gonna get in any time soon?

wobeng:
any update on this feature? @jlafon @garrettheel .

PR #160, opened 2016-09-18
Use Attribute Name Based on Attribute Class

lita:
If you use `attr_name`, such that the name in the Dynamo database is different from the ORM, update_item would do the wrong thing and create a new attribute using the ORM name.

This PR fixes that bug.


lita:
@danielhochman @jmphilli

pynamodb/models.py
```diff
@@ -349,12 +349,12 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
         if len(expected_values):
             kwargs.update(expected=self._build_expected_values(expected_values, UPDATE_FILTER_OPERATOR_MAP))
         kwargs[pythonic(ATTR_UPDATES)] = {
-            attribute: {
+            attribute_cls.attr_name: {
```
jmphilli: i wonder if we should pull the name out up [here](https://github.com/jlafon/PynamoDB/pull/160/files#diff-66d6c6d3d504b9e11a4e236e67133ddeR343) and save to a local var

pynamodb/models.py
```diff
@@ -349,12 +349,12 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
...
                 ACTION: action.upper() if action else None,
             }
         }
         if action is not None and action.upper() != DELETE:
-            kwargs[pythonic(ATTR_UPDATES)][attribute][VALUE] = {ATTR_TYPE_MAP[attribute_cls.attr_type]: value}
+            kwargs[pythonic(ATTR_UPDATES)][attribute_cls.attr_name][VALUE] = {ATTR_TYPE_MAP[attribute_cls.attr_type]: value}
```
jmphilli: we could then use that var here...

pynamodb/models.py
lita: did you want to save it to a local var for readability?

jmphilli: ya just so we don't have to keep accessing it that way.. seems error prone if someone changes this in the future, but may just be a matter of opinion

lita: ok

danielhochman:
+1

Issue #166, opened 2016-10-13
Model.refresh() doesn't respect attribute deletion


MarSoft:
Let's consider this situation.
I have a model, like this:

```
class User(Model):
    name = UnicodeAttribute(hash_key=True)
    subscriptions = UnicodeSetAttribute()
```

Now, let's there is a user with just one subscription:

```
user = User(name='john', subscriptions=set(['news']))
user.save()
```

Then, in some other place of the program, we unsubscribe user from `news`:

```
u = User.get('john')
u.subscriptions.discard('news')
u.save()
```

And finally, in the initial object, we request a refresh:

```
user.refresh()
```

**What happens:**
`user` object has stale value of `subscriptions` field
**Expected behaviour:**
Disappeared value should be noticed, and `user.subscriptions` should be set to `None`.

**How to fix:**
The problem seems to be in `Model._deserialize` method. It walks over received data object and then for each field found in the model it updates that field. What would probably be better is to walk over model fields and update them from incoming data, or set them to `None` if not found.

PR #167, opened 2016-10-13
Fix issue 166

MarSoft:
As described in the issue #166, this change should make `Model.refresh()` respect nullified attributes.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -1193,6 +1193,7 @@ class Model(with_metaclass(MetaModel)):
                 value = attrs[name].get(attr_type, None)
                 if value is not None:
                     setattr(self, name, attr.deserialize(value))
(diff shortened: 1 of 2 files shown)


MarSoft:
Two failing tests were caused by 2 reasons:
1. `Model._from_data` passed `hash_key` separate from other values, and hash_key was nullified by new implementation of `_deserialize`
2. My new implementation of `Model._deserialize`, just like previous one, didn't respect custom attribute names (#168). Even worse, `ModelTestCase.test_refresh` was broken too - it checked "internal" attribute name on the model (`item.user_name`) instead of "custom" one. Fixed this as well.

garrettheel:

jpinner-lyft:
Re-implemented on top of master in #388.

PR #178, opened 2016-10-31
List of non map attribute bug

jmphilli:


danielhochman:
incorporated into #175

PR #180, opened 2016-11-01
Make "unprocessed keys for batch operation" log at info level.

bedge:
Setting log level to info supporesses almost everything.
This feels like one thing that would be useful to see at info level.


danielhochman:
Thanks!

PR #183, opened 2016-11-03
Throw when raw MapAttribute is used

jmphilli:


jmphilli:
@danielhochman

jmphilli:
@yedpodtrzitko

yedpodtrzitko:
Thanks for the response. I've seen the examples you linked, but that doesn't work for me in this scenario, unfortunately.
I have a table which stores a generic JSON document in one of its attributes, and each JSON document can be a bit different (so I can't define the structure). The obvious solution would be to use the obsolete `JsonAttribute`, right? No. Please dont ask me any details about this, but there's another microservice working with this table, and its maintainer decided to replace PynamoDB with Flywheel, which supports a generic Map Attribute, so I can't store it as a serialized json, as the other microservice expects `Map` datatype there (yay microservices).

Anyway, I think I won't be the only one who would like to use a generic `MapAttribute`. So what I could imagine instead would be a flag on the `MapAttribute` class, which would allow skipping the validation (something like `MapAttribute(raw=True)` ).

The only change needed would be in `_serialize()`:

```python
if isinstance(attr, MapAttribute) and not attr.raw: # modified line
    if not value.validate():
        raise ValueError("Attribute '{0}' is not correctly typed".format(attr.attr_name))
    value = value.get_values()
``` 

this way it will be serialized into the `Map` datatype without enforcing any structure.

jmphilli:
hmm ya, i think this is a use case we do want to support, but i think it might take a little longer to build. 

Basically, i think we could have a `MapAttribute` that is not subclassed and has dicts, strings, numbers, whatever. We would have to look at each attribute in the class and serialize apporpriately.

Could you use a `Map` of `Json` in the interim? or a string or something?

yedpodtrzitko:
> but i think it might take a little longer to build. 

I dont think so. I am using a modified version of the code in which just the three lines nested under `if isinstance(attr, MapAttribute)` are not executed, and it's serialized correctly:

generic map serialized

jmphilli:
@yedpodtrzitko -- can you open a PR?

yedpodtrzitko:
> can you open a PR?
sure

PR #186, opened 2016-11-04
allow raw map attrs

jmphilli:
@yedpodtrzitko @danielhochman

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -1203,7 +1203,7 @@ class Model(with_metaclass(MetaModel)):
                     continue
                 elif null_check:
                     raise ValueError("Attribute '{0}' cannot be None".format(attr.attr_name))
-            if isinstance(attr, MapAttribute):
+            if isinstance(value, MapAttribute):
                 if not value.validate():
                     raise ValueError("Attribute '{0}' is not correctly typed".format(attr.attr_name))
                 value = value.get_values()
--- a/pynamodb/tests/test_model.py
+++ b/pynamodb/tests/test_model.py
@@ -3009,4 +3009,26 @@ class ModelTestCase(TestCase):
         map_serialized = {'M': {'foo': {'S': 'bar'}}}
         instance = ExplicitRawMapModel(map_attr=map_native)
         serialized = instance._serialize()
-        self.assertEqual(serialized['attributes']['map_attr'], map_serialized)
+        self.assertEqual(sorted(serialized['attributes']['map_attr']), sorted(map_serialized))
+
+    def test_raw_map_serialize_fun_one(self):
+        map_native = {
+            'foo': 'bar', 'num': 1, 'bool_type': True,
+            'other_b_type': False, 'floaty': 1.2, 'listy': [1,2,3],
+            'mapy': {'baz': 'bongo'}
+        }
+        map_serialized = {'M': {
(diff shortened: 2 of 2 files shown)


pynamodb/models.py
```diff
@@ -1203,7 +1203,7 @@ def _serialize(self, attr_map=False, null_check=True):
                     continue
                 elif null_check:
                     raise ValueError("Attribute '{0}' cannot be None".format(attr.attr_name))
-            if isinstance(attr, MapAttribute):
+            if isinstance(value, MapAttribute):
```
yedpodtrzitko: Oh, that's a really elegant solution.

jmphilli:
@yedpodtrzitko does this work for you? i like this better than introducing the `raw` field. I think it was just a bug where we were using the `attr` when we should've been using `value`. The `AttributeDict` class that those values comes from needs to be refactored. That's probably what i'll work on next. I copied your test and it looks like it works. I'm gonna let @danielhochman review since he's way better at 🐍 than I'll ever be 😄

yedpodtrzitko:
@jmphilli yes - this works fine, and I like it better than my PR too.

jmphilli:
@yedpodtrzitko -- daniel will release a new version with this patch later today. thank you again for the report

2x2xplz:
Hi, thanks for adding. Can you provide a quick code example for using this? A Python dict can be saved to Dynamo via MapAttribute however the reverse doesn't appear to work (probably because I am using it wrong). TableConnection works to get Map records, but then you have to wade through the raw Item.

yedpodtrzitko:
@2x2xplz could you show us the code which doesnt work?

2x2xplz:
@yedpodtrzitko  it's not that the code doesn't work -- I'd like to see an example because I don't know the syntax to convert a retrieved Map to a dict. For example these all fail:

```
def update_voting_status(tuple_list, new_status):
    for tuple in tuple_list:
        record = MyModel.get( tuple[0] , range_key=int(tuple[1]) )
        voting_status = record.voting_status
        print(voting_status) # <pynamodb.attributes.MapAttribute object at 0x0000000007C12390>
        #print(voting_status.get('votes'))  # error
        #print(voting_status.deserialize.get('votes')) # error
        #print(voting_status.keys()) # error
```
How do I convert a MapAttribute into a dict? Thanks

yedpodtrzitko:
@2x2xplz I'd say create a separate issue with more details (a minimal code sample how to reproduce the issue + traceback you're getting)

jmphilli:
```python
class MyModel(Model):
    class Meta:
        table_name = 'myModel'
        if settings.DYNAMODB_URL:
            host = settings.DYNAMODB_URL
    map_attr = MapAttribute(default={'foo': 'bar'})

>>> MyModel.create_table(read_capacity_units=10, write_capacity_units=10, wait=True)
>>> thing = MyModel(id='abc', map_attr={'num': 1, 'bool': True, 'str': 'hi'})
>>> thing.save()
{}
>>> new_thing = MyModel.get('abc')
>>> new_thing.map_attr
{u'num': 1, u'bool': True, u'str': u'hi'}
>>> new_thing.map_attr['num']
1

```

looks right to me @2x2xplz

yedpodtrzitko:
this was fixed by #191

PR #188, opened 2016-11-06
basic FilterExpression support for low-level api Scans

2x2xplz:
added 'long' data type to SERIALIZE_KEY_MAP and SERIALIZE_CLASS_MAP for 2.7 compatibility (AWS Lambda currently runs 2.7 only)


2x2xplz:
Also added very basic support for plain string FilterExpressions for Scans via raw Connection or TableConnection (not via Model). No support for Expression Attribute Names or Values. Barely tested but it's just an extra string parameter.

PR #194, opened 2016-11-08
Multiple attrs update

### LINKED ISSUE #157: Support for atomic multiple attribute update
In PynamoDB, the high-level API gives support for atomic updates using update_item. However, the current API supports updating one attribute at a time only. Sometimes, we need to atomically update multiple attributes of an item so that all of them succeed or fail. 

Currently, the only way to achieve this is o use the low-level API. It will be very helpful to support this in the high-level API as well.

yedpodtrzitko:
this allows to update multiple attributes at once using a new function `Model.update` (code based on the existing method `Model.update_item()`

This fixes #157 and #195

### DIFF
--- a/pynamodb/constants.py
+++ b/pynamodb/constants.py
@@ -166,7 +166,6 @@ QUERY_OPERATOR_MAP = {
 # These are the valid select values for the Scan operation
 # See: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html#DDB-Scan-request-Select
 NOT_NULL = 'NOT_NULL'
-#NULL = 'NULL' # declared as a type above
 CONTAINS = 'CONTAINS'
 NOT_CONTAINS = 'NOT_CONTAINS'
 ALL_ATTRIBUTES = 'ALL_ATTRIBUTES'
@@ -250,4 +249,3 @@ CONDITIONAL_OPERATOR = 'ConditionalOperator'
 AND = 'AND'
 OR = 'OR'
 CONDITIONAL_OPERATORS = [AND, OR]
-
--- a/pynamodb/constants.py
+++ b/pynamodb/constants.py
@@ -93,8 +93,11 @@ BINARY = 'Binary'
 BINARY_SET = 'BinarySet'
 MAP = 'Map'
 LIST = 'List'
+NULL = 'NULL'
+NULL_SHORT = 'NULL'
+
 SHORT_ATTR_TYPES = [STRING_SHORT, STRING_SET_SHORT, NUMBER_SHORT, NUMBER_SET_SHORT,
-                    BINARY_SHORT, BINARY_SET_SHORT, MAP_SHORT, LIST_SHORT, BOOLEAN_SHORT]
+                    BINARY_SHORT, BINARY_SET_SHORT, MAP_SHORT, LIST_SHORT, BOOLEAN_SHORT, NULL_SHORT]
 ATTR_TYPE_MAP = {
     STRING: STRING_SHORT,
     STRING_SET: STRING_SET_SHORT,
@@ -110,7 +113,8 @@ ATTR_TYPE_MAP = {
     BINARY_SET_SHORT: BINARY_SET,
     MAP: MAP_SHORT,
     LIST: LIST_SHORT,
-    BOOLEAN: BOOLEAN_SHORT
+    BOOLEAN: BOOLEAN_SHORT,
+    NULL: NULL_SHORT,
 }
 # Constants needed for creating indexes
 LOCAL_SECONDARY_INDEX = 'LocalSecondaryIndex'
@@ -162,7 +166,7 @@ QUERY_OPERATOR_MAP = {
 # These are the valid select values for the Scan operation
 # See: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html#DDB-Scan-request-Select
 NOT_NULL = 'NOT_NULL'
-NULL = 'NULL'
+#NULL = 'NULL' # declared as a type above
 CONTAINS = 'CONTAINS'
 NOT_CONTAINS = 'NOT_CONTAINS'
 ALL_ATTRIBUTES = 'ALL_ATTRIBUTES'
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -1261,9 +1261,9 @@ class Model(with_metaclass(MetaModel)):
                 attrs[attributes][attr.attr_name] = serialized
             else:
                 if attr.is_hash_key:
-                    attrs[HASH] = list(serialized.values())[0]
+                    attrs[HASH] = serialized[ATTR_TYPE_MAP[attr.attr_type]]
                 elif attr.is_range_key:
-                    attrs[RANGE] = list(serialized.values())[0]
+                    attrs[RANGE] = serialized[ATTR_TYPE_MAP[attr.attr_type]]
                 else:
                     attrs[attributes][attr.attr_name] = serialized
 
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -323,6 +323,8 @@ class Model(with_metaclass(MetaModel)):
         :param action: The action to take if this item already exists.
             See: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#DDB-UpdateItem-request-AttributeUpdate
         """
+        DeprecationWarning("`Model.update_item` is deprecated in favour of `Model.update` now")
+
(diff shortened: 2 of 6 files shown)


danielhochman:
i think i'd prefer to move multiple update functionality to a method called `update()` and deprecate `update_item()`. `update_item()` never works how anyone would expect it to.

thoughts?

yedpodtrzitko:
Could be. The code will be a wee bit simpler without the backward compatibility + I'll also try to implement a fix for #195 there.

yedpodtrzitko:
@danielhochman pls let me know if you have any feedback on the current implementation. Othewise I'll just update docs, and it's good to go.

yedpodtrzitko:
@danielhochman 
- `update()` implemented (at least in the `Model`)
- saving `null` value for attributes with `null=True` is now handled properly

let me know if you have any suggestions about the code. Otherwise I'll just add docs & it's good to go.

pynamodb/models.py
```diff
@@ -323,6 +323,8 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
         :param action: The action to take if this item already exists.
             See: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#DDB-UpdateItem-request-AttributeUpdate
         """
+        DeprecationWarning("`Model.update_item` is deprecated in favour of `Model.update` now")
```
yedpodtrzitko: ah! fixed.

pynamodb/constants.py
```diff
@@ -162,7 +166,7 @@
 # These are the valid select values for the Scan operation
 # See: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html#DDB-Scan-request-Select
 NOT_NULL = 'NOT_NULL'
-NULL = 'NULL'
+#NULL = 'NULL' # declared as a type above
```
jmphilli: can we just delete this maybe?

yedpodtrzitko:
@jmphilli @danielhochman anything else left to do here?

pynamodb/models.py
```diff
@@ -1211,21 +1260,32 @@ def _serialize(self, attr_map=False, null_check=True):
...
         return attrs
 
     @classmethod
+    def _serialize_value(cls, attr, value):
+        """
+        Serializes a value for use with DynamoDB
+
+        :param value: a value to serialize
+        """
+        serialized = attr.serialize(value)
```
jmphilli: the passed value is already serialized on [1210](https://github.com/jlafon/PynamoDB/pull/194/files#diff-66d6c6d3d504b9e11a4e236e67133ddeL1210) right?

we should either remove that or skip this

pynamodb/models.py
```diff
@@ -1211,21 +1260,32 @@ def _serialize(self, attr_map=False, null_check=True):
...
+        if serialized is None:
+            if not attr.null:
+                raise ValueError("Attribute '{0}': cannot be None".format(attr.attr_name))
+            return {NULL: True}
```
jmphilli: can we just set `serialized = True` and let the `ATTR_TYPE_MAP` on line 1286 handle the null bit? seems nicer maybe? don't really know tbh

pynamodb/models.py
```diff
@@ -357,6 +361,51 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
...
+        }
+
+        if pythonic(RANGE_KEY) in save_kwargs:
+            kwargs[pythonic(RANGE_KEY)] = save_kwargs[pythonic(RANGE_KEY)]
+
+        if expected_values:
+            kwargs['expected'] = self._build_expected_values(expected_values, UPDATE_FILTER_OPERATOR_MAP)
+
+        for attr, params in attributes.items():
+            attribute_cls = self._get_attributes()[attr]
```
jmphilli: might move the `self._get_attributes()` call outside the loop?

pynamodb/models.py
```diff
@@ -357,6 +361,51 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
...
+            action = params['action'] and params['action'].upper()
```
jmphilli: won't this return a `bool` ?

pynamodb/models.py
```diff
@@ -357,6 +361,51 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
...
+            attr_values = {ACTION: action}
+            if action != DELETE:
```
jmphilli: can we add a test for `DELETE`

pynamodb/models.py
yedpodtrzitko: no, either `False` (in case `params['action']` is evaluated to `False`) or the value of the second statement)
>>> a = {'none': False, 'one': 'yes'}
```
>>> a['none'] and a['none'].upper()
False
>>> a['one'] and a['one'].upper()
'YES'
```

pynamodb/models.py
yedpodtrzitko: sure

pynamodb/models.py
yedpodtrzitko: ok

pynamodb/models.py
yedpodtrzitko: No, this is a fix for #195

yedpodtrzitko:
@jmphilli bump

pynamodb/models.py
```diff
@@ -357,6 +361,52 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
                 setattr(self, name, attr.deserialize(value.get(ATTR_TYPE_MAP[attr.attr_type])))
         return data
 
+
+    def update(self, attributes,  conditional_operator=None, **expected_values):
```
jmphilli: would be good to do now since this is new...

jmphilli: cc @danielhochman

pynamodb/models.py
```diff
@@ -1198,34 +1248,49 @@ def _serialize(self, attr_map=False, null_check=True):
...
+
             if attr_map:
-                attrs[attributes][attr.attr_name] = {
-                    ATTR_TYPE_MAP[attr.attr_type]: serialized
-                }
+                attrs[attributes][attr.attr_name] = serialized
             else:
                 if attr.is_hash_key:
-                    attrs[HASH] = serialized
+                    attrs[HASH] = list(serialized.values())[0]
```
jmphilli: ?

yedpodtrzitko: I replaced `serialized = attr.serialize(value)` above with `_serialize_value()` in order to avoid double call of `attr.serialize()` you did not like.

Since `_serialize_value()` returns value in the DynamoDB serialization format (ie. `{TYPE: VALUE}`), I need to extract the value here. 

List conversion is done because Python 3 the does not support indexing of the type returned by `values()`

pynamodb/models.py
yedpodtrzitko: @jmphilli you might need to pass a value for `action` beside the `value`. That would mean a bunch of small dictionaries like `attribute_name={'value': 1, 'action': 'add'}`.

jmphilli:
daniel is out for a week. i'll look at this again today, but i can only give my seal of approval which doesn't really do much for you unfortunately lol.

yedpodtrzitko:
np, I appreciate your feedback.

pynamodb/models.py
```diff
@@ -1198,34 +1248,49 @@ def _serialize(self, attr_map=False, null_check=True):
...
-                elif null_check:
-                    raise ValueError("Attribute '{0}' cannot be None".format(attr.attr_name))
             if isinstance(value, MapAttribute):
                 if not value.validate():
                     raise ValueError("Attribute '{0}' is not correctly typed".format(attr.attr_name))
                 value = value.get_values()
-            serialized = attr.serialize(value)
-            if serialized is None:
+
+            serialized = Model._serialize_value(attr, value, null_check)
```
jmphilli: self._serialize_value

pynamodb/models.py
```diff
@@ -357,6 +361,52 @@ def update_item(self, attribute, value=None, action=None, conditional_operator=N
...
+        if expected_values:
+            kwargs['expected'] = self._build_expected_values(expected_values, UPDATE_FILTER_OPERATOR_MAP)
+
+        attrs = self._get_attributes()
+        for attr, params in attributes.items():
+            attribute_cls = attrs[attr]
+            action = params['action'] and params['action'].upper()
+            attr_values = {ACTION: action}
+            if action != DELETE:
+                attr_values[VALUE] = Model._serialize_value(attribute_cls, params['value'])
```
jmphilli: self._serialize_value

PR #197, opened 2016-11-10
Fixing the session headers being ignored because of prepared requests

anandswaminathan:
Ref: http://docs.python-requests.org/en/master/user/advanced/

The requests session headers are ignored in the request, if the prepared request object is passed in.
The fix calls requests_session.prepare_request by creating a request object from the aws_request_object

@danielhochman


pynamodb/connection/base.py
```diff
@@ -228,6 +229,23 @@ def _log_error(self, operation, response):
...
+        """
+        Create a prepared request object from request_dict, and operation_model
+        """
+        boto_prepared_request = self.client._endpoint.create_request(request_dict, operation_model)
+
+        # The call requests_session.send(final_prepared_request) ignores the headers which are
+        # part of the request session. In order to include the requests session headers inside
+        # the request, we create a new request object, and call prepare_request with the newly
+        # created request object
+        raw_request_with_params = Request(boto_prepared_request.method,
```
danielhochman: this indentation is not consistent with the rest of the code base

please use

```
raw_request_with_params = Request(
    boto_prepared_request.method,
    boto...,
    ...,
)
```

danielhochman:
+1 after comments

PR #201, opened 2016-11-18
use attributes_to_get argument

lukedeo:
Ensure that `attributes_to_get` is used when using `batch_get`.


jmphilli:
odd that this was missed... can you add a test that fails before your change?

devganesh:
If you try out the example [here](http://pynamodb.readthedocs.io/en/latest/batch.html) with the `attributes_to_get` parameter non-null, you will see that the `batch_get()` method returns _all_ fields regardless of what's in the parameter.

jmphilli:
this looks great! would love to see a test 😜

PR #203, opened 2016-11-23
add equals method

jmphilli:


yedpodtrzitko:
tests?

jmphilli:
daniel and i are chatting about how this would work with floats...

yedpodtrzitko:
The "easy" solution I can think of is to add a new parameters: `NumberAttribute.__init__(... float_cmp_deviation=0.0001)` + add `NumberAttribute.__eq__()`, which would take the value of `float_cmp_deviation` into consideration. 

Is there anything I am missing?

PR #206, opened 2016-12-02
Add distinct DoesNotExist classes per model

ikonst:
It should be possible to do:
```py
try:
    doStuffWithModels()
except MyModel.DoesNotExist:
    print('My model does not exist')  # we'd end up here for MyOtherModel as well
except MyOtherModel.DoesNotExist:
    print('My other model does not exist')
```

Currently, `MyModel.DoesNotExist == MyOtherModel.DoesNotExist`. This change makes every model have its own subclass of `pynamodb.exceptions.DoesNotExist`.

btw, this is similar in spirit to what mongoengine does in https://github.com/MongoEngine/mongoengine/blob/3c455cf1c13f923db40de4a855dd402217bee3e1/mongoengine/base/metaclasses.py#L402-L407.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -186,6 +186,8 @@ class MetaModel(type):
             if META_CLASS_NAME not in attrs:
                 setattr(cls, META_CLASS_NAME, DefaultMeta)
 
+            # create a custom Model.DoesNotExist derived from pynamodb.exceptions.DoesNotExist,
+            # so that "except Model.DoesNotExist:" would not catch other models' exceptions
             if 'DoesNotExist' not in attrs:
                 exception_attrs = {'__module__': attrs.get('__module__')}
                 if hasattr(cls, '__qualname__'):  # On Python 3, Model.DoesNotExist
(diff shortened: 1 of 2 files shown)


pynamodb/tests/test_model.py
```diff
@@ -1925,6 +1925,24 @@ def fake_dynamodb(*args):
             self.assertRaises(UserModel.DoesNotExist, UserModel.get, 'foo', 'bar')
 
         with patch(PATCH_METHOD) as req:
+            req.return_value = {}
+            try:
+                UserModel.get('foo')
+            except SimpleUserModel.DoesNotExist:
+                assert False  # DoesNotExist exceptions must be distinct per-model
```
mxr: Consider `self.fail('DoesNotExist exceptions must be distinct per-model')` instead

pynamodb/models.py
```diff
@@ -186,6 +186,12 @@ def __init__(cls, name, bases, attrs):
             if META_CLASS_NAME not in attrs:
                 setattr(cls, META_CLASS_NAME, DefaultMeta)
 
+            if 'DoesNotExist' not in attrs:
```
danielhochman: can you add a comment explaining what this code is for? basically repeating what's in the PR description already

PR #208, opened 2016-12-11
Replace Delorean with dateutil, as dateutil is brought in by botocore

funkybob:
No point depending on two libs that provide the same services.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ if sys.argv[-1] == 'publish':
 install_requires = [
     'six',
     'botocore>=1.0.0',
-    'python-dateutil>=2.6.0',
+    'python-dateutil>=2.1,<3.0.0',
 ]
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,6 @@ if sys.argv[-1] == 'publish':
     sys.exit()
 
 install_requires = [
-    'Delorean',
     'six',
     'botocore>=1.0.0'
 ]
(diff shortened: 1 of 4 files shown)


Dunedan:
I'd really appreciate seeing this being merged, as `Delorean` and its dependencies pull in additional ~25MB of unnecessary packages.

funkybob:
@brandond yeah, that's why I made the change - great lib, terribly weighty deps...

setup.py
```diff
@@ -12,7 +12,6 @@
     sys.exit()
 
 install_requires = [
-    'Delorean',
```
garrettheel: Can you declare `python-dateutil` here to make it clear we're relying on it?

garrettheel:
Tests still use `Delorean`. Can you replace these with some that sanity check `dateutil` instead?

setup.py
```diff
@@ -12,9 +12,9 @@
...
     'six',
-    'botocore>=1.0.0'
+    'botocore>=1.0.0',
+    'python-dateutil>=2.6.0',
```
garrettheel: Can you add `<3.0.0` here to match botocore also?

garrettheel:
Thanks for the updates! One more small comment and then this should be good to go.

garrettheel:
Thanks for contributing

geeknam:
Any chance of releasing a pypi version that includes this change? This would make my packaged lambdas much smaller.

PR #213, opened 2017-01-05
Update `Model::__init__` to accept dict or subclassed instance and instantiate appropriately

jmphilli:
The `Model::__init__` method should instantiate any subclassed `MapAttribute` if passed a dictionary, or set the passed instantiated instance on the current model.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -391,20 +391,25 @@ class MapAttribute(with_metaclass(MapAttributeMeta, Attribute)):
         """
         Sets the attributes for this object
         """
-        for attr_name, attr in self._get_attributes().aliased_attrs():
-            if attr.attr_name in attrs:
-                value = attrs.get(attr_name)
-                if not isinstance(value, collections.Mapping) or type(attr) == MapAttribute:
-                    setattr(self, attr_name, attrs.get(attr.attr_name))
-                else:
-                    # it's a sub model which means we need to instantiate that type first
-                    # pass in the attributes of that model, then set the field on this object to point to that model
-                    sub_model = value
-                    instance = type(attr)(**sub_model)
-                    setattr(self, attr_name, instance)
+        class_attributes = self._get_attributes().aliased_attrs()
+        if len(class_attributes) > 0:
+            for attr_name, attr in class_attributes:
+                if attr.attr_name in attrs:
+                    value = attrs.get(attr_name)
+                    if not isinstance(value, collections.Mapping) or type(attr) == MapAttribute:
+                        setattr(self, attr_name, attrs.get(attr.attr_name))
+                    else:
+                        # it's a sub model which means we need to instantiate that type first
+                        # pass in the attributes of that model, then set the field on this object to point to that model
+                        sub_model = value
+                        instance = type(attr)(**sub_model)
+                        setattr(self, attr_name, instance)
 
-            elif attr_name in attrs:
-                setattr(self, attr_name, attrs.get(attr_name))
+                elif attr_name in attrs:
+                    setattr(self, attr_name, attrs.get(attr_name))
+        else:  # it's a raw MapAttribute
+            for in_memory_attribute_name, in_memory_attribute_value in six.iteritems(attrs):
+                setattr(self, in_memory_attribute_name, in_memory_attribute_value)
 
     def get_values(self):
         attributes = self._get_attributes()
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -391,9 +391,10 @@ class MapAttribute(with_metaclass(MapAttributeMeta, Attribute)):
         """
         Sets the attributes for this object
         """
-        class_attributes = self._get_attributes().aliased_attrs()
+        class_attributes = self._get_attributes()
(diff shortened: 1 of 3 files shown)


pynamodb/attributes.py
```diff
@@ -387,27 +387,44 @@ def _get_attributes(cls):
                 cls._attributes[item] = instance
         return cls._attributes
 
+    def _get_in_memory_attributes(self):
+        try:
+            in_memory_attributes = getattr(self, 'in_memory_attributes')
```
yedpodtrzitko: I'd go with `in_memory_attributes = self.in_memory_attributes` (which would raise `AttributeError` too), or `if hasattr(self, 'in_memory_attributes'): ...` to avoid raising the exception at all.

pynamodb/attributes.py
```diff
@@ -387,27 +387,44 @@ def _get_attributes(cls):
...
-                    # pass in the attributes of that model, then set the field on this object to point to that model
-                    sub_model = value
-                    instance = type(attr)(**sub_model)
-                    setattr(self, attr_name, instance)
-
-            elif attr_name in attrs:
-                setattr(self, attr_name, attrs.get(attr_name))
+        class_attributes = self._get_attributes()
+        aliased_attributes = class_attributes.aliased_attrs()
+        if len(class_attributes) > 0:
```
danielhochman: pythonic is `if class_attributes:`

jmphilli:
@danielhochman @garrettheel

@yedpodtrzitko -- this is going to be a breaking change for your code i'm pretty sure.

currently a raw `MapAttribute` is treated like a dict in application code...
e.g.
```python
class Book(Model):
  id = UnicodeAttribute()
  author = MapAttribute()
  publisher = publisher()

class Publisher(MapAttribute):
  lat = NumberAttribute()
  lng = NumberAttribute()
  name = UnicodeAttribute()

p = Publisher(lat=12, lng=3, name=’somewhere in the blue’)
book = Book(id=’abc’, author={‘fname’: ‘justin’, ‘lname’: ‘p’, age: 3}, publisher=p)

assert book.publisher.lat == 12 # works because publisher is a python object
assert book.author.fname == ‘justin’ # raises AttributeError because author is a dict
```

We want to change it so that you can pass in python dicts and get the correct typing back.
e.g. we want this to work

```python
book = Book(id=’abc’, author={‘fname’: ‘justin’, ‘lname’: ‘p’, age: 3}, publisher={‘lat’: 12, ‘lng’:3, ‘name’: ‘somewhere in the blue’})
```

in addition to accepting raw dicts and constructing the subclass correctly in the constructor, we want to standardize the `get` api

we want this to work

```python
...
assert book.author.fname == ‘justin’ # works, is True
```

we want to break the functionality that allows you to access raw MapAttributes as a python dictionary, which i'm pretty sure will break your application code @yedpodtrzitko 

how do you feel about that? how big of a change is it on your side to make the necessary changes?

yedpodtrzitko:
Hi,
thanks for asking. I'm fine with it, go ahead .)

pynamodb/models.py
```diff
@@ -1231,10 +1232,18 @@ def _set_attributes(self, **attrs):
...
-                setattr(self, attr_name, attrs.get(attr.attr_name))
+                attribute_name = attr.attr_name
             elif attr_name in attrs:
-                setattr(self, attr_name, attrs.get(attr_name))
+                attribute_name = attr_name
+            if attribute_name is not None:
+                attribute_to_set = attrs.get(attribute_name)
+                if isinstance(attribute_to_set, Mapping):
+                    kls = getattr(self.__class__, attribute_name).__class__
+                    if kls is not None.__class__ and attribute_to_set is not None:
```
danielhochman: why not check if attr value is None instead of kls equivalence?

pynamodb/attributes.py
```diff
@@ -387,27 +387,42 @@ def _get_attributes(cls):
...
+                        # it's a sub model which means we need to instantiate that type first
+                        # pass in the attributes of that model, then set the field on this object to point to that model
+                        sub_model = value
+                        instance = type(attr)(**sub_model)
+                        setattr(self, attr_name, instance)
+
+                elif attr_name in attrs:
+                    setattr(self, attr_name, attrs.get(attr_name))
+        else:  # it's a raw MapAttribute
+            in_memory_attributes = self._get_in_memory_attributes()
```
lita: nit: I would move this into a separate method  and call it `_process_raw_map_attribute` rather than putting a comment.

pynamodb/attributes.py
```diff
@@ -387,27 +387,42 @@ def _get_attributes(cls):
...
-                    sub_model = value
-                    instance = type(attr)(**sub_model)
-                    setattr(self, attr_name, instance)
-
-            elif attr_name in attrs:
-                setattr(self, attr_name, attrs.get(attr_name))
+        class_attributes = self._get_attributes()
+        aliased_attributes = class_attributes.aliased_attrs()
+        if class_attributes:
+            for attr_name, attr in aliased_attributes:
```
lita: nit: if we can move this into a separate method as well and call it `_process_class_attributes` that would be nice.

jmphilli:
gheel is doing it

PR #220, opened 2017-01-14
Remove AttributeDict and be explicit about attr names

garrettheel:
`AttributeDict` existed to help with translation between the _python_ name for an attribute (i.e the class attr) and the _dynamodb_ name, which differ when the `attr_name` kwarg is specified. In practice, this shim made things confusing and caused bugs (see below).

This PR replaces `AttributeDict` with explicit calls when conversion is needed. It also factors out an `AttributeContainer` mixin so that `Model` and `MapAttribute` can share these common traits (this was previously duplicated).

**Bug fix**: Overridden `attr_name`s are no longer allowed when referring to attributes
```python
class MyModel(Model):
  custom_id = NumberAttribute(attr_name="id")

MyModel(id=1)  # This will now raise a ValueError
MyModel.query(id=1)  # This will now also raise a ValueError when the generator is evaluated (e.g when iterated over). .save(), .count(), etc. are also affected
MyModel(custom_id=1).id  # This will raise an AttributeError now
```

**Bug fix**: An overridden `attr_name` in a MapAttribute is now serialized correctly

```python
class MyMap(MapAttribute):
    custom_id = NumberAttribute(attr_name="id")

class MyModel(Model):
  map = MyMap()

m = MyModel(map=MyMap(custom_id=1)).serialize()  # This used to serialize with "custom_id" instead of "id"
```

**Bug fix**: `update` and `update_item` now correctly set attributes returned in the response when `attr_name` is overridden

```python
class SimpleUserModel(Model):
    user_name = UnicodeAttribute(hash_key=True)
    email = UnicodeAttribute()
    custom_aliases = UnicodeSetAttribute(attr_name='aliases')

item = SimpleUserModel('foo', email='bar')
item.update_item('custom_aliases', set(['lita']), action='add')

item.custom_aliases  # this used to return None, now returns the correct value
```


pynamodb/attributes.py
```diff
@@ -64,6 +63,54 @@ def get_value(self, value):
...
+    _dynamo_to_python_attrs = None
+
+    @classmethod
+    def _initialize_attributes(cls):
+        """
+        Initialize attributes on the class.
+        """
+        cls._attributes = {}
+        cls._dynamo_to_python_attrs = {}
+        for item in dir(cls):
```
yedpodtrzitko: iterate through `cls.__dict__` rather than using `dir()`

yedpodtrzitko: that way the nested `getattr(getattr())` below can be removed, since each thing in `__dict__` has `__class__`

garrettheel: This is the original code from `Model` unchanged, but will look at cleaning it up as you suggested

yedpodtrzitko: Thank you. I noticed it later too, but it would be better to clean it up while the code is touched.

garrettheel: Agreed :) I just refactored this a bit - let me know what you think

garrettheel:
@danielhochman wanna take a look at this?

garrettheel:
Also, there's one outstanding thing I'd like to fix but it may have further reaching implications. I'd like to get opinions.

```python
class MyModel(Model):
  custom_id = NumberAttribute(attr_name="id")

m MyModel(id=1)  # This currently does not raise an error
```

I think the above snippet should raise a `ValueError`, as we should only support a single way of accessing attributes in python. Does anyone agree/disagree?

yedpodtrzitko:
Yup I agree, that should raise an error. Also what Zen says:

> Errors should never pass silently
> There should be one-- and preferably only one --obvious way to do it

danielhochman:
+1

PR #221, opened 2017-01-18
Ensure defaults are respected for MapAttribute

garrettheel:
Defaults were never being set on `MapAttribute`s.

Also removes some complicated type validation logic that was never _actually_ validating types. Leaving that to a future PR, as it's a much larger task.


jmphilli:
+1

PR #222, opened 2017-01-18
Exponential backoff retries for ProvisionedThroughputExceeded errors

JohnEmhoff:
Currently this is a hard failure -- retry with backoff instead. Handling this client side is hard because `batch_write.commit()` clears its list of pending operations on failure.

Related issue: https://github.com/jlafon/PynamoDB/issues/218

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -332,9 +332,10 @@ class Connection(object):
                     if is_last_attempt_for_exceptions:
                         log.debug('Reached the maximum number of retry attempts: %s', attempt_number)
                         raise
-                    elif response.status_code < 500:
-                        # We don't retry on a ConditionalCheckFailedException or other 4xx because we assume they will
-                        # fail in perpetuity. Retrying when there is already contention could cause other problems
+                    elif response.status_code < 500 and code != 'ProvisionedThroughputExceededException':
+                        # We don't retry on a ConditionalCheckFailedException or other 4xx (except for
+                        # throughput related errors) because we assume they will fail in perpetuity.
+                        # Retrying when there is already contention could cause other problems
                         # in part due to unnecessary consumption of throughput.
                         raise
                     else:


pynamodb/connection/base.py
```diff
@@ -332,9 +332,10 @@ def _make_api_call(self, operation_name, operation_kwargs):
                     if is_last_attempt_for_exceptions:
                         log.debug('Reached the maximum number of retry attempts: %s', attempt_number)
                         raise
-                    elif response.status_code < 500:
-                        # We don't retry on a ConditionalCheckFailedException or other 4xx because we assume they will
-                        # fail in perpetuity. Retrying when there is already contention could cause other problems
+                    elif response.status_code < 500 and code != 'ProvisionedThroughputExceededException':
```
danielhochman: let's make it configurable

brandond:
@JohnEmhoff Are you still working on this?

JohnEmhoff:
I haven't had a chance to update it to be configurable, no. Honestly I'm
not sure what the use case is to have it off. For example, when writing,
pynamo clears the batch object ahead of the write -- the application code
thus can't retry the writes that failed because it doesn't know what they
are.

bedge:
I"m doing this (some ugly variant anyway) in all my clients already, so definitely a +1 to merge. 
I see no reason not to do this for all cases.

lukedeo:
+1 here as well

bedge:
Any updates on this? Merging this addresses a known issue, which as of right now is a PITA to deal with. 
Is there really any downside to retrying?

danielhochman:
at Lyft we prefer backpressure to go to the client. i will merge since we actually don't use the retry within PynamoDB in most cases.

it actually seems like a bigger bug that batch_write will clear on error. retry won't guarantee success, particularly when experiencing ProvisionedThroughputExceptions.

can someone create an issue to track that in more detail?

bedge:
Agree that the batch_write clear on error is a more significant issue. However, this addresses a specific case we see regularly, which is that capacity provisioning lags the dynamodb I/O load and retries early on in the capacity scale-up phase eliminate a bunch of client-side retries.

thanks!

danielhochman:
@JohnEmhoff @bedge @lukedeo @brandond released as part of 2.1.5

PR #223, opened 2017-01-19
Always create map attributes when setting a dict

garrettheel:
Alternate implementation of https://github.com/jlafon/PynamoDB/pull/213

Note that this changes the semantics of `MapAttribute`s a little such that you will _always_ get back a `MapAttribute` instance when accessing via a parent (Model or MapAttribute) and be required to call `.as_dict()` to convert into a raw dict.

The aim of doing this is consistency -- whether you instantiate a model like `Model(attr={...})` or `Model(attr=MapAttribute())` you should get the same result back when accessing via `model.attr`.

This is a breaking change so we should be sure to document it well in release notes.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -419,6 +419,7 @@ class MapAttribute(AttributeContainer, Attribute):
                                            null=null,
                                            default=default,
                                            attr_name=attr_name)
+        self._get_attributes()  # Ensure attributes are always inited
         self.attribute_values = {}
         self._set_defaults()
         self._set_attributes(**attrs)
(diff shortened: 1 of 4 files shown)


pynamodb/attributes.py
```diff
@@ -406,6 +406,7 @@ def __init__(self, hash_key=False, range_key=False, null=None, default=None, att
                                            null=null,
                                            default=default,
                                            attr_name=attr_name)
+        self._is_raw_map = bool(type(self) == MapAttribute)
```
yedpodtrzitko: `is`  instead of `==` pls

also explicit cast to `bool` is not necessary (?)

garrettheel: I was under the impression that `==` is correct (and more pythonic?) for checking that the _value_ of two types is the same

yedpodtrzitko: Hm... I've been using it always that way without thinking about it.

The constant objects like `True` `False` `None` are compared via `is`. `type` is a reference to the same object, and you want to be sure it is the very same object, so `is` feels right to me there. But... I dont insist on it that much.

```python
class Foo(dict):
    pass

a = Foo()
b = Foo()
id(type(a))
# -> 28865504
id(type(b))
# -> 28865504
```

garrettheel: Yep, I understand that this seems to be a bit of a gray area there. My feeling is that we're getting back references to a class which is _not_ one of the constants, so they should be compared with `==` incase there is some situation where they're stored separately in memory (I don't know of a case where this happens off hand, but it seems like it plausibly could happen)

pynamodb/attributes.py
```diff
@@ -406,6 +406,7 @@ def __init__(self, hash_key=False, range_key=False, null=None, default=None, att
                                            null=null,
                                            default=default,
                                            attr_name=attr_name)
+        self._get_attributes()  # Ensure attributes are always inited
```
jmphilli: nice, i prefer this as well

pynamodb/attributes.py
```diff
@@ -471,16 +468,38 @@ def serialize(self, values):
...
+            deserialized_dict[key] = attr_class.deserialize(attr_value)
+
+        # If this is a subclass of a MapAttribute (i.e typed), instantiate an instance
+        if not self.is_raw():
+            return type(self)(**deserialized_dict)
         return deserialized_dict
 
+    @classmethod
+    def is_raw(cls):
+        return cls == MapAttribute
```
garrettheel: I don't love this, but I don't know of another way of doing this without catching subclasses, since `issubclass(MapAttribute, MapAttribute)` also returns True.

danielhochman:
+1

PR #226, opened 2017-01-27
add a provision to get LastEvaluatedKey from last executed query, 

aonamrata:
so we can paginate manually. We are building an api where we would like to send the limit and page_no in and would like to return a chunk of data. 
eg: 
GET /posts/user1?limit=5
```{
items : [...],
"pagination": {
   "type": "standard",
  "nextPageKey": "abc",
  "limit": 5,
  "total_records": 7
 } }
```
GET /posts/user1?nextPageKey=abc&limit=5
```{
items : [...],
"pagination": {
   "type": "standard",
  "nextPageKey": "",
  "limit": 5,
  "total_records": 2
 } }
```
In order to make 2 queries() that can continue data set, we need a way to get the last querie's `LastEvaluatedKey` value.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -205,6 +205,7 @@ class Model(AttributeContainer):
     _index_classes = None
     _throttle = NoThrottle()
     DoesNotExist = DoesNotExist
+    _last_query_evaluated_key = None
 
     def __init__(self, hash_key=None, range_key=None, **attrs):
         """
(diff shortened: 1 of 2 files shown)


garrettheel:
This is likely to be error-prone when used in `gevent` environments, where `cls._last_query_evaluated_key` will be incorrect given any context switching happening between the set and when you read it from the class.

aonamrata:
hmm, I see what you mean @garrettheel. It will have to be called exactly after query() to get the right key. 
Maybe send a param in query say, getNextKey as boolean and return the key in yield item itself.. that way concurrency might not be a problem. We will just have to get it from model obj 🤔

jl-DaDar:
maybe a better approach is to pass in last_evaluated_key as an object and then update the object within the scan, query, rate_limited_scan?

PR #230, opened 2017-02-10
use range key when not none, otherwise batch get all without range

jmphilli:
allows us to look up all models by hash key when range is defined
```python
identifiers = [(identifier, None) for identifier in identifiers]
for result in model.batch_get(identifiers):
```


jmphilli:
@garrettheel

jmphilli:
@danielhochman

garrettheel:
Can you add a test for this?

jmphilli:
oh fo sho -- does the api make sense?

Issue #153, opened 2016-08-24
Thread safety issue


arjun27:
While doing a `put_item` in a table, we're frequently getting an key error on `endpoint_resolver`. The same key error can be seen for `data_loader` and `credential_provider`, but `endpoint_resolver` is far more common.

```
Traceback (most recent call last): 
File "./instrumentation/utils.py", line 69, in save_driver_log 
  driver_log.save() 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/models.py", line 379, in save 
  data = self._get_connection().put_item(*args, **kwargs) 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/connection/table.py", line 83, in put_item 
  return_item_collection_metrics=return_item_collection_metrics) 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/connection/base.py", line 806, in put_item 
  operation_kwargs.update(self.get_identifier_map(table_name, hash_key, range_key, key=ITEM)) 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/connection/base.py", line 648, in get_identifier_map 
  tbl = self.get_meta_table(table_name) 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/connection/base.py", line 399, in get_meta_table 
  data = self.dispatch(DESCRIBE_TABLE, operation_kwargs) 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/connection/base.py", line 235, in dispatch 
  data = self._make_api_call(operation_name, operation_kwargs) 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/connection/base.py", line 250, in _make_api_call 
  operation_model = self.client._service_model.operation_model(operation_name) 
File "/app/.heroku/python/lib/python2.7/site-packages/pynamodb/connection/base.py", line 387, in client 
  self._client = self.session.create_client(SERVICE_NAME, self.region, endpoint_url=self.host) 
File "/app/.heroku/python/lib/python2.7/site-packages/botocore/session.py", line 809, in create_client 
  endpoint_resolver = self.get_component('endpoint_resolver') 
File "/app/.heroku/python/lib/python2.7/site-packages/botocore/session.py", line 690, in get_component 
  return self._components.get_component(name) 
File "/app/.heroku/python/lib/python2.7/site-packages/botocore/session.py", line 875, in get_component 
    del self._deferred[name] 
[...]

PR #234, opened 2017-02-21
Make sessions threadsafe

AusIV:
I ran into the thread safety issue described here:

https://github.com/jlafon/PynamoDB/issues/153

According to boto3's docs, each thread should have its own session instance:

http://boto3.readthedocs.io/en/latest/guide/resources.html#multithreading

I resolved this by modifying the session property on the connection
class. It makes self._session into a thread local variable, and
attaches a session to self._session.session. The session property still
has the same behavior, but now it's creating a session on a
per-connection-per-thread basis instead of just on a per-thread basis.

This issue was intermittent to begin with, and I hadn't figured out a
way to reliably reproduce it, or I would have written a test. It still
passes PynamoDB's unit tests, and I haven't run into the thread safety
issue described in #153 since I've been running the patched version.


AusIV:
Not sure why the checks failed on Python 2.6. It doesn't look like it even executed the tests.

garrettheel:

victoreufrasio:
I've added a comment in issue #153 that shows how to reproduce this error.

jpinner-lyft:
replaced by #393

PR #250, opened 2017-03-28
Attribute List and Map support UTCDateTimeAttribute

appsolute:
I've used UTCDateTimeAttribute in MapAttribute and got this error.
Error: Unknown value: <type 'datetime.datetime'>

I've figure out that there is no datetime support in SERIALIZE_CLASS_MAP when serialize in Map or List.

Example
```python
from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute, MapAttribute, UTCDateTimeAttribute
from datetime import datetime

# define table
class SiteLDevice(MapAttribute):
	name = UnicodeAttribute(null=True)
	created = UTCDateTimeAttribute(null=False)

class Site(Model):
	"""
	A Site
	"""
	class Meta:
		table_name = "Site3"
		region = 'ap-southeast-1'
	
	name = UnicodeAttribute(range_key=True)
	server = UnicodeAttribute(hash_key=True)
	devices = ListAttribute(of=SiteLDevice)

# create table
Site.create_table(read_capacity_units=1, write_capacity_units=1)

# create new site
site = Site()
site.name = 'test'
site.server = 'testserver.abc.com'
site.devices = [ SiteLDevice(name='device 1', created= datetime.now()) ]
site.save()
```

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -11,6 +11,7 @@ from pynamodb.constants import (
     MAP, MAP_SHORT, LIST, LIST_SHORT, DEFAULT_ENCODING, BOOLEAN, ATTR_TYPE_MAP, NUMBER_SHORT, NULL
 )
 import collections
+from datetime import datetime
 
 
 class Attribute(object):
@@ -601,6 +602,7 @@ SERIALIZE_CLASS_MAP = {
     float: NumberAttribute(),
     int: NumberAttribute(),
     str: UnicodeAttribute(),
+    datetime: UTCDateTimeAttribute(),
 }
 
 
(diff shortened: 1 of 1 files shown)


yedpodtrzitko:
tests pls

appsolute:
Could you please guide me which file which suit for this test case?

I need this test case to save/retrieve with Dynamodb to make sure it works.

jmphilli:
i'm not sure we want to include this. 

Dynamo does not support date time attributes natively which is why it's not in the mapattribute type map.
It would be nice to keep that type map clean. Could you override your model's save method and convert your `datetimes` to strings? Then you can just have a UnicodeAttribute on the model. (you would have to update your db reads to transform the string into a datetime as well)

jonathonlui:
I created an alternative solution to the `datetime` serialize issue as part of #254 

Relevant changes in https://github.com/pynamodb/PynamoDB/pull/254/commits/776c0cb75a42ee9be7443d9ca5b52d3af1fff070

PR #254, opened 2017-04-17
ListAttribute & MapAttribute serialize use correct class like deserialize

jonathonlui:
This PR updates `ListAttribute` so that it uses the `serialize` method from its type (if provided) and updates `MapAttribute` so that it uses the `serialize` methods of its attributes. This mirros the behavior in the `deserialize` methods.

This means using a `ListAttribute` with an `of` with a custom `Attribute` caused an unexpected behavior. For example:

``` python
from pynamodb.attributes import Attribute, UnicodeAttribute, MapAttribute, ListAttribute
from pynamodb.constants import STRING


class CustomAttribute(Attribute):
    attr_type = STRING

    def serialize(self, value):
        print('CustomAttribute.serialize')
        return value

    def deserialize(self, value):
        print('CustomAttribute.deserialize')
        return value


class DogAttribute(MapAttribute):
    name = UnicodeAttribute()
    custom = CustomAttribute()


if __name__ == '__main__':
    list_of_dogs = [
        DogAttribute(name='Fido', custom='test'),
        DogAttribute(name='Spot', custom='other-test'),
    ]
    list_attribute = ListAttribute(of=DogAttribute)
    ser = list_attribute.serialize(list_of_dogs)
    list_attribute.deserialize(ser)
```

Without this PR `CustomAttribute.serialize` is never called. The output is:

```
CustomAttribute.deserialize
CustomAttribute.deserialize
```

With this PR the output is:

```
CustomAttribute.serialize
CustomAttribute.serialize
CustomAttribute.deserialize
CustomAttribute.deserialize
```

### DIFF
--- a/pynamodb/tests/test_attributes.py
+++ b/pynamodb/tests/test_attributes.py
@@ -696,6 +696,7 @@ class MapAndListAttributeTestCase(TestCase):
         class Person(MapAttribute):
             name = UnicodeAttribute()
             age = NumberAttribute()
+            extra_arbitrary_data = JSONAttribute()
 
             def __lt__(self, other):
                 return self.name < other.name
@@ -707,10 +708,14 @@ class MapAndListAttributeTestCase(TestCase):
         person1 = Person()
         person1.name = 'john'
         person1.age = 40
+        person1.extra_arbitrary_data = {'married': True}
 
         person2 = Person()
         person2.name = 'Dana'
         person2.age = 41
+        person2.extra_arbitrary_data = {'net_income': 200000,
+                                        'dog': 'Fido'}
+
         inp = [person1, person2]
 
         list_attribute = ListAttribute(default=[], of=Person)
--- a/pynamodb/tests/test_attributes.py
+++ b/pynamodb/tests/test_attributes.py
@@ -641,6 +641,20 @@ class MapAttributeTestCase(TestCase):
         self.assertEqual(json.dumps({'map_attr': {'foo': 'bar'}}),
                          json.dumps(item.typed_map.as_dict()))
 
+    def test_serialize_datetime(self):
+
+        class CustomMapAttribute(MapAttribute):
+            date_attr = UTCDateTimeAttribute()
+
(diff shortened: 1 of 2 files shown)


jmphilli:
this looks good, i totally get what you're doing here.
can you add a test that looks a lot like the description here so that going forward we know why we made this choice

pynamodb/tests/test_attributes.py
```diff
@@ -707,17 +722,69 @@ def __eq__(self, other):
...
         list_attribute = ListAttribute(default=[], of=Person)
         serialized = list_attribute.serialize(inp)
         deserialized = list_attribute.deserialize(serialized)
         self.assertEqual(sorted(deserialized), sorted(inp))
 
+    def test_list_of_map_with_of_and_custom_attribute(self):
+
+        # Create a couple of mock functions to use
+        # to test that the CustomAttribute serialize/deserialize are called
+        serialize_mock = Mock()
```
garrettheel: An alternative way of doing this is by using `attribute1 = Mock(wraps=CustomMapAttribute())`. This will give you something that acts like a `CustomMapAttribute` but you can also inspect method calls in the usual mock way (i.e `attribute1.serialize.called`, etc.)

pynamodb/tests/test_attributes.py
```diff
@@ -707,17 +722,69 @@ def __eq__(self, other):
...
+        inp = [attribute1, attribute2]
+
+        list_attribute = ListAttribute(default=[], of=CustomMapAttribute)
+        serialized = list_attribute.serialize(inp)
+        deserialized = list_attribute.deserialize(serialized)
+        assert deserialized == inp
+
+        # Confirm that the the serialize/deserialize are called
+        # with the expected values
+        serialize_mock.assert_has_calls([
```
garrettheel: Unfortunately the version of mock we're using is pretty old and doesn't have methods like these. Can you rebase off https://github.com/pynamodb/PynamoDB/commit/4f56221d7860e3f27f129963a40f96034a3f14f3 so we can be sure these are passing? Thanks!

garrettheel:
Fixed in #286

PR #256, opened 2017-04-21
Consume all pages when counting

### LINKED ISSUE #243: count(...) is broken when there are too many items.
tl dr: The count function is broken when there are many items per hash key.  It is likely that it is returning the count for only the first page.  

I'm getting a different number from pynamo count than from using boto or from querying for all items and counting them locally:
```
>>> NumberPoolServiceIndex.count(hash_key='number_pool')
8528
>>> dynamodb_table.query_count(index=settings.DYNAMODB_NUMBER_POOL_TABLE_SERVICE_INDEX, s__eq='number_pool')
10175
>>> len(list(NumberPoolServiceIndex.query('number_pool')))
10175
```

Here is the log when getting the count:
```
>>> NumberPoolServiceIndex.count(hash_key='number_pool')
DEBUG:pynamodb.connection.base:Calling Query with arguments {'IndexName': 'switchboard-production-number-pool-service-index', 'ReturnConsumedCapacity': 'TOTAL', 'TableName': 'switchboard-production-number-pool', 'KeyConditions': {u's': {'ComparisonOperator': 'EQ', 'AttributeValueList': [{u'S': u'number_pool'}]}}, 'Select': 'COUNT'}
DEBUG:pynamodb.connection.base: Query consumed 73.0 units
8528
```

and here is the log for querying for all items:
```
>>> len(list(NumberPoolServiceIndex.query('number_pool')))
DEBUG:pynamodb.models:Fetching first query page
DEBUG:pynamodb.connection.base:Calling Query with arguments {'IndexName': 'switchboard-production-number-pool-service-index', 'TableName': 'switchboard-production-number-pool', 'ReturnConsumedCapacity': 'TOTAL', 'KeyConditions': {u's': {'ComparisonOperator': 'EQ', 'AttributeValueList': [{u'S': u'number_pool'}]}}}
DEBUG:pynamodb.connection.base: Query consumed 73.0 units
DEBUG:pynamodb.models:Fetching query page with exclusive start key: {u's': {u'S': u'number_pool'}, u'lu': {u'N': u'1489565278284'}, u'pn': {u'S': u'+15015749931'}}
DEBUG:pynamodb.connection.base:Calling Query with arguments {'ExclusiveStartKey': {u's': {u'S': u'number_pool'}, u'lu': {u'N': u'1489565278284'}, u'pn': {u'S': u'+15015749931'}}, 'IndexName': 'switchboard-production-number-pool-service-index', 'TableName': 'switchboard-production-number-pool', 'ReturnConsumedCapacity': 'TOTAL', 'KeyConditions': {u's': {'ComparisonOperator': 'EQ', 'AttributeValueList': [{u'S': u'number_pool'}]}}}
DEBUG:pynamodb.connection.base: Query consumed 14.5 units
10182
```

It appears that there are multiple pages and its quite likely that the count function is just returning the count for the first page.

garrettheel:
Fixes #243

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -542,16 +542,25 @@ class Model(AttributeContainer):
             non_key_attribute_classes=non_key_attribute_classes,
             filters=filters)
 
-        data = cls._get_connection().query(
-            hash_key,
-            index_name=index_name,
-            consistent_read=consistent_read,
-            key_conditions=key_conditions,
-            query_filters=query_filters,
-            limit=limit,
-            select=COUNT
-        )
-        return data.get(CAMEL_COUNT)
+        count_buffer = 0
+        last_evaluated_key = None
+        started = False
+        while not started or last_evaluated_key:
+            started = True
+            data = cls._get_connection().query(
+                hash_key,
+                index_name=index_name,
+                consistent_read=consistent_read,
+                key_conditions=key_conditions,
+                query_filters=query_filters,
+                exclusive_start_key=last_evaluated_key,
+                limit=limit,
+                select=COUNT
+            )
+            count_buffer += data.get(CAMEL_COUNT)
+            last_evaluated_key = data.get(LAST_EVALUATED_KEY, None)
+
+        return count_buffer
 
     @classmethod
     def query(cls,
--- a/pynamodb/tests/test_model.py
+++ b/pynamodb/tests/test_model.py
@@ -1477,6 +1477,40 @@ class ModelTestCase(TestCase):
             }
             deep_eq(args, params, _assert=True)
 
+    def test_index_multipage_count(self):
+        with patch(PATCH_METHOD) as req:
+            last_evaluated_key = {
+                'user_name': {'S': u'user'},
+                'user_id': {'S': '1234'},
+            }
+            req.side_effect = [
+                {'Count': 1000, 'LastEvaluatedKey': last_evaluated_key},
+                {'Count': 42}
+            ]
+            res = CustomAttrNameModel.uid_index.count('foo')
+            self.assertEqual(res, 1042)
+
+            args_one = req.call_args_list[0][0][1]
+            params_one = {
+                'KeyConditions': {
+                    'user_id': {
+                        'ComparisonOperator': 'EQ',
+                        'AttributeValueList': [{'S': u'foo'}]
+                    }
+                },
+                'IndexName': 'uid_index',
+                'TableName': 'CustomAttrModel',
+                'ReturnConsumedCapacity': 'TOTAL',
+                'Select': 'COUNT'
+            }
+
+            args_two = req.call_args_list[1][0][1]
+            params_two = copy.deepcopy(params_one)
+            params_two['ExclusiveStartKey'] = last_evaluated_key
+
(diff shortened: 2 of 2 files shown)


jmphilli:
tests?

lita:
can you add a test? 😹

pynamodb/tests/test_model.py
```diff
@@ -1477,6 +1477,40 @@ def test_index_count(self):
             }
             deep_eq(args, params, _assert=True)
 
+    def test_index_multipage_count(self):
```
lita: Can we also add a test for when the data is None?

pynamodb/tests/test_model.py
```diff
@@ -1477,6 +1477,40 @@ def test_index_count(self):
...
+            last_evaluated_key = {
+                'user_name': {'S': u'user'},
+                'user_id': {'S': '1234'},
+            }
+            req.side_effect = [
+                {'Count': 1000, 'LastEvaluatedKey': last_evaluated_key},
+                {'Count': 42}
+            ]
+            res = CustomAttrNameModel.uid_index.count('foo')
+            self.assertEqual(res, 1042)
```
lita: nit: 
```
{
    'Count': 42
}
```

lita:
looks good other than adding that extra test. +1

pynamodb/tests/test_model.py
garrettheel: Do you mean when `LastEvaluatedKey` is None on the first page? If so, this is covered by the previous test

pynamodb/models.py
```diff
@@ -542,16 +542,24 @@ def count(cls,
...
-            key_conditions=key_conditions,
-            query_filters=query_filters,
-            limit=limit,
-            select=COUNT
-        )
-        return data.get(CAMEL_COUNT)
+        count_buffer = 0
+        last_evaluated_key = None
+        data = None
+        while not data or last_evaluated_key:
```
mosoto: probably a stupid question but because I'm not aware of the output of query I'll ask it anyway - will query always return an object?  I'm wondering about the case where there are no items and I want to make sure this won't loop forever because of that "not data".

garrettheel: The most minimal object that should be returned is `{'Count': 0}`, which should be truth-y here. This is a good point though, so I replaced this check with an explicit `started` bool to be more explicit.

mosoto:
+1

pynamodb/models.py
```diff
@@ -542,16 +542,25 @@ def count(cls,
...
-            key_conditions=key_conditions,
-            query_filters=query_filters,
-            limit=limit,
-            select=COUNT
-        )
-        return data.get(CAMEL_COUNT)
+        count_buffer = 0
+        last_evaluated_key = None
+        started = False
+        while not started or last_evaluated_key:
```
jmphilli: can't you save the `started` var if you just set `last_evaluated_key` to a sentinel value?

garrettheel: Yeah, but I see this as a little hacky if there are two "null" states

pynamodb/tests/test_model.py
lita: Ah, yup. you are right.

PR #257, opened 2017-04-21
Use pytest to run unit tests

lita:
This is the start of moving off of `TestCase`


tox.ini
```diff
@@ -6,4 +6,3 @@ envlist = py26,py27,py33,py34,pypy
 
 [testenv]
 deps = -rrequirements.txt
-commands = nosetests
```
garrettheel: we'll wanna replace this with the `py.test` command

requirements.txt
```diff
@@ -3,8 +3,9 @@ mock==1.0.1
 Delorean==0.4.5
 botocore==1.1.2
 flake8==2.4.1
-nose==1.3.7
 requests==2.7.0
 six==1.8.0
 tox==2.1.1
 python-coveralls==2.5.0
+pytest==3.0.7
+pytest-cov==2.4.0
```
yedpodtrzitko: since these are not production requirements, it would be better to put them into a separate file, eg.:

requirements-dev.txt
```
-rrequirements.txt
pytest==3.0.7
pytest-cov==2.4.0
```

...and then use this file below in `tox.ini` as a dependecies file

lita: You make a good point. I want to make that change should be a separate PR rather than put it in this one to keep the history clean.

tox.ini
lita: fixed.

PR #258, opened 2017-04-23
Fix MapAttribute.__getattr__ raising wrong exception

jonathonlui:
`MapAttribute.__getattr__` incorrectly raises `KeyError` when an attribute does not exist. 

From [Python lang ref]( https://docs.python.org/3/reference/datamodel.html#object.__getattr__):
> `object.__getattr__(self, name)`
> ... should return the (computed) attribute value or raise an `AttributeError` exception.

This PR fixes `MapAttribute.__getattr__` so that it raises the correct `AttributeError` when the attribute does not exist.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -434,7 +434,11 @@ class MapAttribute(AttributeContainer, Attribute):
         return self.attribute_values[item]
 
     def __getattr__(self, attr):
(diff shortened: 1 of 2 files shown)


garrettheel:

jpinner-lyft:
Merged as part of https://github.com/pynamodb/PynamoDB/pull/292

PR #264, opened 2017-04-30
Remove unittest.TestCase in model unit tests

lita:


lita:
I parametrize `test_save` but wasn't sure if that is how you all wanted to go about doing it. If so, I can parametrize the other tests as well.

ikonst:
@lita I suppose this is now abandoned?

PR #265, opened 2017-05-02
Separate Test Requirements from Production Requirements

lita:


requirements.txt
```diff
@@ -1,10 +1,4 @@
 sphinx-rtd-theme==0.1.8
```
garrettheel: I'd consider `sphinx-rtd-theme` and `requests` to be dev-requirements here too. We use the vendored `requests` from botocore in pynamo and only rely on the installed version for testing currently

PR #267, opened 2017-05-04
Support rate_limited_scan against DynamoDBLocal

### LINKED ISSUE #266: rate_limited_scan fails with DynamoDBLocal
DynamoDB local (http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) does not provide the ability to return consumed capacity even when requested. If you request it, the request is accepted as valid but the information is not returned.

I confirmed this in a testing environment, and it has been independently noted by https://github.com/mhart/dynalite

Attempts to use this call fail due to:

```
                   for item in data.get(ITEMS):
                        yield item
    
                        if limit is not None:
                            limit -= 1
                            if not limit:
                                return
    
>                   latest_scan_consumed_capacity = data.get(CONSUMED_CAPACITY).get(CAPACITY_UNITS)
E                   AttributeError: 'NoneType' object has no attribute 'get'
```

scode:
This allows `rate_limited_scan` to operate correctly when run against
a DynamoDBLocal instance (or other implementation) rather than real DynamoDB.

Per discussion with @garrettheel we're preferring to make it a configuration option
such that we never risk accidentally allowing the non-rate limited fallback in production.

Fixes #266


pynamodb/connection/base.py
```diff
@@ -1020,7 +1020,15 @@ def rate_limited_scan(self,
                             if not limit:
                                 return
 
-                    latest_scan_consumed_capacity = data.get(CONSUMED_CAPACITY).get(CAPACITY_UNITS)
+                    if CONSUMED_CAPACITY in data:
```
scode: I just want to call out that I think the most concerning problem with this is that if for whatever reason a real production use case ends up lacking this information (bug in client or server) we'll just happily truck on with effectively silent failure rather than explicitly failing.

scode: (this comment no longer applies in this version)

docs/local.rst
```diff
@@ -10,7 +10,8 @@ PynamoDB has been tested with two DynamoDB compatible servers, `DynamoDB Local <
 and `dynalite <https://github.com/mhart/dynalite>`_.
 
 All you need to do is set the ``host`` attribute on your ``Model``'s ``Meta`` class to the hostname and port
-that your server is listening on.
+that your server is listening on. If you use `rate_limited_scan` on your models with DymamoDB, you must also
+set `allow_rate_limited_scan_without_consumed_capacity` to `True` in the `settings file <settings.rst>`_.
```
garrettheel: Nit: it may be better to link to the specific setting on the page rather than the page as a whole.

Also note that you'll need to use double backticks for code blocks in rst

docs/settings.rst
```diff
@@ -52,6 +52,21 @@ to DynamoDB.
...
+Default: ``False``
+
+If ``True``, ``rate_limited_scan()`` will proceed silently (without
+rate limiting) if the DynamoDB server does not return consumed
+capacity information. This can be useful when running tests against an
+implementation that does not support returning consumed capacity
+information, such as `DynamoDB Local
+<http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html>`_.
+
+You may wish to consider use of `dynalite
```
garrettheel: Can we include this recommendation (and a note about this setting) in `local.rst` instead?

docs/settings.rst
```diff
@@ -52,6 +52,21 @@ to DynamoDB.
...
+<https://github.com/mhart/dynalite>`_ as an alterantive to DynamoDB
```
garrettheel: `s/alterantive/alternative`

pynamodb/connection/base.py
```diff
@@ -1020,7 +1027,19 @@ def rate_limited_scan(self,
...
+                        latest_scan_consumed_capacity = data.get(CONSUMED_CAPACITY).get(CAPACITY_UNITS)
+                    else:
+                        if allow_rate_limited_scan_without_consumed_capacity:
+                            latest_scan_consumed_capacity = 0
+                        else:
+                            raise ScanError('Rate limited scan not possible because the server did not send back'
```
garrettheel: I think the last two sentences are probably overkill here. Anyone interested enough in this will read the settings and find all of the information they need.

garrettheel: Also `s/anwyay/anyway`

scode:
@garrettheel I believe I have addressed the feedback.

docs/local.rst
```diff
@@ -9,8 +9,11 @@ used with any server that provides the same API as DynamoDB.
 PynamoDB has been tested with two DynamoDB compatible servers, `DynamoDB Local <http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Tools.DynamoDBLocal.html>`_
 and `dynalite <https://github.com/mhart/dynalite>`_.
 
-All you need to do is set the ``host`` attribute on your ``Model``'s ``Meta`` class to the hostname and port
-that your server is listening on.
+To use a local server, you need to set the ``host`` attribute on your ``Model``'s ``Meta`` class to the hostname and port
+that your server is listening on. If you are using DynamoDB Local and also use ``rate_limited_scan`` on your models, you must also
```
garrettheel: Small thing, can you use a `note` directive for the new paragraph? See https://raw.githubusercontent.com/rtfd/readthedocs.org/master/docs/features/embed.rst for an example

garrettheel:
Thanks for the fix @scode!

PR #268, opened 2017-05-05
Expose a low level `update_expression` api

jmphilli:
This allows us to call complicated `update_expressions` with names and values.
 - [ ] Add Tests to table + base connection
 - [ ] Clean up code


jmphilli:
@garrettheel @lita

Issue #241, opened 2017-03-07
Nullable attribute of MapAttribute saved with value None is deserialized to True


Lizards:
A nullable value on a MapAttribute set to None gets deserialized to True when the item is retrieved from DynamoDB.  Below are sample models and some output showing more detail.

Is this a bug or am I misunderstanding?

Models:
```python
class NullableDemo(Model):
    """ OK: bar attribute can be None """
    class Meta:
        table_name = 'NullableDemo'
        host = 'http://172.17.0.2:8000'

    name = UnicodeAttribute(hash_key=True)
    foo = UnicodeAttribute(default='foo')
    bar = UnicodeAttribute(null=True)


class NullableDemoAttribute(MapAttribute):
    foo = UnicodeAttribute(default='foo')
    bar = UnicodeAttribute(null=True)


class NullableDemoWithMap(Model):
    """ Not OK: custom_map.bar deserializes to True when saved as None """
    class Meta:
        table_name = 'NullableDemoWithMap'
        host = 'http://172.17.0.2:8000'

    name = UnicodeAttribute(hash_key=True)
    custom_map = NullableDemoAttribute()`
```

Here's shell output showing the attribute values on the model instances as well as the raw output from DynamoDB via boto3:
```
In [1]: import boto3
   ...: from test import NullableDemo, NullableDemoWithMap, NullableDemoAttribute
   ...: dynamodb = boto3.client('dynamodb', region_name='us-east-1', endpoint_url='http://172.17.0.2:8000')
   ...: 

In [2]: nullable_ok = NullableDemo('demo')
   ...: nullable_ok.save()
   ...: nullable_ok = NullableDemo.get('demo')
   ...: nullable_ok.foo, nullable_ok.bar
   ...: 
Out[2]: (u'foo', None)

In [3]: dynamodb.scan(TableName='NullableDemo')['Items']
   ...: 
Out[3]: [{u'foo': {u'S': u'foo'}, u'name': {u'S': u'demo'}}]

In [4]: nullable_ok.foo = 'updated foo'
   ...: nullable_ok.bar = None
   ...: nullable_ok.save()
   ...: nullable_ok = NullableDemo.get('demo')
   ...: nullable_ok.foo, nullable_ok.bar
   ...: 
Out[4]: (u'updated foo', None)

In [5]: dynamodb.scan(TableName='NullableDemo')['Items']
[...]

PR #272, opened 2017-05-09
Fix for null deserialize

jmphilli:
https://github.com/pynamodb/PynamoDB/issues/241 -- fix for
cc @garrettheel @lita

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -519,7 +519,7 @@ class MapAttribute(AttributeContainer, Attribute):
 
 def _get_value_for_deserialize(value):
     key = list(value.keys())[0]
-    if key is NULL:
+    if key == NULL:
         return None
     return value[list(value.keys())[0]]
 
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -518,6 +518,9 @@ class MapAttribute(AttributeContainer, Attribute):
         return _get_class_for_deserialize(value)
 
 def _get_value_for_deserialize(value):
+    key = list(value.keys())[0]
+    if key is NULL:
+        return None
     return value[list(value.keys())[0]]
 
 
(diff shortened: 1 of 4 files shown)


pynamodb/attributes.py
```diff
@@ -518,6 +518,9 @@ def _get_deserialize_class(cls, key, value):
         return _get_class_for_deserialize(value)
 
 def _get_value_for_deserialize(value):
+    key = list(value.keys())[0]
+    if key is NULL:
```
omrishaked: Did you mean `if key == 'NULL':`?

lita: I believe `key is NULL` should work as the strings are primitives.

omrishaked: Should we add a test for this?

pynamodb/attributes.py
```diff
@@ -517,8 +520,12 @@ def _get_deserialize_class(cls, key, value):
             return cls._get_attributes().get(key)
         return _get_class_for_deserialize(value)
 
+
 def _get_value_for_deserialize(value):
-    return value[list(value.keys())[0]]
+    key = list(value.keys())[0]
```
danielhochman: i think `next(iter(value.keys()))` is a better option

garrettheel: `six.iterkeys` if we want to support py2 and py3. I don't think performance is a real concern here though, as this always only has one element.

pynamodb/attributes.py
```diff
@@ -493,8 +493,11 @@ def deserialize(self, values):
             attr_value = _get_value_for_deserialize(v)
             key = self._dynamo_to_python_attr(k)
             attr_class = self._get_deserialize_class(key, v)
+            deserialized_value = None
```
lita: since you know the attr_value is gonna be null in line 493, it is easier to read by doing

```
if not attr_value:
    continue
```

lita: nvm i see you need to add the key in there.

jmphilli: i need to set it tho

lita: yea yea, i see that now. sorry about that.

PR #278, opened 2017-05-11
Add support for signals via blinker

lita:


pynamodb/tests/test_signals.py
```diff
@@ -0,0 +1,50 @@
...
+
+@mock.patch('pynamodb.connection.Connection.session')
+@mock.patch('pynamodb.connection.Connection.requests_session')
+@mock.patch('pynamodb.connection.base.uuid')
+def test_signal(mock_uuid, requests_session_mock, session_mock):
+    pre_recorded = []
+    post_recorded = []
+    UUID = '123-abc'
+
+    def record_pre_boto_send(sender, operation_name, table_name, req_uuid):
```
lita: I have it set up to pass in the Connection class as the emitter, which was the simplest way to implement this.

Unfortunately, the Connection class does give away a lot of information. Also, it is not very intuitive that you have to have this as an argument.

I have another implementation where we have a wrapper around the `pre_boto_send.connect` call so you don't have to handle the sender, but it requires more bookkeeping and the code got complicated. thought I push the simple change for now and see what other people thought.

pynamodb/connection/signals.py
```diff
@@ -0,0 +1,44 @@
...
+    from blinker import Namespace
+
+    signals_available = True
+except ImportError:
+    class Namespace(object):
+        def signal(self, name, doc=None):
+            return _FakeSignal(name, doc)
+
+
+    class _FakeSignal(object):
```
lita: I made installing blinker optional, as this is what Flask and MongoEngine do as well.

pynamodb/tests/test_signals.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def record_post_boto_send(sender, operation_name, table_name, req_uuid):
+        post_recorded.append((operation_name, table_name, req_uuid))
+
+    good_response_content = {'TableDescription': {'TableName': 'table', 'TableStatus': 'Creating'}}
+    good_response = requests.Response()
+    good_response.status_code = 200
+    good_response._content = json.dumps(good_response_content).encode('utf-8')
+
+    pre_boto_send.connect(record_pre_boto_send)
```
rowillia: Can you move the wiring up/removing of the signals to a setup/teardown method so others don't have to do the `try/finally` dance?

pynamodb/connection/base.py
```diff
@@ -281,20 +283,23 @@ def _make_api_call(self, operation_name, operation_kwargs):
...
-
+        table_name = operation_kwargs.get(TABLE_NAME)
         prepared_request = self._create_prepared_request(request_dict, operation_model)
 
         for i in range(0, self._max_retry_attempts_exception + 1):
             attempt_number = i + 1
             is_last_attempt_for_exceptions = i == self._max_retry_attempts_exception
 
             try:
+                req_uuid = uuid.uuid4()
```
rowillia: IMHO we should move the hooks to surround linke 266 above so there is only one event fired even if there are retries and so we don't have to worry about a receiver throwing a `ValueError` and having pynamo interpret it as a failure it should retry.

pynamodb/tests/test_signals.py
lita: yup

pynamodb/connection/base.py
lita: done

pynamodb/tests/test_signals.py
lita: So I tried to do this and the issue is that you have to pass in the method `record_pre_boto_send` when you connect/disconnect. so there isn't a clean way to do this in setup/teardown methods unless you set some global state or move the callbacks into fixtures. Feel like that is over engineering for now. 

Also Flask tests it this way: https://github.com/pallets/flask/blob/master/tests/test_signals.py#L49-L71

pynamodb/connection/base.py
```diff
@@ -261,7 +263,11 @@ def dispatch(self, operation_name, operation_kwargs):
                 operation_kwargs.update(self.get_consumed_capacity_map(TOTAL))
         self._log_debug(operation_name, operation_kwargs)
 
+        table_name = operation_kwargs.get(TABLE_NAME)
+        req_uuid = uuid.uuid4()
+        pre_boto_send.send(self, operation_name=operation_name, table_name=table_name, req_uuid=req_uuid)
```
lita: Are we cool with not catching exceptions if the send fails?

pynamodb/connection/base.py
```diff
@@ -270,6 +277,18 @@ def dispatch(self, operation_name, operation_kwargs):
             log.debug("%s %s consumed %s units",  data.get(TABLE_NAME, ''), operation_name, capacity)
         return data
 
+    def send_post_boto_callback(self, operation_name, req_uuid, table_name):
+        try:
+            post_boto_send.send(self, operation_name=operation_name, table_name=table_name, req_uuid=req_uuid)
```
garrettheel: nit: `s/boto/boto3`

pynamodb/connection/base.py
```diff
@@ -270,6 +277,18 @@ def dispatch(self, operation_name, operation_kwargs):
...
+        except Exception as e:
+            log.error("post_boto callback threw an exception. {0}".format(e))
```
garrettheel: Any reason not to do `log.exception` here instead?

pynamodb/tests/test_signals.py
```diff
@@ -0,0 +1,106 @@
...
+from pynamodb.connection.signals import _FakeNamespace
+from pynamodb.connection.signals import pre_boto_send, post_boto_send
+
+try:
+    import blinker
+except ImportError:
+    blinker = None
+
+
+pytestmark = pytest.mark.skipif(
```
garrettheel: Why would we support running tests without `blinker` available? Since we control the requirements, this doesn't need to be a concern. We should instead additionally test the behavior where the system _thinks_ blinker is not installed to ensure that works as expected

pynamodb/connection/base.py
garrettheel: How about `request_uuid` for consistency?

danielhochman: i would remove boto from the signal names altogether. it's not unlikely that in the future that we will remove boto(core) entirely.

pynamodb/connection/signals.py
```diff
@@ -0,0 +1,50 @@
+"""
```
danielhochman: why is this file under connection/ ?

lita: Because they are being used in connection/base.py and not with the higher level api calls, I thought it made sense to put it in there. That way when you are registering it, you know you are registering with a lower level connection call.

But if you feel like it makes more sense to move it to the main module, I can move it. I went back and forth for awhile, so could go either way.

pynamodb/connection/base.py
lita: sure, I can update that.

pynamodb/connection/base.py
lita: yea, I can do that instead.

pynamodb/tests/test_signals.py
lita: Do you know of a good way to test the behavior without blinker being installed? I have tried to mock out `import` through `__builtins__` and `sys.modules` with no luck. From what it looks like, Python caches the modules. I have also tried using the `reload` function, but that didn't work either.

If you have ideas, please let me know.

pynamodb/tests/test_signals.py
```diff
@@ -0,0 +1,106 @@
...
+        pre_dynamo_send.disconnect(record_pre_dynamo_send)
+        post_dynamo_send.disconnect(record_post_dynamo_send)
+
+
+def test_fake_signals():
+    _signals = _FakeNamespace()
+    pre_dynamo_send = _signals.signal('pre_dynamo_send')
+    with pytest.raises(RuntimeError):
+        pre_dynamo_send.connect(lambda x: x)
+    pre_dynamo_send.send(object, operation_name="UPDATE", table_name="TEST", req_uuid="something")
```
lita: @garrettheel i test the case where blinker is not installed this way. Where I just use the _FakeNamespace and make sure no exceptions happen when we call `send` but when you try to connect to it, it causes a runtime error.

pynamodb/tests/test_signals.py
lita: I can remove the pytestmark

jmphilli:
-_-

docs/signals.rst
```diff
@@ -0,0 +1,43 @@
+Signals
+=======
+Starting with Pynamo 4.0.0, there is support for signalling in
```
lita: yea this version might not be true.

docs/signals.rst
```diff
@@ -0,0 +1,41 @@
+Signals
+=======
+Starting with Pynamo 4.0.0, there is support for signalling in Pynamo.  This support is provided by the excellent `blinker`_ library and will gracefully fall back if it is not available. `blinker`_  is not installed by default, as we didn't want to additional overhead to network calls if it wasn't needed.
```
lita: To update the version number later.

docs/signals.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Signals allow certain senders to notify subscribers that something happened. PynamoDB only sends signals before and after every DynamoDB API call.
```
garrettheel: s/only/currently ?

docs/signals.rst
```diff
@@ -0,0 +1,42 @@
...
+
+We purposely only send the operation name and the table name to the callback, as we want to prevent users from doing business logic in the signal. Keep in mind that signals are intended to notify subscribers. Subscribers should not try to modify data or do database calls within the callback. Also, adding too much logic in the callback will degrade the performance of your database writes.
```
garrettheel: You could slap this in a `.. note::` and rephrase it to something along the lines of: "it is recommended not to implement business logic in signals" and give a concrete reason why.

docs/signals.rst
```diff
@@ -0,0 +1,42 @@
...
+
+
+Subscribing to Signals
+----------------------
+
+We fire `pre_dynamo_send` before the network call and `post_dynamo_send` after the network call to DynmaoDB.
```
garrettheel: It's a little awkward to use first-person in the writing here. Maybe something like:

```
The signals currently supported are:
 * `pre_dynamo_send` - ...
 * `post_dynamo_send` - ...
```

Also `s/DynmaoDB/DynamoDB`

docs/signals.rst
```diff
@@ -0,0 +1,42 @@
...
+================  ===========
+*sender*          The object that fired that method.
+*operation_name*  The string name of the `Dynamo action`_
+*table_name*      The name of the table the operation is called upon.
+*req_uuid*        A unique identifer so subscribers can correlate the before and after events.
+================  ===========
+
+To subscribe to a signal, just need to import the signal object and connect your callback, like so.
+
+.. code:: python
```
garrettheel: Can we use the terminology `dynamodb` instead of `dynamo` everywhere in this PR? I did a pass through the API docs and we pretty much always include the full name in anything external-facing. e.g http://pynamodb.readthedocs.io/en/latest/api.html

docs/signals.rst
```diff
@@ -0,0 +1,42 @@
...
+        pre_recorded.append((operation_name, table_name, req_uuid))
+
+    def record_post_dynamo_send(sender, operation_name, table_name, req_uuid):
+        post_recorded.append((operation_name, table_name, req_uuid))
+
+    pre_dynamo_send.connect(record_pre_dynamo_send)
+    post_dynamo_send.connect(record_post_dynamo_send)
+ 
+.. _blinker: https://pypi.python.org/pypi/blinker
+.. _Dynamo action: https://github.com/pynamodb/PynamoDB/blob/cd705cc4e0e3dd365c7e0773f6bc02fe071a0631/
```
garrettheel: This SHA doesn't exist

docs/signals.rst
```diff
@@ -0,0 +1,42 @@
...
+================  ===========
+Arguments         Description
+================  ===========
+*sender*          The object that fired that method.
+*operation_name*  The string name of the `Dynamo action`_
+*table_name*      The name of the table the operation is called upon.
+*req_uuid*        A unique identifer so subscribers can correlate the before and after events.
+================  ===========
+
+To subscribe to a signal, just need to import the signal object and connect your callback, like so.
```
garrettheel: grammar here: `, just need to`

requirements-dev.txt
```diff
@@ -4,6 +4,7 @@ pytest==3.0.7
 pytest-cov==2.4.0
 python-coveralls==2.5.0
 tox==2.1.1
+blinker==1.4
```
garrettheel: Where did we settle on adding this as a feature to `setup.py`?

lita: I don't want to add this to setup.py and want blinker to be optional. This should be an optional feature and the user would need to install blinker separately. Thus, we have that try/catch around the import.

Issue #22, opened 2014-05-18
create_table should read the capacity from the meta class


ComboStyle:
by this example:
Thread.create_table(read_capacity_units=1, write_capacity_units=1, wait=True)

it would be easier to just do:
Thread.create_table(wait=True)

and the capacity units will be read from the model's meta like they do when creating an Index

PR #280, opened 2017-05-15
Fix attribute discovery with Model inheritance

garrettheel:
There was a regression with attribute discovery in https://github.com/pynamodb/PynamoDB/pull/22, where we switched from using `dir(cls)` to `cls.__dict__` to discover attributes. Turns out this doesn't include properties from base classes.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -77,8 +77,18 @@ class AttributeContainer(object):
         """
         cls._attributes = {}
         cls._dynamo_to_python_attrs = {}
-        for item_name, instance in six.iteritems(cls.__dict__):
-            if issubclass(type(instance), (Attribute,)):
+
+        for item_name in dir(cls):
+            try:
+                item_cls = getattr(getattr(cls, item_name), "__class__", None)
+            except AttributeError:
+                continue
+
+            if item_cls is None:
+                continue
+
+            if issubclass(item_cls, (Attribute, )):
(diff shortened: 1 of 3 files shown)


pynamodb/attributes.py
```diff
@@ -77,8 +77,18 @@ def _initialize_attributes(cls):
         """
         cls._attributes = {}
         cls._dynamo_to_python_attrs = {}
-        for item_name, instance in six.iteritems(cls.__dict__):
-            if issubclass(type(instance), (Attribute,)):
+
+        for item_name in dir(cls):
```
lita: is there a reason why we don't use `inspect` to do this instead?

garrettheel: I haven't done enough research to give a definitive reason why, but it seems _more_ hacky to do things this way than the current

lita: For readability, I prefer inspect.getmembers. I believe they are almost the same thing. Only difference is getmembers doesn't return metaclass attributes. 

The only downside I see is you have to import the inspect module.

garrettheel: Discussed offline that we'd stick with using `dir` for now (what we've historically always used) and reconsider alternate approaches later

mxr: @lita `inspect` could have performance implications (even though its API is nicer). Not saying this will happen here but something to watch out for. See http://go.lyft.com/optimizing-python

Issue #261, opened 2017-04-27
Why MapAttribute is required for ListAttribute?


leejaedus:
I think `ListAttribute` could get common`Attribute` like `UnicodeAttribute`
But it couldn't.

On pynamodb/attributes.py#L560-563
```python
        if of:
            if not issubclass(of, MapAttribute):
                raise ValueError("'of' must be subclass of MapAttribute")
            self.element_type = of
```

And so on, in this document section, http://pynamodb.readthedocs.io/en/latest/attributes.html#list-attributes

```python
from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute

class Office(Model):
    class Meta:
        table_name = 'OfficeModel'
    office_id = NumberAttribute(hash_key=True)
    employees = ListAttribute(of=UnicodeAttribute)
```

I wanna know what a right intent is. Thanks

gliv:
I'm running into the same issue and would be happy to see an example around `ListAttribute` that I could play with. 

Additionally, if I don't enforce the type with "of" (which should be possible according to the docs), I get a `Invalid attribute value type` (PutError, ValidationException). So obviously ListAttribute doesn't seem to work the way I would imagine (which is totally possible ;-) ). 

For something called ListAttribute, like in the example:

```
class Office(Model):
    class Meta:
        table_name = 'OfficeModel'
    office_id = NumberAttribute(hash_key=True)
    employees = ListAttribute(of=UnicodeAttribute)
```
Since the docs don't tell me otherwise, I would expect to be able to use it like a list, maybe something like:

```
ny_office = Office(3, employees=['John', 'Mike', 'Frank'])
ny_office.save()
```
Unfortunately that's not working for me. And probably I am missing something.

@jonathonlui and @jmphilli you looked into de/serializing with the right attribute type in PR [#254](https://github.com/pynamodb/PynamoDB/pull/254), you seem savvy on this, maybe you guys can help us out?

jmphilli:
Ya this is my bad. Let me take a look

leejaedus:
Could you give me a chance to fix that?

#281

jmphilli:
`of` is not required.
the doc is incorrect.

[here](https://github.com/pynamodb/PynamoDB/blob/0b3349d5d2b1b646c2889524516624eb51a01249/pynamodb/tests/test_model.py#L3001) is a raw (untyped) list attribute example.

[...]

PR #281, opened 2017-05-21
Remove ListAttribute.element_type restriction

leejaedus:
From this article <http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html>

At List Section,
>A list type attribute can store an ordered collection of values. Lists are enclosed in square brackets: [ ... ]
>
>A list is similar to a JSON array. There are no restrictions on the data types that can be stored in a >list element, and the elements in a list element do not have to be of the same type.

So I think the ListAttribute.element_type restriction should be removed.

#261

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -585,8 +585,6 @@ class ListAttribute(Attribute):
                                             default=default,
                                             attr_name=attr_name)
         if of:
-            if not issubclass(of, MapAttribute):
-                raise ValueError("'of' must be subclass of MapAttribute")
             self.element_type = of
 
     def serialize(self, values):


leejaedus:
And I think that it needs testcase

pynamodb/attributes.py
```diff
@@ -585,8 +585,6 @@ def __init__(self, hash_key=False, range_key=False, null=None, default=None, att
                                             default=default,
                                             attr_name=attr_name)
         if of:
-            if not issubclass(of, MapAttribute):
```
jmphilli: mmm this isn't right. `of` is not required.
the doc is incorrect.

[here](https://github.com/pynamodb/PynamoDB/blob/0b3349d5d2b1b646c2889524516624eb51a01249/pynamodb/tests/test_model.py#L3001) is a raw (untyped) list attribute example.

I updated the docs [here](https://github.com/pynamodb/PynamoDB/pull/282)

jmphilli: close in favor of a new one? or you wanna do with this branch?

leejaedus: close and I'll pick new one

PR #285, opened 2017-05-30
Added UTCDateTime to SERIALIZE_KEY_MAP and SERIALIZE_CLASS_MAP

phil-hachey:
Fixes issue where a UTCDateTime type can't be used with attributes


jcbertin:
This solution was already submitted and rejected in PR #250. An Alternative solution is proposed in PR #286 which is a rebased version of @jonathonlui's PR #254.

phil-hachey:
OK thanks!

PR #286, opened 2017-05-30
Ensure correct serialize is called for List and Map attributes

jcbertin:
Fixed PR #254 as asked by @garrettheel.

### DIFF
--- a/pynamodb/tests/test_attributes.py
+++ b/pynamodb/tests/test_attributes.py
@@ -10,7 +10,7 @@ from datetime import datetime
 from dateutil.parser import parse
 from dateutil.tz import tzutc
 
-from mock import patch
+from mock import patch, Mock, call
 
 from pynamodb.compat import CompatTestCase as TestCase
 from pynamodb.constants import UTC, DATETIME_FORMAT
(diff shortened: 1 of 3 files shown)


pynamodb/tests/test_attributes.py
```diff
@@ -734,17 +749,72 @@ def __eq__(self, other):
...
         list_attribute = ListAttribute(default=[], of=Person)
         serialized = list_attribute.serialize(inp)
         deserialized = list_attribute.deserialize(serialized)
         self.assertEqual(sorted(deserialized), sorted(inp))
 
+    def test_list_of_map_with_of_and_custom_attribute(self):
+
+        # Create a couple of mock functions to use
+        # to test that the CustomAttribute serialize/deserialize are called
+        serialize_mock = Mock()
```
garrettheel: Did you want to have a crack at the approach I mentioned in https://github.com/pynamodb/PynamoDB/pull/254 which simplifies this logic? I'm happy to help out if you need a hand, I know `mock` can get pretty confusing!

pynamodb/tests/test_attributes.py
```diff
@@ -723,6 +737,7 @@ def test_list_of_map_with_of(self):
         class Person(MapAttribute):
             name = UnicodeAttribute()
             age = NumberAttribute()
+            extra_arbitrary_data = JSONAttribute()
```
jcbertin: Yes, this would failed before the PR: when the Person is serialized, the extra_arbitrary_data  attribute was mapped to a MapAttribute instead of JSONAttribute because _get_class_for_serialize() returns MapAttribute class for a dict.

pynamodb/tests/test_attributes.py
jcbertin: The problem here is that we should wrap the custom attribute in CustomMapAttribute class, but then the __lt__ method would fail because Mock objects are unorderable type.

pynamodb/tests/test_attributes.py
garrettheel: Gotcha, makes sense. I think I'd prefer to see an individual test case for this to make it clear that's what we're testing for too, but this can stay as a higher-level test

jcbertin: Created correct MapAttribute serialize test.

pynamodb/tests/test_attributes.py
garrettheel: In starting to rewrite this test I realised that there are a bunch of changes I want to make to how tests work to improve this, so happy to merge this for now and clean it up there

PR #292, opened 2017-06-07
Key attribute_values off the field name instead of attr_name.

johnliu:
@garrettheel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,4 @@ script:
 
 after_success:
   - coveralls
+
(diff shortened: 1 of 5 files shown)


johnliu:
@jmphilli @garrettheel any clue what the failure is? looks like a dependency failure?

garrettheel:
@johnliu this is an annoying bug with that specific version of python and setup.py for six. Changing https://github.com/pynamodb/PynamoDB/blob/master/.travis.yml#L14 to `1.9.0` should fix it

garrettheel:
Thanks for fixing this!

Issue #249, opened 2017-03-28
true / false as string in UnicodeSetAttribute is returned as True and False boolean


jl-DaDar:
just found this problem in UnicodeSetAttribute. true and false string stored in the set is converted to boolean...

jmphilli:
https://github.com/pynamodb/PynamoDB/pull/294
 looks like it works to me -- can you write a test that breaks so i can see what you mean?

jmphilli:
nvm found it

jmphilli:
the `3.0.1` release fixes this. can you try that version out in a dev environment and see if it works for you? be warned there are breaking changes

PR #294, opened 2017-06-09
Deprecate json decode functionality in `UnicodeStringSet`

jmphilli:
https://github.com/pynamodb/PynamoDB/issues/249


pynamodb/attributes.py
```diff
@@ -227,20 +227,10 @@ def element_serialize(self, value):
 
     def element_deserialize(self, value):
         """
-        This deserializes what we get from mongo back into a str
-        Serialization previously json encoded strings. This caused them to have
-        extra double quote (") characters. That no longer happens.
-        This method allows both types of serialized values to be read
         :param value:
```
danielhochman: you could delete this docstring since it provides no value

PR #297, opened 2017-06-09
Enforce attr access through field name instead of attr_name

garrettheel:
Alternate approach to #292.

Note that this will break any existing code which relies on accessing via `attr_name`s instead of python field names.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -40,11 +40,13 @@ class Attribute(object):
 
     def __set__(self, instance, value):
         if instance:
-            instance.attribute_values[self.attr_name] = value
+            attr_name = instance._dynamo_to_python_attrs.get(self.attr_name, self.attr_name)
+            instance.attribute_values[attr_name] = value
 
     def __get__(self, instance, owner):
         if instance:
-            return instance.attribute_values.get(self.attr_name, None)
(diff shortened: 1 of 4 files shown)


pynamodb/attributes.py
```diff
@@ -40,11 +40,13 @@ def __init__(self,
...
         if instance:
-            instance.attribute_values[self.attr_name] = value
+            attr_name = instance._dynamo_to_python_attrs.get(self.attr_name, self.attr_name)
+            instance.attribute_values[attr_name] = value
 
     def __get__(self, instance, owner):
         if instance:
-            return instance.attribute_values.get(self.attr_name, None)
+            attr_name = instance._dynamo_to_python_attrs.get(self.attr_name, self.attr_name)
+            return instance.attribute_values.get(attr_name, None)
```
johnliu: oof — I'm fine with whatever approach you want but this is a mouth full, also super confusing which is which attr_name at this point

garrettheel: I agree and I'm not a huge fan of all of the conversions we have to do between these. This is mostly a function of earlier design decisions made.

That said, I'm OK with this for now because it's still easy to identify all of the places that these things happen if we want to do a larger refactor later

johnliu:
🐑

jmphilli:
+1

garrettheel:
Closing in favor of #292 which was updated instead

PR #311, opened 2017-07-07
Add consistent_read parameter to Model.scan

lita:


setup.py
```diff
@@ -13,7 +13,7 @@
 
 install_requires = [
     'six',
-    'botocore>=1.0.0',
+    'botocore>=1.2.0',
```
danielhochman: can you update requirements.txt as well?

pynamodb/models.py
```diff
@@ -723,6 +724,7 @@ def rate_limited_scan(cls,
             allow_rate_limited_scan_without_consumed_capacity=allow_rate_limited_scan_without_consumed_capacity,
             max_sleep_between_retry=max_sleep_between_retry,
             max_consecutive_exceptions=max_consecutive_exceptions,
+            consistent_read=consistent_read
```
danielhochman: can you use trailing commas?

setup.py
lita: yes!

pynamodb/models.py
lita: yup

pynamodb/models.py
```diff
@@ -765,7 +768,8 @@ def scan(cls,
             limit=page_size,
             scan_filter=key_filter,
             total_segments=total_segments,
-            conditional_operator=conditional_operator
+            conditional_operator=conditional_operator,
+            consistent_read=consistent_read
```
garrettheel: Can you document this new param in the methods too?

requirements.txt
```diff
@@ -1,2 +1,2 @@
-botocore==1.1.2
+botocore>=1.2.0
```
danielhochman: i would still pin to a specific version here

pynamodb/models.py
lita: done!

PR #312, opened 2017-07-07
Improve clarity around count() without a hash_key

Lordnibbler:
After being confused by [this logic](https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/models.py#L527-L528), @garrettheel and I thought it would be a good idea to clarify the `count()` documentation.

This elaborates slightly on the somewhat unexpected behavior that happens if you do not pass an explicit `hash_key` position argument to the `count` method -- it returns the count of all rows in the table.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -123,6 +123,17 @@ Alternatively, you can retrieve the table item count by calling the `count` meth
     print(UserModel.count())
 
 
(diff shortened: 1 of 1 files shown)


Lordnibbler:
JK let's do this instead https://github.com/pynamodb/PynamoDB/pull/313

PR #315, opened 2017-07-11
Support query only by page_size

matibek:
There are situations where we want to make a page_size "limited" query (i.e. without carrying about the number of returned items)

This PR adds that capability.


pynamodb/models.py
```diff
@@ -648,6 +648,9 @@ def query(cls,
                 limit -= 1
             yield cls.from_raw_data(item)
 
+        if limit is None and page_size is not None:
```
garrettheel: What about when the total number of results matching the query is `> page_size`?

From: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html
> Also, if the processed data set size exceeds 1 MB before DynamoDB reaches this limit, it stops the operation and returns the matching values up to the limit

matibek: Please see my changes. I want to make sure that we don't scan more than the `page_size`. Incase when the result size is > 1MB, we keep scanning until we reach a total amount of `page_size` scan.

jpinner-lyft:
@matibek Closing this PR since the pagination code has changed.
Please take a look at #395 to see if that addresses your concerns.

Issue #219, opened 2017-01-11
Support new DynamoDB API


garrettheel:
This is a pretty big change, so keeping track of things in here.

We should move off of the legacy API for conditions whilst keeping backwards compatibility with PynamoDB 2.x for now (with a view to deprecate things at some point in the future).

See the following DynamoDB doc for more information on the changes:

http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LegacyConditionalParameters.html

This should be a major version bump up to 3 as it will change a lot and likely introduce a new API for some things.

jpinner-lyft:
Complete Expression API Support: https://github.com/pynamodb/PynamoDB/pull/352

PR #319, opened 2017-08-01
Replace AttributesToGet with ProjectionExpression / ExpressionAttributeNames

jpinner-lyft:
This PR is a first step towards moving off of the legacy conditional parameters (#219).

As part of this change, the `attributes_to_get` keyword arguments now support Document Paths:
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.Attributes.html

Note that in order to not need special handling for reserved words,
all attributes are replaced with expression attribute name placeholders:
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ReservedWords.html

Breaking Changes:

Neither `[` nor `]` are allowed characters for attribute names; however, `.` is.
Any attribute names that include a `.` (defined using the `attr_name` keyword argument)
are no longer able to be requested since the name is interpreted as a nested path.
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html

Invalid document paths (e.g. empty paths or non-integer array indexes) now throw a `ValueError`
during request construction instead of a `ClientError` during response processing.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -6,6 +6,7 @@ from __future__ import division
 import logging
 import math
 import random
+import re
 import time
 import uuid
 from base64 import b64decode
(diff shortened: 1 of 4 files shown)


jmphilli:
:thumbsup:

PR #321, opened 2017-08-02
Replace KeyConditions with KeyConditionExpression / ExpressionAttributeValues

jpinner-lyft:
This PR is a second step towards moving off of the legacy conditional parameters (#219).

With this change, the `key_conditions` query parameter now uses the KeyConditionExpression API.
To support this, this change introduces primitive support for the subset of Condition Expressions required for partition and sort key comparisons along with support for Expression Attribute Values:
http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-KeyConditionExpression
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ExpressionAttributeValues.html

In order to maintain backwards compatibility, the Expression and Condition objects take attribute path and attribute value maps as parameters. To simplify development, helper functions have been added to the Attribute class:

```
>>> from pynamodb.attributes import UnicodeAttribute
>>> partition_key = UnicodeAttribute(attr_name='foo')
>>> sort_key = UnicodeAttribute(attr_name='bar')
>>> condition = (partition_key == 'Hello') & sort_key.startswith('World!')
>>> name_placeholders, expression_attribute_values = {}, {}
>>> condition.serialize(name_placeholders, expression_attribute_values)
'#0 = :0 AND begins_with (#1, :1)'
>>> name_placeholders
{'foo': '#0', 'bar': '#1'}
>>> expression_attribute_values
{':0': {'S': u'Hello'}, ':1': {'S': u'World!'}}
>>> 
```

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -6,7 +6,6 @@ from __future__ import division
 import logging
 import math
 import random
-import re
 import time
 import uuid
 from base64 import b64decode
@@ -25,7 +24,7 @@ from pynamodb.constants import (
     RETURN_CONSUMED_CAPACITY_VALUES, RETURN_ITEM_COLL_METRICS_VALUES, COMPARISON_OPERATOR_VALUES,
     RETURN_ITEM_COLL_METRICS, RETURN_CONSUMED_CAPACITY, RETURN_VALUES_VALUES, ATTR_UPDATE_ACTIONS,
     COMPARISON_OPERATOR, EXCLUSIVE_START_KEY, SCAN_INDEX_FORWARD, SCAN_FILTER_VALUES, ATTR_DEFINITIONS,
-    BATCH_WRITE_ITEM, CONSISTENT_READ, ATTR_VALUE_LIST, DESCRIBE_TABLE, KEY_CONDITIONS,
+    BATCH_WRITE_ITEM, CONSISTENT_READ, ATTR_VALUE_LIST, DESCRIBE_TABLE, KEY_CONDITION_EXPRESSION,
     BATCH_GET_ITEM, DELETE_REQUEST, SELECT_VALUES, RETURN_VALUES, REQUEST_ITEMS, ATTR_UPDATES,
     PROJECTION_EXPRESSION, SERVICE_NAME, DELETE_ITEM, PUT_REQUEST, UPDATE_ITEM, SCAN_FILTER, TABLE_NAME,
     INDEX_NAME, KEY_SCHEMA, ATTR_NAME, ATTR_TYPE, TABLE_KEY, EXPECTED, KEY_TYPE, GET_ITEM, UPDATE,
@@ -36,17 +35,19 @@ from pynamodb.constants import (
     CONSUMED_CAPACITY, CAPACITY_UNITS, QUERY_FILTER, QUERY_FILTER_VALUES, CONDITIONAL_OPERATOR,
     CONDITIONAL_OPERATORS, NULL, NOT_NULL, SHORT_ATTR_TYPES, DELETE,
     ITEMS, DEFAULT_ENCODING, BINARY_SHORT, BINARY_SET_SHORT, LAST_EVALUATED_KEY, RESPONSES, UNPROCESSED_KEYS,
-    UNPROCESSED_ITEMS, STREAM_SPECIFICATION, STREAM_VIEW_TYPE, STREAM_ENABLED, EXPRESSION_ATTRIBUTE_NAMES)
+    UNPROCESSED_ITEMS, STREAM_SPECIFICATION, STREAM_VIEW_TYPE, STREAM_ENABLED,
+    EXPRESSION_ATTRIBUTE_NAMES, EXPRESSION_ATTRIBUTE_VALUES, KEY_CONDITION_OPERATOR_MAP)
 from pynamodb.exceptions import (
     TableError, QueryError, PutError, DeleteError, UpdateError, GetError, ScanError, TableDoesNotExist,
     VerboseClientError
 )
+from pynamodb.expressions.condition import Path
+from pynamodb.expressions.projection import create_projection_expression
 from pynamodb.settings import get_settings_value
 from pynamodb.signals import pre_dynamodb_send, post_dynamodb_send
 from pynamodb.types import HASH, RANGE
 
 BOTOCORE_EXCEPTIONS = (BotoCoreError, ClientError)
-PATH_SEGMENT_REGEX = re.compile(r'([^\[\]]+)((?:\[\d+\])*)$')
 
 log = logging.getLogger(__name__)
 log.addHandler(NullHandler())
@@ -937,7 +938,7 @@ class Connection(object):
         if return_consumed_capacity:
             operation_kwargs.update(self.get_consumed_capacity_map(return_consumed_capacity))
         if attributes_to_get is not None:
-            projection_expression = self._get_projection_expression(attributes_to_get, name_placeholders)
+            projection_expression = create_projection_expression(attributes_to_get, name_placeholders)
             args_map[PROJECTION_EXPRESSION] = projection_expression
         if name_placeholders:
             args_map[EXPRESSION_ATTRIBUTE_NAMES] = self._reverse_dict(name_placeholders)
@@ -966,7 +967,7 @@ class Connection(object):
         operation_kwargs = {}
         name_placeholders = {}
         if attributes_to_get is not None:
-            projection_expression = self._get_projection_expression(attributes_to_get, name_placeholders)
+            projection_expression = create_projection_expression(attributes_to_get, name_placeholders)
             operation_kwargs[PROJECTION_EXPRESSION] = projection_expression
         if name_placeholders:
             operation_kwargs[EXPRESSION_ATTRIBUTE_NAMES] = self._reverse_dict(name_placeholders)
@@ -1141,7 +1142,7 @@ class Connection(object):
         operation_kwargs = {TABLE_NAME: table_name}
         name_placeholders = {}
         if attributes_to_get is not None:
-            projection_expression = self._get_projection_expression(attributes_to_get, name_placeholders)
+            projection_expression = create_projection_expression(attributes_to_get, name_placeholders)
             operation_kwargs[PROJECTION_EXPRESSION] = projection_expression
         if name_placeholders:
             operation_kwargs[EXPRESSION_ATTRIBUTE_NAMES] = self._reverse_dict(name_placeholders)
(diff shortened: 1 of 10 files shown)


pynamodb/connection/base.py
```diff
@@ -1198,11 +1199,53 @@ def query(self,
         """
         operation_kwargs = {TABLE_NAME: table_name}
         name_placeholders = {}
+        expression_attribute_values = {}
+
+        # Create KeyConditionExpression
+        tbl = self.get_meta_table(table_name)
```
lita: nit: can we just call this `table`?

lita: oh nvm, it was like this before.

jpinner-lyft: sure -- i just moved it but i'm happy to rename

pynamodb/connection/base.py
```diff
@@ -1198,11 +1199,53 @@ def query(self,
...
+            hash_keyname = tbl.get_index_hash_keyname(index_name)
+            if not hash_keyname:
+                raise ValueError("No hash key attribute for index: {0}".format(index_name))
+        else:
+            hash_keyname = tbl.hash_keyname
+        # TODO: refactor to use condition expression context when it is available
+        pk_name = _substitute_names(hash_keyname, name_placeholders)
+        pk_value = self._get_value_placeholder(table_name, hash_keyname, hash_key, expression_attribute_values)
+        key_condition_expression =  "{0} = {1}".format(pk_name, pk_value)
+        if key_conditions is None or len(key_conditions) == 0:
```
lita: Is it possible to move these into methods, as the code is getting complex?

The key_conditions validation/transformation logic should be in one method that returns sk_value_list and sk_name. 

Then can we split up the if statements into other methods? or another method that will eventually returns a `key_condition_expression`. it is hard to follow the code the way it is right now.

jmphilli:
have you tried this with the version of `boto` that pynamo requires? it's pretty low / out of date

pynamodb/connection/base.py
```diff
@@ -1198,11 +1199,53 @@ def query(self,
...
+            pass  # No comparisons on sort key
+        elif len(key_conditions) > 1:
+            raise ValueError("Multiple attributes are not supported in key_conditions: {0}".format(key_conditions))
```
garrettheel: What about sort key? Multiple expressions operating on the same value (partition or sort)?

pynamodb/expressions/__init__.py
```diff
@@ -0,0 +1,35 @@
+import re
```
jmphilli: i generally dislike code in the `__init__` files. garrett (aka 🐍  man) might think it's pythonic or something tho

jpinner-lyft: i can put it in a util.py

jmphilli: ya that looks nice, but there's a fun behavior where people do something like
```python
import pynamodb.expressions
# stuff
o = pynamodb.expressions.condition.Operand(attr)
get_projection_expression() # this resolves 
```
it's not a big deal, and again, python people probably have better opinions than i do on this, but i always have trouble finding function definitions when they're defined in the `__init__` file.

as for the api perspective, we could just put these functions into a class that has the `placeholders` defined on it possibly. i'm generally uncomfortable mutating parameters passed into funcs (`placeholders`) 

the problem with that then of course being that the model + connection/base need reference to the same placeholder instance

pynamodb/connection/base.py
```diff
@@ -1198,11 +1199,39 @@ def query(self,
...
+        else:
+            hash_keyname = tbl.hash_keyname
+
+        key_condition_expression = self._get_condition_expression(table_name, hash_keyname, '__eq__', hash_key)
+        if key_conditions is None or len(key_conditions) == 0:
+            pass  # No comparisons on sort key
+        elif len(key_conditions) > 1:
+            raise ValueError("Multiple attributes are not supported in key_conditions: {0}".format(key_conditions))
+        else:
+            (key, condition), = key_conditions.items()
```
jmphilli:
```python
(key, condition), _ = key_conditions.items()  # this explicitly documents that you are aware of the second part of the tuple and in fact do not want it. otherwise people may think it's a stray comma
```

pynamodb/connection/base.py
```diff
@@ -1223,49 +1252,20 @@ def query(self,
...
         try:
             return self.dispatch(QUERY, operation_kwargs)
         except BOTOCORE_EXCEPTIONS as e:
             raise QueryError("Failed to query items: {0}".format(e), e)
 
-    @staticmethod
-    def _get_projection_expression(attributes_to_get, placeholders):
-        expressions = [_substitute_names(attribute, placeholders) for attribute in attributes_to_get]
-        return ', '.join(expressions)
+    def _get_condition_expression(self, table_name, attribute_name, operator, *values):
```
jmphilli: i feel like this should be in that `expressions.utils` class no?

jpinner-lyft: done

pynamodb/expressions/__init__.py
jpinner-lyft: done

pynamodb/connection/base.py
jpinner-lyft:
```
>>> d = {'key': 'value'}
>>> (key, value), _ = d.items()
  File "<stdin>", line 1, in <module>
ValueError: need more than 1 value to unpack
```

pynamodb/expressions/condition.py
```diff
@@ -0,0 +1,88 @@
...
+        # This seemed preferable to other options such as merging value1 <= attribute & attribute <= value2
+        # into one condition expression. DynamoDB only allows a single sort key comparison and having this
+        # work but similar expressions like value1 <= attribute & attribute < value2 fail seems too brittle.
+        return Between(self.attribute, value1, value2)
+
+    def begins_with(self, prefix):
+        return BeginsWith(self.attribute, prefix)
+
+    def startswith(self, prefix):
+        # A 'pythonic' replacement for begins_with to match string behavior (e.g. "foo".startswith("f"))
```
lita: This is more pythonic, but I prefer not having two ways of doing things. Would it be possible to just have `startswith` call `BeginsWith(self.attribute, prefix)` and forgo begins_with?

jpinner-lyft: done

pynamodb/expressions/util.py
```diff
@@ -0,0 +1,29 @@
...
+
+PATH_SEGMENT_REGEX = re.compile(r'([^\[\]]+)((?:\[\d+\])*)$')
+
+
+def substitute_names(expression, placeholders):
+    """
+    Replaces names in the given expression with placeholders.
+    Stores the placeholders in the given dictionary.
+    """
+    path_segments = expression.split('.')
```
lita: Would this break if we pass in an attribute name that has a '.' in it? Or was this already not working in Pynamodb before

jpinner-lyft: Currently it wouldn't do the right thing -- I hope to address this in my next changeset.

lita: sounds good. Just want to make sure.

pynamodb/connection/base.py
```diff
@@ -1198,11 +1199,39 @@ def query(self,
         """
         operation_kwargs = {TABLE_NAME: table_name}
         name_placeholders = {}
+        expression_attribute_values = {}
+
+        tbl = self.get_meta_table(table_name)
+        if tbl is None:
+            raise TableError("No such table: {0}".format(table_name))
```
garrettheel: I don't think this is valid failure case, I'd expect `get_meta_data` to raise an exception if it doesn't exist.

pynamodb/connection/base.py
```diff
@@ -1223,49 +1252,20 @@ def query(self,
...
-                operation_kwargs[KEY_CONDITIONS][key] = {
-                    ATTR_VALUE_LIST: [
-                        {
-                            attr_type: self.parse_attribute(value)
-                        } for value in condition.get(ATTR_VALUE_LIST)
-                    ],
-                    COMPARISON_OPERATOR: operator
-                }
+        if name_placeholders:
+            operation_kwargs[EXPRESSION_ATTRIBUTE_NAMES] = self._reverse_dict(name_placeholders)
```
garrettheel: would really prefer for this sort of logic to live somewhere in the `expression` module

pynamodb/expressions/condition.py
```diff
@@ -0,0 +1,85 @@
+from pynamodb.expressions.util import get_value_placeholder, substitute_names
+
+
+class Path(object):
```
garrettheel: Can we add a `__repr__` for introspection?

pynamodb/connection/base.py
jpinner-lyft: previously written code -- can follow up in a different PR

pynamodb/connection/base.py
jpinner-lyft: so the problem here is that the `names_placeholders` array has to be shared across multiple expressions (key condition + filter for example) so no single serialization can do the reverse

i can move the function itself into the utils but at this point the use is localized to the base connection class

pynamodb/expressions/condition.py
jpinner-lyft: Next change set is going to attempt to address support for 'dotted' attributes and better support for document paths. Will include in that PR since I'm not sure yet how this class with change.

pynamodb/expressions/condition.py
```diff
@@ -0,0 +1,85 @@
...
+        # work but similar expressions like value1 <= attribute & attribute < value2 fail seems too brittle.
+        return Between(self.path, value1, value2)
+
+    def startswith(self, prefix):
+        # A 'pythonic' replacement for begins_with to match string behavior (e.g. "foo".startswith("f"))
+        return BeginsWith(self.path, prefix)
+
+
+class Condition(object):
+    format_string = '{path} {operator} {0}'
```
jlafon: Is the {0} in format_string meant to only take the first element of values (line 51)?
```python
>>> format_string.format(*(1,2), path='foo', operator='<')
'foo < 1'`

File: pynamodb/expressions/condition.py [hunk 13]
```diff
@@ -0,0 +1,85 @@
...
+        if self.logical_operator:
+            other_condition = self.other_condition.serialize(placeholder_names, expression_attribute_values)
+            return '{0} {1} {2}'.format(condition, self.logical_operator, other_condition)
+        return condition
+
+    def __and__(self, other):
+        if not isinstance(other, Condition):
+            raise TypeError("unsupported operand type(s) for &: '{0}' and '{1}'",
+                            self.__class__.__name__, other.__class__.__name__)
+        self.logical_operator = 'AND'
```
jlafon: A trivial thing, but AND is already a constant from constants.py.

pynamodb/expressions/condition.py
```diff
@@ -0,0 +1,85 @@
...
+    def __bool__(self):
+        # Prevent users from accidentally comparing the condition object instead of the attribute instance
+        raise TypeError("unsupported operand type(s) for bool: {0}".format(self.__class__.__name__))
+
+
+class Between(Condition):
+    format_string = '{path} {operator} {0} AND {1}'
+
+    def __init__(self, attribute, value1, value2):
+        super(Between, self).__init__(attribute, 'BETWEEN', value1, value2)
```
jlafon: Also a constant.

jlafon:
Very glad to see somebody updating this away from the legacy DynamoDB API. Nice work - especially all of those tests.

pynamodb/expressions/condition.py
jpinner-lyft: Yes, line 51 having a variable length arg array is mean to accommodate the `between` expression which takes two args. The `Between` class overwrites the format string as seen on Line 75. All the other expressions (thus far) only compare against a single value.

pynamodb/expressions/condition.py
jpinner-lyft: good point -- will incorporate the change into the next PR (https://github.com/pynamodb/PynamoDB/pull/322)

pynamodb/expressions/condition.py
jpinner-lyft: same

PR #322, opened 2017-08-05
Support document paths and attribute names containing dot characters.

jpinner-lyft:
This PR introduces support for both Document Paths and '.' characters in expressions.

Expressions are generated assuming that the path given in the object is a document path;
for example, `Path('foo.bar.baz[0]') == value` is translated into `#0.#1.#2[0] = :0` with the following expression attribute names: `{'#0': 'foo', '#1': 'bar', "#2': 'baz'}`.

With this change, expressions generated from Attribute objects will use whatever attribute name is defined by the `attr_name` field, allowing support for dotted attribute names; for example,
`ListAttribute(attr_name='foo.bar.baz')[0] == value` is translated into `#0[0] = :0` with the following expression attribute names: `{'#0: 'foo.bar.baz'}`.

### DIFF
--- a/pynamodb/expressions/condition.py
+++ b/pynamodb/expressions/condition.py
@@ -1,10 +1,21 @@
+from copy import copy
+from pynamodb.constants import AND, BETWEEN
 from pynamodb.expressions.util import get_value_placeholder, substitute_names
 
 
 class Path(object):
 
-    def __init__(self, path):
+    def __init__(self, path, attribute_name=False):
         self.path = path
+        self.attribute_name = attribute_name
+
+    def __getitem__(self, idx):
+        # list dereference operator
+        if not isinstance(idx, int):
+            raise TypeError("list indices must be integers, not {0}".format(type(idx).__name__))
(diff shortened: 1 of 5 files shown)


pynamodb/expressions/condition.py
```diff
@@ -74,12 +108,12 @@ def __bool__(self):
...
+        super(Between, self).__init__(path, BETWEEN, lower, upper)
 
 
 class BeginsWith(Condition):
     format_string = '{operator} ({path}, {0})'
 
-    def __init__(self, attribute, prefix):
-        super(BeginsWith, self).__init__(attribute, 'begins_with', prefix)
+    def __init__(self, path, prefix):
+        super(BeginsWith, self).__init__(path, 'begins_with', prefix)
```
jmphilli: const?

jpinner-lyft: the current `BEGINS_WITH` const is for the old api which is capitalized and not lowercase :(

PR #323, opened 2017-08-07
Fix __get__ for MapAttributes to return the type of the MapAttribute …

rowillia:
…as opposed to the instance when accessed from a AttributeContainer class.

For example, the code below should just work:

```python
from pynamodb.attributes import MapAttribute, NumberAttribute

class MyMap(MapAttribute):
    foo = NumberAttribute()

class MyModel(Model):
    class Meta:
        table_name = 'tbl'
    map_attr = MyMap()

MyMap.map_attr.foo.attr_name
```

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -465,6 +465,19 @@ class MapAttribute(AttributeContainer, Attribute):
         self._set_defaults()
         self._set_attributes(**attrs)
 
+    def __get__(self, instance, owner):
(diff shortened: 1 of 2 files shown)


jpinner-lyft:
:lock: this doesn't quite work -- let me play around with this a bit more

jpinner-lyft:
fix for this issue: https://github.com/pynamodb/PynamoDB/pull/325

PR #326, opened 2017-08-14
Introduce AttributeContainerMeta

jpinner-lyft:
This pull request is a follow-up to https://github.com/pynamodb/PynamoDB/pull/324.

This change moves the `_initialize_attributes` call to the init method of a new metaclass for `AttributeContainer`. The result of this is both `_attributes` and `_dynamo_to_python_attrs` are valid as soon as the class is created as opposed to after the class is first instantiated. Contrast the following code snippet to the examples in the description of the above mentioned pull request.
```
>>> from pynamodb.attributes import Attribute, MapAttribute
>>> class SubMapAttribute(MapAttribute):
...     a = Attribute()
... 
>>> class SubSubMapAttribute(SubMapAttribute):
...     b = Attribute()
... 
>>> MapAttribute._attributes
{}
>>> SubMapAttribute._attributes
{'a': <pynamodb.attributes.Attribute object at 0x109ae9e90>}
>>> SubSubMapAttribute._attributes
{'a': <pynamodb.attributes.Attribute object at 0x109ae9e90>, 'b': <pynamodb.attributes.Attribute object at 0x109afaa50>}
```

In addition, this change re-introduces `MapAttributeMeta` to undo a breaking change from the prior pull request. `type(MapAttribute)` now continues to return `MapAttributeMeta`

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -2,6 +2,7 @@
 PynamoDB attributes
 """
 import six
+from six import add_metaclass
 import json
 from base64 import b64encode, b64decode
 from datetime import datetime
@@ -118,11 +119,13 @@ class AttributePath(Path):
         return {ATTR_TYPE_MAP[self.attribute.attr_type]: self.attribute.serialize(value)}
 
 
-class AttributeContainer(object):
+class AttributeContainerMeta(type):
 
-    _dynamo_to_python_attrs = None
+    def __init__(cls, name, bases, attrs):
+        super(AttributeContainerMeta, cls).__init__(name, bases, attrs)
+        AttributeContainerMeta._initialize_attributes(cls)
 
-    @classmethod
+    @staticmethod
     def _initialize_attributes(cls):
         """
         Initialize attributes on the class.
@@ -157,6 +160,10 @@ class AttributeContainer(object):
                 else:
                     instance.attr_name = item_name
 
+
+@add_metaclass(AttributeContainerMeta)
+class AttributeContainer(object):
+
     @classmethod
     def _get_attributes(cls):
         """
@@ -164,9 +171,6 @@ class AttributeContainer(object):
 
         :rtype: dict[str, Attribute]
         """
-        if '_attributes' not in cls.__dict__:
-            # Each subclass of AttributeContainer needs its own attributes map.
-            cls._initialize_attributes()
         return cls._attributes
 
     @classmethod
@@ -176,8 +180,6 @@ class AttributeContainer(object):
 
         This covers cases where an attribute name has been overridden via "attr_name".
         """
-        if cls._attributes is None:
-            cls._initialize_attributes()
         return cls._dynamo_to_python_attrs.get(dynamo_key, dynamo_key)
 
     def _set_defaults(self):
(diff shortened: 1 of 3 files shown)


pynamodb/attributes.py
```diff
@@ -118,11 +119,13 @@ def _serialize(self, value):
...
-class AttributeContainer(object):
+class AttributeContainerMeta(type):
 
-    _dynamo_to_python_attrs = None
+    def __init__(cls, name, bases, attrs):
+        super(AttributeContainerMeta, cls).__init__(name, bases, attrs)
+        AttributeContainerMeta._initialize_attributes(cls)
 
-    @classmethod
+    @staticmethod
```
jmphilli: if you switch to static, you have to remove the `cls` parameter in the func.

jpinner-lyft: This function is now in the metaclass, so the `cls` here is the class being created, just like in `__init__`.

pynamodb/attributes.py
```diff
@@ -473,6 +475,13 @@ def deserialize(self, value):
         return None
 
 
+class MapAttributeMeta(AttributeContainerMeta):
+    """
+    This is only here for backwards compatibility: i.e. so type(MapAttribute) == MapAttributeMeta
+    """
```
lita: don't you need to add a `pass` in here? I am surprised this is syntactically correct in python 2.7.

lita:
Can you add a test for these case? More specifically the `MapAttributeMeta` type checking? 

I am not much worried about the _attribritues changing (as those are "protected" attributes), but it would be good to make sure we didn't break subclasses models.

I am mildly concerned that now that we are adding a metaclasses to both `Attribute` and `MapAttribute`, our users can't add their own meta classes on top of that (without some horrible code), making this a breaking change. I am just questioning if this more magic then needed.

pynamodb/attributes.py
jpinner-lyft: a string is valid statement in python

jpinner-lyft:
@lita Will add a test that checks the type of `MapAttribute`.

This change doesn't add a metaclass to `Attribute` -- it does add a metaclass to `AttributeContainer` however the two subclasses `MapAttribute` and `Model` already had metaclasses so any users who were subclassing and adding metaclasses already had to make `MetaModel` or `MapAttributeMeta` base classes of their metaclass.

As to your question of is this more magic than we need. `MapAttributeMeta` doesn't need to exist but it was added in v2.0.0. `AttributeContainer` requires class level initialization so should have had a metaclass to introspect the class attributes instead of making them lazily instantiated.

lita:
My bad. I meant adding a metaclass onto `MapAttribute`. None of the attributes have metaclass thus far. I do agree we already have metaclass on `Model`, but that has been that way since the very beginning to create the connection with the Dynamo table. Now we are introducing more complexity to the attribute class and I want to make sure that it is worth doing. 

I am not quite sure what introspection at the class level buys us vs lazily instantiating them. If it is just to fix the typing, then we can fix that without introducing metaclasses. The main con I see is that we are introduing a breaking change to MapAttribute for people who want to add a meta class on top of it. I am not completely against this, but I am not seeing what this buys us.

jpinner-lyft:
@lita MapAttribute had a metaclass since v2.0.0 -- my breaking change was removing it in #324

pynamodb/attributes.py
lita: Oh i mean, I am surprised you don't have to do:
```python
class MapAttributeMeta(AttributeContainerMeta):
"""
    This is only here for backwards compatibility: i.e. so type(MapAttribute) == MapAttributeMeta
"""
    pass

lita:
Talked in Slack. Looks like knowing the attributes at the class level will be good for buidling the query pattern. This change is mostly for backwards compatitblity.

PR #329, opened 2017-08-14
Support complete condition expression syntax.

jpinner-lyft:
This PR is a third step towards moving off of the legacy conditional parameters (#219).

This PR implements the complete operator and function syntax defined for condition expressions:
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html

In order to support "size" which is both an operand and a function, we make `Path` a subclass of `Operand` and create a special subclass called `Size` just to implement these comparisons.

This change also adds support for primitive values in comparisons (e.g. `size('foo') == 3`).

Note: not all restrictions on the function operands have been enforced and there is no way yet to determine if a given comparison is valid for use in a key condition expression.

Open Question: what is the use case / appropriate syntax for Attribute.is_type()?

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -76,6 +76,11 @@ class Attribute(object):
             return self is other
         return AttributePath(self).__eq__(other)
 
+    def __ne__(self, other):
+        if other is None or isinstance(other, Attribute):  # handle object identity comparison
+            return self is not other
+        return AttributePath(self).__ne__(other)
+
     def __lt__(self, other):
         return AttributePath(self).__lt__(other)
 
@@ -94,8 +99,24 @@ class Attribute(object):
     def between(self, lower, upper):
         return AttributePath(self).between(lower, upper)
 
+    def is_in(self, *values):
+        return AttributePath(self).is_in(*values)
+
+    def exists(self):
+        return AttributePath(self).exists()
+
+    def not_exists(self):
+        return AttributePath(self).not_exists()
+
+    def is_type(self):
+        # What makes sense here? Are we using this to check if deserialization will be successful?
+        return AttributePath(self).is_type(ATTR_TYPE_MAP[self.attr_type])
+
     def startswith(self, prefix):
         return AttributePath(self).startswith(prefix)
+
+    def contains(self, item):
+        return AttributePath(self).contains(item)
 
 
 class AttributePath(Path):
--- a/pynamodb/expressions/condition.py
+++ b/pynamodb/expressions/condition.py
@@ -56,6 +56,7 @@ class Size(Operand):
     """
 
     def __init__(self, path):
+        # prevent circular import -- AttributePath imports Path
         from pynamodb.attributes import Attribute, AttributePath
         if isinstance(path, Path):
             self.path = Path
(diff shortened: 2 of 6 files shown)


pynamodb/expressions/condition.py
```diff
@@ -112,8 +191,70 @@ def __init__(self, path, lower, upper):
...
+
+    def __init__(self, condition1, condition2):
+        super(Or, self).__init__(None, OR, condition1, condition2)
+
+
+class Not(Condition):
+    format_string = '({operator} {0})'
+
+    def __init__(self, condition):
+        super(Not, self).__init__(None, 'NOT', condition)
```
jmphilli: there's a reason this `NOT` isn't a const right?

pynamodb/expressions/condition.py
```diff
@@ -112,8 +191,70 @@ def __init__(self, path, lower, upper):
...
+
+
+class IsType(Condition):
+    format_string = '{operator} ({path}, {0})'
+
+    def __init__(self, path, attr_type):
+        if attr_type not in SHORT_ATTR_TYPES:
+            raise ValueError("{0} is not a valid attribute type. Must be one of {1}".format(
+                attr_type, SHORT_ATTR_TYPES))
+        super(IsType, self).__init__(path, 'attribute_type', {'S': attr_type})
```
jmphilli: not a const on purpose?

jpinner-lyft: used STRING_SHORT

pynamodb/expressions/condition.py
jpinner-lyft: there just wasn't one in constants.py

pynamodb/expressions/condition.py
```diff
@@ -41,13 +36,84 @@ def between(self, lower, upper):
...
+        return {ATTR_TYPE_MAP[attr_class.attr_type]: attr_class.serialize(value)}
+
+
+class Size(Operand):
+    """
+    Size is a special operand that represents the result of calling the 'size' function on a Path operand.
+    """
+
+    def __init__(self, path):
+        from pynamodb.attributes import Attribute, AttributePath
```
jmphilli: can we kick this up to the top of the file?

jpinner-lyft: we can't -- circular input

jpinner-lyft: commented

pynamodb/attributes.py
```diff
@@ -76,6 +76,11 @@ def __eq__(self, other):
             return self is other
         return AttributePath(self).__eq__(other)
 
+    def __ne__(self, other):
+        if other is None or isinstance(other, Attribute):  # handle object identity comparison
```
jmphilli: ? sure this is right?

pynamodb/expressions/condition.py
```diff
@@ -41,13 +36,85 @@ def between(self, lower, upper):
...
+    def is_in(self, *values):
+        values = [self._serialize(value) for value in values]
+        return In(self, *values)
+
+    def _serialize(self, value):
+        # Check to see if value is already serialized
+        if isinstance(value, dict) and len(value) == 1 and list(value.keys())[0] in SHORT_ATTR_TYPES:
+            return value
+        # Serialize value based on its type
+        from pynamodb.attributes import _get_class_for_serialize
```
jmphilli: cc @garrettheel , we need to consider all of these circular imports.

jmphilli:
+1

Issue #211, opened 2016-12-31
Use region from aws configuration by default


miguelgrinberg:
It would be nice if pynamodb used the region I have in my ~/.aws/credentials file by default. That's the behavior I get from boto and the aws cli. Instead, pynamodb tables that don't have a region explicitly stated always go to us-east-1.

yedpodtrzitko:
I can imagine the advantage on the dev machine, but how would that work in a production? Would you need to deploy `~/.aws/credentials` as well? Sounds cumbersome.

miguelgrinberg:
@yedpodtrzitko I'm not sure I understand why that would be cumbersome.

Boto has a procedure by which it loads credentials ([link](http://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials)). It doesn't always have to be a ~/.aws/credentials file, there are many other ways to feed credentials to boto, and all support passing a region. By taking a hardcoded choice for us-east-1, the standard procedure is being altered.

In your example of a production machine, say an EC2 instance, I would want credentials to be deployed to the EC2 metadata, so that there is no need to mess with actual credential files. I haven't tested this, but what would happen if I deploy an EC2 instance with a role from another region attached, then try to use pynamodb from that EC2? The id and secret will be picked up from the instance metadata, but then pynamodb will try to access us-east-1, which may or may not work depending on what role is attached to the instance. What I think pynamodb should do, is let boto3 figure out a default region using its own mechanisms.

brandond:
I just do something like this:

```python
from pynamodb.models import Model
from botocore.session import Session

class MyModel(Model):
    class Meta:
        table_name = 'my-model'
        region = Session().get_config_variable('region')
    myKey = UnicodeAttribute(hash_key=True)
```

That will use the region from the default profile, Lambda environment, etc.

jordal:
I agree this would be a nice feature. Would anyone volunteer to turn @brandond's snippet into a PR that always sets the default region appropriately?

jpinner-lyft:
@danielhochman @garrettheel could the same effect be achieved by simply removing pynamo's default of the region?

I would naively expect passing `None` to botocore's `create_client` call to cause it to go through its normal configuration steps, including using a region from the environment or from a credentials file if one is configured.

I don't believe botocore's lookup includes retrieving instance metadata as done in #330, but if it does not, I don't really think that is something we should include (a user can still choose to configure the region in the model's Meta class using that metadata).

[...]

PR #330, opened 2017-08-15
use region information from current instance solves issue #211

sebastianseitz:
A solution for the region lookup without denying the possiblity to set it in the global settings file.

1) If override is set: use this
2) Try to get information from instance meta
3) Try to get region from `AWS_DEFAULT_REGION` environment variable https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html


jpinner-lyft:
:lock: pending discussion in #211 :lock:

jpinner-lyft:
Likely outcome is to remove region default which will instead fallback to botocore's normal configuration mechanisms for region. This mechanism does not include looking up the instance metadata.

However, as part of removing the default we would include documentation or examples on how to set a region override in the `Meta` class which could include the instance metadata lookup as a code snippet.

PR #336, opened 2017-08-22
Conditional Operations using Condition Expressions

jpinner-lyft:
This PR is the fifth step towards moving off of the legacy conditional parameters (#219).

With this change, `delete_item`, `put_item`, and `update_item` (along with their corresponding model functions, `delete`, `save`, and `update`) now take an additional keyword arg, `condition`.

Any condition expression can be used for these conditional operations. Using legacy conditional parameters will emit a `warning`. Using both condition expressions and legacy parameters will raise a `ValueError`.


docs/conditional.rst
```diff
@@ -26,113 +24,95 @@ Suppose that you have defined a `Thread` Model for the examples below.
...
 
-Specifying that the conditions should be applied with AND or OR is achieved through the use of the `conditional_operator` keyword,
-which can be `and` or `or`.
+PynamoDB supports creating condition expressions from attributes using a simple syntax. Attributes can be compared against values.
+Any value provided will be serialized using the serializer defined for that attribute.
 
 .. code-block:: python
 
-    thread_item = Thread('Existing Forum', 'Example Subject')
+    print(Thread.forum_name == 'Some Forum')
```
lita: Since we are supporting 2.7 with these changes, we should make the documentation python 2.7/3 agnostic. So it would be better to use `str`.

Also, I would find it useful to document the result of commands rather than me pasting them into the interpreter.

docs/conditional.rst
```diff
@@ -142,7 +122,7 @@ You can check for conditional operation failures by inspecting the cause of the
 .. code-block:: python
 
     try:
-        thread_item.save(forum_name__null=False)
+        thread_item.save(Thread.forum_name.exists())
     except PutError as e:
```
lita: This doesn't have to be in this PR. But this kind of sucks. Can't we throw a generic ConditionalError that is subclassed from a PutError that people can catch? I imagine people would want to do conditional writes quite often to save on reads.

docs/conditional.rst
```diff
@@ -26,113 +24,95 @@ Suppose that you have defined a `Thread` Model for the examples below.
...
 
 
-AND vs. OR
-^^^^^^^^^^
+Condition Expressions
+^^^^^^^^^^^^^^^^^^^^^
 
-Specifying that the conditions should be applied with AND or OR is achieved through the use of the `conditional_operator` keyword,
-which can be `and` or `or`.
+PynamoDB supports creating condition expressions from attributes using a simple syntax. Attributes can be compared against values.
```
lita: We should state that `we support conditional expressions by from attributes using Python's built-in comparison operators. For DynamoDB expressions that don't have a built-in operator, they are a function call.`

I like the code examples and we should keep them. But we should have a complete list where we describe the arguments it takes, as it is not quite as clear. If we created a table with the following columns, I think it would clear up a lot of things:
* the operator 
* a short description (optional as it might be self-explanatory).
* an example of use (how many arguments it takes, etc)
* the result

Here are some examples:
https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not
https://docs.python.org/2/library/stdtypes.html#comparisons
https://pandas.pydata.org/pandas-docs/stable/api.html#binary-operator-functions

pynamodb/connection/base.py
```diff
@@ -879,6 +893,8 @@ def put_item(self,
         """
         Performs the PutItem operation and returns the result
         """
+        self._check_condition('condition', condition, expected, conditional_operator)
```
lita: Why do you need to pass in the string `'condition'`? Is it just for the error message?

docs/conditional.rst
jpinner-lyft: I'm not sure I follow, this works under python 2.7:
```
(venv) jpinner-mbp121:PynamoDB jeffpinner (conditional-operations)$ python
Python 2.7.11 (default, Jun  3 2016, 13:12:55) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pynamodb.attributes import UnicodeAttribute
>>> forum_name = UnicodeAttribute(attr_name='forum')
>>> print(forum_name == 'Some Forum')
forum = Some Forum
```

As for pasting the result from the interpreter -- I was following the style of the existing examples in this file, and as far as I can tell, this gets render as a copyable code-block in the documentation, not as an interpreter snippet, so adding the results would prevent a user from copy-pasting.

docs/conditional.rst
jpinner-lyft: Will try to put something together. I agree this isn't great, see:
https://github.com/pynamodb/PynamoDB/pull/108
https://github.com/pynamodb/PynamoDB/pull/112

pynamodb/connection/base.py
jpinner-lyft: Yes, this is to render the name of the parameter in the error message. When we use this for query and scan filters, the parameter name will be different.

docs/conditional.rst
jpinner-lyft: The problem here, by the way, is that both `PutError`, `DeleteError`, and `UpdateError` can have the `ConditionalCheckFailedException` error code. So I'd have to make three subclasses and users would have to choose the correct one.

jpinner-lyft:
I've moved the documentation changes into a standalone pull request.
https://github.com/pynamodb/PynamoDB/pull/338

I'm not sure how documentation updates take affect and don't want them to land until we publish a new version.

pynamodb/connection/base.py
lita: Hmm, can we make this a const?

PR #339, opened 2017-08-23
Query and Scan Filters using Condition Expressions

jpinner-lyft:
This PR is the sixth step towards moving off of the legacy conditional parameters (#219).

With this change `query` and `scan` now take additional keyword args for `range_key_condition` (query only) and `filter_condition`.

Using legacy filter parameters will emit a warning. Using both condition expressions and legacy parameters will raise a ValueError.

Note: The legacy filter parameters allow some invalid commands to be created. The new parameters correctly detect invalid commands and raise a ValueError.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -1319,8 +1319,10 @@ class Connection(object):
             # FilterExpression does not allow key attributes. Check for hash and range key name placeholders
             hash_key_placeholder = name_placeholders.get(hash_keyname)
             range_key_placeholder = range_keyname and name_placeholders.get(range_keyname)
-            if hash_key_placeholder in filter_expression or \
-                (range_key_placeholder and range_key_placeholder in filter_expression):
+            if (
+                hash_key_placeholder in filter_expression or
+                (range_key_placeholder and range_key_placeholder in filter_expression)
+            ):
                 raise ValueError("'filter_condition' cannot contain key attributes")
             operation_kwargs[FILTER_EXPRESSION] = filter_expression
         if attributes_to_get:
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -109,6 +109,24 @@ class MetaTable(object):
                     if schema_key.get(KEY_TYPE) == HASH:
                         return schema_key.get(ATTR_NAME)
 
+    def get_index_range_keyname(self, index_name):
+        """
+        Returns the name of the hash key for a given index
+        """
+        global_indexes = self.data.get(GLOBAL_SECONDARY_INDEXES)
+        local_indexes = self.data.get(LOCAL_SECONDARY_INDEXES)
+        indexes = []
+        if local_indexes:
+            indexes += local_indexes
+        if global_indexes:
+            indexes += global_indexes
+        for index in indexes:
+            if index.get(INDEX_NAME) == index_name:
+                for schema_key in index.get(KEY_SCHEMA):
+                    if schema_key.get(KEY_TYPE) == RANGE:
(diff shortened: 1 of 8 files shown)


pynamodb/connection/base.py
```diff
@@ -109,6 +109,24 @@ def get_index_hash_keyname(self, index_name):
                     if schema_key.get(KEY_TYPE) == HASH:
                         return schema_key.get(ATTR_NAME)
 
+    def get_index_range_keyname(self, index_name):
+        """
+        Returns the name of the hash key for a given index
+        """
+        global_indexes = self.data.get(GLOBAL_SECONDARY_INDEXES)
+        local_indexes = self.data.get(LOCAL_SECONDARY_INDEXES)
+        indexes = []
```
jmphilli: do we do this index aggregation anywhere else? if so, could be good to abstract to a func / lazy property

jpinner-lyft: we do it in one other functino,`get_index_range_keyname` -- i can make a private `_get_indexes` if you like

pynamodb/connection/base.py
```diff
@@ -1267,11 +1310,19 @@ def query(self,
...
 
-        operation_kwargs[KEY_CONDITION_EXPRESSION] = key_condition_expression.serialize(
+        operation_kwargs[KEY_CONDITION_EXPRESSION] = key_condition.serialize(
             name_placeholders, expression_attribute_values)
-
+        if filter_condition is not None:
+            filter_expression = filter_condition.serialize(name_placeholders, expression_attribute_values)
+            # FilterExpression does not allow key attributes. Check for hash and range key name placeholders
+            hash_key_placeholder = name_placeholders.get(hash_keyname)
+            range_key_placeholder = range_keyname and name_placeholders.get(range_keyname)
```
jmphilli: isn't the result of the right hand side a bool?

pynamodb/connection/base.py
jpinner-lyft: another option is to do a `_get_index(name)` function

pynamodb/connection/base.py
```diff
@@ -1267,11 +1310,19 @@ def query(self,
...
+            if hash_key_placeholder in filter_expression or \
```
jmphilli: can you paren the whole thing?
```python
if (
    hash_key_placeholder in filter_expression or 
    (range_key_placeholder and range_key_placeholder in filter_expression
):
```

pynamodb/expressions/condition.py
```diff
@@ -137,6 +137,10 @@ def __init__(self, path, operator, *values):
         self.operator = operator
         self.values = values
 
+    # http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-KeyConditionExpression
+    def is_valid_range_key_condition(self, path):
+        return str(self.path) == path and self.operator in ['=', '<', '<=', '>', '>=', BETWEEN, 'begins_with']
```
jmphilli: we don't have consts for the operator strings?

pynamodb/connection/base.py
jpinner-lyft:
```
>>> d = {'key': 'value'}
>>> print(None and d.get(None))
None
>>> print('key' and d.get('key'))
value
```

pynamodb/connection/base.py
jpinner-lyft: done

pynamodb/expressions/condition.py
jpinner-lyft: only for `BETWEEN` since it's in the old api

PR #341, opened 2017-08-24
Replace AttributeUpdates with UpdateExpression

jpinner-lyft:
This PR introduces a translation from the legacy attribute update syntax to update expressions.
It is just a "complete enough" implementation in order to make update commands functional.

Further work is required to provide proper low-level and high-level APIs that tie into the "Path" and "Attribute" objects. This is purely an initial step given that the exact design of those APIs is not yet complete.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -34,9 +34,9 @@ from pynamodb.constants import (
     WRITE_CAPACITY_UNITS, GLOBAL_SECONDARY_INDEXES, PROJECTION, EXCLUSIVE_START_TABLE_NAME, TOTAL,
     DELETE_TABLE, UPDATE_TABLE, LIST_TABLES, GLOBAL_SECONDARY_INDEX_UPDATES,
     CONSUMED_CAPACITY, CAPACITY_UNITS, QUERY_FILTER, QUERY_FILTER_VALUES, CONDITIONAL_OPERATOR,
-    CONDITIONAL_OPERATORS, NULL, NOT_NULL, SHORT_ATTR_TYPES, DELETE,
+    CONDITIONAL_OPERATORS, NULL, NOT_NULL, SHORT_ATTR_TYPES, DELETE, PUT,
     ITEMS, DEFAULT_ENCODING, BINARY_SHORT, BINARY_SET_SHORT, LAST_EVALUATED_KEY, RESPONSES, UNPROCESSED_KEYS,
-    UNPROCESSED_ITEMS, STREAM_SPECIFICATION, STREAM_VIEW_TYPE, STREAM_ENABLED,
+    UNPROCESSED_ITEMS, STREAM_SPECIFICATION, STREAM_VIEW_TYPE, STREAM_ENABLED, UPDATE_EXPRESSION,
     EXPRESSION_ATTRIBUTE_NAMES, EXPRESSION_ATTRIBUTE_VALUES, KEY_CONDITION_OPERATOR_MAP,
     CONDITION_EXPRESSION, FILTER_EXPRESSION, FILTER_EXPRESSION_OPERATOR_MAP, NOT_CONTAINS, AND)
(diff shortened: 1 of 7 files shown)


pynamodb/connection/base.py
```diff
@@ -866,20 +867,28 @@ def update_item(self,
...
 
-        operation_kwargs[ATTR_UPDATES] = {}
-        for key, update in attribute_updates.items():
-            value = update.get(VALUE)
-            attr_type, value = self.parse_attribute(value, return_type=True)
+        update_expression = Update()
+        # We sort the keys here for determinism. This is mostly done to simplify testing.
+        keys = list(attribute_updates.keys())
+        keys.sort()
+        for key in keys:
```
danielhochman: `for key in sorted(keys):`

jpinner-lyft: updated

pynamodb/connection/base.py
```diff
@@ -866,20 +867,27 @@ def update_item(self,
         if not attribute_updates:
             raise ValueError("{0} cannot be empty".format(ATTR_UPDATES))
 
-        operation_kwargs[ATTR_UPDATES] = {}
-        for key, update in attribute_updates.items():
-            value = update.get(VALUE)
-            attr_type, value = self.parse_attribute(value, return_type=True)
+        update_expression = Update()
+        # We sort the keys here for determinism. This is mostly done to simplify testing.
+        keys = list(attribute_updates.keys())
```
danielhochman: you no longer need the call to `list()`

jpinner-lyft: removed

pynamodb/connection/base.py
```diff
@@ -866,20 +867,26 @@ def update_item(self,
...
             if action not in ATTR_UPDATE_ACTIONS:
                 raise ValueError("{0} must be one of {1}".format(ACTION, ATTR_UPDATE_ACTIONS))
-            operation_kwargs[ATTR_UPDATES][key] = {
-                ACTION: action,
-            }
-            if action.upper() != DELETE:
-                operation_kwargs[ATTR_UPDATES][key][VALUE] = {attr_type: value}
+            value = update.get(VALUE)
+            attr_type, value = self.parse_attribute(value, return_type=True)
+            if attr_type is None and action != DELETE:
```
jmphilli: don't need the `.upper()` anymore?

pynamodb/expressions/update.py
```diff
@@ -0,0 +1,161 @@
...
+        if attr_type != NUMBER_SHORT:
+            raise ValueError("{0} must be a number".format(value))
+        super(DecrementAction, self).__init__(path, amount)
+
+
+class AppendAction(SetAction):
+    """
+    A SET action that appends elements to the end of a list.
+    """
+    format_string = '{path} = list_append({path}, {0})'
```
jmphilli: is `list_append` a dynamo primative / op?

jmphilli: it is

pynamodb/connection/base.py
jpinner-lyft: nope, the allowed values are all upper-case

pynamodb/expressions/update.py
jpinner-lyft: yeah -- i don't use it here yet but i thought it easiest to add all the allowed values

pynamodb/connection/base.py
```diff
@@ -1438,6 +1441,9 @@ def _check_condition(self, name, condition, expected_or_filter, conditional_oper
                 raise ValueError("'{0}' must be an instance of Condition".format(name))
             if expected_or_filter or conditional_operator is not None:
                 raise ValueError("Legacy conditional parameters cannot be used with condition expressions")
+        else:
```
lita: nit: this can be an `elif`

PR #344, opened 2017-08-29
Support setting attributes on "raw" MapAttribute instances.

jpinner-lyft:
This PR attempts to provide parity between "raw" MapAttributes and MapAttribute subclasses.

MapAttribute subclasses can set bound attribute values using the Attribute data descriptors.
Prior to this change, "raw" MapAttribute instances could only set name-value pairs during initialization.

With this change, all MapAttribute instances are instantiated as AttributeContainers and can access data stored on the internal attribute_values dictionary (either via the __getattr__ and __setattr__ methods in the case of "raw" MapAttributes, or via the Attribute data descriptors in the case of subclasses).

MapAttribute instances that are class objects of other AttributeContainers (Models or other MapAttribute subclasses) are "re-initialized" as Attributes by AttributeContainerMeta during class creation.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -8,6 +8,7 @@ from base64 import b64encode, b64decode
 from datetime import datetime
 from dateutil.parser import parse
 from dateutil.tz import tzutc
+from inspect import getargspec
 from pynamodb.constants import (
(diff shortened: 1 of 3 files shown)


jpinner-lyft:
This is a duplicate of https://github.com/pynamodb/PynamoDB/pull/342 which could not be merged due to coveralls being broken.

jpinner-lyft:
original "green" coverage run: https://coveralls.io/builds/13029174

PR #351, opened 2017-09-08
Reduce the surface area of the expressions api.

jpinner-lyft:


pynamodb/expressions/operand.py
```diff
@@ -40,86 +74,94 @@ def is_in(self, *values):
...
     """
     Value is an operand that represents an attribute value.
     """
+    format_string = '{0}'
 
     def __init__(self, value, attribute=None):
         # Check to see if value is already serialized
         if isinstance(value, dict) and len(value) == 1 and list(value.keys())[0] in SHORT_ATTR_TYPES:
-            self.value = value
+            (self.short_attr_type, value), = value.items()
```
jmphilli:
```python
(self.short_attr_type, value), _ = value.items()
```

PR #352, opened 2017-09-08
Complete Expression API

jpinner-lyft:


pynamodb/connection/base.py
```diff
@@ -839,6 +839,7 @@ def update_item(self,
                     table_name,
                     hash_key,
                     range_key=None,
+                    actions=None,
```
jmphilli: should we append this to the end instead of the middle for users that are using positional args?

pynamodb/connection/table.py
```diff
@@ -53,6 +53,7 @@ def delete_item(self, hash_key,
     def update_item(self,
                     hash_key,
                     range_key=None,
+                    actions=None,
```
jmphilli: same

pynamodb/connection/base.py
jpinner-lyft: Honestly I think this is ok.
While we could do that here we can't do that with `Model.update` because of how legacy conditionals are represented. If a user is using positional args it will fail with a ValueError instead of fail quietly.

PR #355, opened 2017-09-16
Add TTL Support (fixes #259)

### LINKED ISSUE #259: Time To Live (TTL) support?
Would be cool if we could set Time To Live (TTL) in the Meta options:

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html

brandond:


brandond:
Sorry - forgot to move this out of WIP. I've been using this in production for over a month and it's been working great. I'll rebase and then I think it's ready to go.

brandond:
I realized that I'd gotten lazy about Py26 compat. Fixing that stuff now. I'll also note that this is going to require a botocore requirement bump to ensure that the TTL methods are available.

pynamodb/__init__.py
```diff
@@ -7,4 +7,4 @@
 """
 __author__ = 'Jharrod LaFon'
 __license__ = 'MIT'
-__version__ = '3.2.1'
+__version__ = '3.2.2'
```
jpinner-lyft: as written this is a backwards-incompatible change for UTCDateTimeAttribute so this should not be included in a patch or minor release

brandond: It should be backwards-compatible in that it writes a different type but will happily read either - just like the boolean change a while back.

pynamodb/attributes.py
```diff
@@ -501,33 +508,108 @@ def deserialize(self, value):
...
+    attr_type = NUMBER
+    epoch = datetime.utcfromtimestamp(0).replace(tzinfo=tzutc())
+
+    def __init__(self, hash_key=False, range_key=False, null=None, default=None, attr_name=None, ttl=False):
+        super(UTCDateTimeAttribute, self).__init__(hash_key=hash_key,
+                                                   range_key=range_key,
+                                                   null=null,
+                                                   default=default,
+                                                   attr_name=attr_name)
+        self.is_ttl = ttl
```
jpinner-lyft: I'm assuming the answer is yes, but can Dynamo handle non-integer numbers in the TTL attribute?

pynamodb/compat.py
```diff
@@ -77,3 +77,10 @@ def getmembers_issubclass(object, classinfo):
             results.append((key, value))
     results.sort()
     return results
+
+
+# Replace this function with timedelta.total_seconds() once we drop Python 2.6 support
+def total_seconds(delta):
```
jpinner-lyft: this returns an integer not a floating point number

pynamodb/__init__.py
brandond: Oh, super fun, I hadn't seen that last time I was poking at the code. I can add migration code and docs as well, then.

Would you prefer I bump the version number in my PR, or just let you wait to merge this until next time you're going to do a major release?

pynamodb/compat.py
brandond: Indeed, I pulled it from cpython 3.6 without thinking. Good catch.

pynamodb/attributes.py
brandond: That behavior is not specifically addressed anywhere, but in my tests it simply ignores anything with a non-Number TTL. The following note comes from the feature announcement:

https://aws.amazon.com/blogs/aws/new-manage-dynamodb-items-using-time-to-live-ttl/
> The TTL attribute can be indexed or projected, but it cannot be an element of a JSON document. As I indicated earlier, it must have the Number data type. You can use IAM to regulate access to this attribute, just as you can do for any other one. Items that do not have the designated TTL attribute will not be considered for deletion. In order to avoid a possible accidental deletion due to a malformed TTL value, items that appear to be older than 5 years will not be deleted.

It appears that the implementation is via a background scan process. Records are deleted if the TTL field is:
* Present
* Numeric
* Not within a JSON document
* In the past
* Less than 5 years in the past

Anything else is ignored.

pynamodb/__init__.py
jpinner-lyft: If it's ok with you I'd prefer to just wait and merge as we're about to release 3.3 or 4.0

pynamodb/attributes.py
jpinner-lyft: Have you tested floating point "Numeric" attributes? (things like `ttl_attribute: {"N": "12345.678"}`)

pynamodb/compat.py
brandond: added division future to address this.

pynamodb/attributes.py
brandond: Yes - if you stick a datetime object with millisecond precision into that field, the resulting number is floating-point instead of integer. DynamoDB doesn't expire things with anything approaching millisecond granularity (it's somewhat 'best effort') but it will otherwise handle them correctly.

brandond:
@jpinner-lyft ready for re-review

pynamodb/attributes.py
```diff
@@ -501,33 +508,114 @@ def deserialize(self, value):
...
+
+    :param ttl: If True, this attribute is used as the TTL for purposes of automatic item expiry and deletion
+
+        Only a single attribute in a table may be used as the TTL. The first attribute with TTL enabled
+        will be used; repeated use of the flag will be ignored.
+
+        Note that expired items may persist up to two days after their TTL expires. See: 
+        https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateTimeToLive.html
+    """
+    attr_type = NUMBER
```
ButtaKnife: This change will break the existing API - any existing models using UTCDateTimeAttribute will be expecting the ISO-8601 formatted strings.  Couldn't we add this as a TTLDateTimeAttribute or EpochTimestampAttribute instead to be more explicit about the new functionality without breaking existing model behavior?

brandond: Like the Boolean update, it reads both the old and new formats, but writes the new one. The opposite behavior (reading both formats but writing the old one) is available from LegacyUTCDateTimeAttribute. I pretty much just copied the work that was done for the boolean switch.

jpinner-lyft: The boolean update could not be safely deployed. See https://github.com/pynamodb/PynamoDB/issues/404

hinnerk: @jpinner-lyft #404 is closed, does that mean that this is ready?

Surgo:
when release it?

mlovretovich:
@brandond any updates on this?

brandond:
I think I might have time to refactor this out to a new attribute class in the next week or two. Honestly it's not that work, most of it involves deleting the useless migration code.

ikonst:
I'd just introduce TTLAttribute since it has very specific semantics. It can map to a `datetime` in Python, but specifically it must be an NUMBER storing seconds since Unix epoch, and furthermore there can be just one per model (I think we can enforce that).

Generally for "nice to have" and complex attributes I'd recommend contributing to [pynamodb-attributes](https://github.com/lyft/pynamodb-attributes) instead, but this one can be legitimately considered a core concept since it maps to a DynamoDB backend concept.

PR #359, opened 2017-09-24
Update map attribute deserialization (fixes #358)

### LINKED ISSUE #358: Deserialization of model with MapAttribute requires all attributes to be declared
Currently, when a model is deserialized, any attributes which are not declared as attributes of the model are simply ignored. However, this is not the case for `MapAttribute`s. If the DynamoDB table has nested attributes within an `'M'` type attribute which are not represented in the corresponding `pynamodb.attributes.MapAttribute` subclass, an `AttributeError` is raised.

**Steps to reproduce**
``` python
from pynamodb.attributes import MapAttribute, NumberAttribute, UnicodeAttribute
from boto3 import resource
from decimal import Decimal


class Location(MapAttribute):
    lat = NumberAttribute()
    lng = NumberAttribute()
    name = UnicodeAttribute()

class Office(Model):
    class Meta:
        table_name = 'OfficeModel'
        host = "http://localhost:8000"

    office_id = NumberAttribute(hash_key=True)
    address = Location()

if not Office.exists():
    Office.create_table(read_capacity_units=1, write_capacity_units=1)

loc = Location(lat=-0.48591, lng=62.26732, name='MyLocation')
office = Office(office_id=123, address=loc)
office.save()

table = resource('dynamodb', endpoint_url='http://localhost:8000').Table('OfficeModel')
modified_office = {
    u'office_id': Decimal('123'),
    u'address': {
        u'foo': 'bar',
        u'latitude': Decimal('-0.48591'),
        u'name': u'MyLocation',
        u'longitude': Decimal('62.26732')
    }
}
table.put_item(TableName='OfficeModel', Item=modified_office)
x = list(Office.scan())
```
This raises the following error
```
  File "mapattribute.py", line 39, in <module>                                                                          
    x = list(Office.scan())                                                                                             
  File "/home/foo/.virtualenvs/pynamodb/lib/python2.7/site-packages/pynamodb/models.py", line 780, in  
scan                                                                                                                    
    yield cls.from_raw_data(item)                                                                                       
  File "/home/foo/.virtualenvs/pynamodb/lib/python2.7/site-packages/pynamodb/models.py", line 509, in  
from_raw_data                                                                                                           
    kwargs[attr_name] = attr.deserialize(attr.get_value(value))                                                         
  File "/home/foo/.virtualenvs/pynamodb/lib/python2.7/site-packages/pynamodb/attributes.py", line 511, 
 in deserialize                                                                                                         
    deserialized_value = attr_class.deserialize(attr_value)
AttributeError: 'NoneType' object has no attribute 'deserialize'
```

The only way to avoid this error is to explicitly declare each attribute in the `MapAttribute` subclass.
``` python
class Location(MapAttribute):
    lat = NumberAttribute()
    lng = NumberAttribute()
    name = UnicodeAttribute()
    foo = NullAttribute()
```
Ideally, in order to facilitate flexible `MapAttribute`s, additional, undeclared attributes would be ignored as they are in `Model`s.

mathewmarcus:
fixes #358

### DIFF
--- a/docs/attributes.rst
+++ b/docs/attributes.rst
@@ -192,3 +192,5 @@ These attributes can then be used inside of Model classes just like any other at
     class CarInfoMap(MapAttribute):
         make = UnicodeAttribute(null=False)
         model = UnicodeAttribute(null=True)
+
+`As with a model and its top-level attributes <https://github.com/pynamodb/PynamoDB/blob/master/docs/quickstart.rst#changing-items>`_, a PynamoDB MapAttribute will ignore sub-attributes it does not know about during deserialization. As a result, if the item in DynamoDB contains sub-attributes not declared as properties of the corresponding MapAttribute, save() will cause those sub-attributes to be deleted.
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -765,6 +765,8 @@ class MapAttribute(Attribute, AttributeContainer):
             attr_value = _get_value_for_deserialize(v)
             key = self._dynamo_to_python_attr(k)
             attr_class = self._get_deserialize_class(key, v)
+            if attr_class is None:
+                continue
             deserialized_value = None
             if attr_value is not None:
                 deserialized_value = attr_class.deserialize(attr_value)
--- a/pynamodb/tests/test_attributes.py
+++ b/pynamodb/tests/test_attributes.py
@@ -603,7 +603,7 @@ class TestMapAttribute:
         expected = {'number_attr': {'N': '10'}, 'unicode_attr': {'S': six.u('Hello')}}
         assert CustomAttrMap().serialize(attribute) == expected
 
-    def test_additional_atts_deserialize(self):
+    def test_additional_attrs_deserialize(self):
         raw_data = {
             'number_attr': {
                 'N': '10'},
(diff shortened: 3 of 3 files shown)


jpinner-lyft:
This is in some sense a breaking change -- albeit one I think we should make.
@garrettheel @scode do you guys concur?

@mathewmarcus do you mind adding some documentation on this behavior to the `Map Attributes` section of `docs/attributes.rst`

mathewmarcus:
No problem

scode:
So it's a win for consistency. The downside of silently ignoring data on reads is particularly bad in combination with `save()` as we just documented in https://github.com/pynamodb/PynamoDB/pull/397 but that's generally already a problem and the cleanest fix for that is probably to support tracking undeclared attributes and preserving them, rather refusing to de-serialize objects in the read path when encountering them. So even in that world, we would not want the current behavior.

Looking at the user impact, you'd have to be in a position of relying on the behavior for MapAttribute but not for anything else. It seems possible that someone is in that position, but it's very specific.

In favor of shipping with a very clear release note that calls out the breaking case. Possibly with a major version bump but that might be going overboard...

mathewmarcus:
@scode  should I add functionality to track/preserve undeclared attributes or just update the attributes documentation with a description of the current behavior of this change?

jpinner-lyft:
@mathewmarcus I would start by just updating the documentation -- properly doing unit-of-work tracking is a much more complicated process and there's the issue that people could currently be relying on `save()` removing undeclared attributes

PR #363, opened 2017-10-01
Add print_cloudformation method

alexdebrie:
First of all, this library is great. Thank you for doing the work to make this.

This PR adds a `print_cloudformation()` method to the `Model` object. I like to manage my infrastructure in CloudFormation, but it's tedious and error-prone to try to sync up my defined Model with the required CloudFormation. This method makes it so I can define my Model with PynamoDB but still manage the table with CloudFormation.

The implementation is pretty straight-forward as it's mostly following the `create_table` logic. I pulled out some of schema enrichment in the `create_table` method into a private method as I need to use it here. Then it's just a few simple functions to convert the schema object into the proper CloudFormation representation.

I updated the documentation in a few spots. I'm also willing to add tests, but I wanted to get approval on the addition and the approach before writing those out. It shouldn't break any existing tests, though I was only able to test Python 2.7 locally as I don't have the other interpreters installed.

Please let me know if you would like any changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,6 +57,13 @@ PynamoDB allows you to create the table if needed (it must exist before you can 
 
     UserModel.create_table(read_capacity_units=1, write_capacity_units=1)
 
+If you would rather manage your table via CloudFormation, you can print out your table representation:
+
+.. code-block:: python
+
+    # Default output is `json`. You can specify `yaml` or `yml` if you prefer.
+    UserModel.print_cloudformation(output='yml', read_capacity_units=1, write_capacity_units=1)
+
 Now, search your table for all users with a last name of 'Smith' and whose
 first name begins with 'J':
 
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -32,6 +32,10 @@ PynamoDB allows you to create the table:
 
     >>> UserModel.create_table(read_capacity_units=1, write_capacity_units=1)
 
+If you prefer to manage with CloudFormation, you can print the CloudFormation representation of your table:
+
+    >>> UserModel.print_cloudformation(read_capacity_units=1, write_capacity_units=1)
+
 Now you can create a user:
 
     >>> user = UserModel('test@example.com', first_name='Samuel', last_name='Adams')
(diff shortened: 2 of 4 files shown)


alexdebrie:
Hey @garrettheel, just checking in on this. Is this something you could see getting merged in, and are there any changes you'd like?

alexdebrie:
@jpinner-lyft I see you merging PRs for this repo. Any thoughts on whether we could get this merged in? Happy to make any changes necessary.

alexdebrie:
Great, thanks for the update @jpinner-lyft. Understand that it can be tough to keep up with OSS projects like this and appreciate your work on it.

Any particular concern with PyYAML or just wanting to limit the number of dependencies in general?

jpinner-lyft:
Not particular to PyYNAM -- just hesitant to increase the number of dependencies.

garrettheel:
Hey @alexdebrie, this looks really useful but I'm not sure that it belongs in PynamoDB itself given that we've tried to keep this pretty lean. What do you think about sticking this in a repo of your own and we can link to it from PynamoDB?

alexdebrie:
Hey @garrettheel, I guess that's fine if you don't want to add it in. Understandable if you all don't use CloudFormation.

I probably won't pull it out into a separate library given that it's pretty tightly tied to the internals of PynamoDB -- building up the entire schema and using a lot of Pynamo constants.

spockNinja:
This feature would be fantastic. There are lots of reasons to build all your aws resources through CloudFormation, and DynamoDB tables are no exception. I really like the pythonic interface that this repo provides, but it will currently be a maintenance nightmare to keep the python models in sync with the CloudFormation template.

stickystyle:
I'd like to +1 this feature as well. 
For my use case, I do a lot of work with the [serverless framework](https://github.com/serverless/serverless) where I would normally define the model once in cloudformation syntax so it gets created in a consistent manner with the rest of the stack, and then again in PynamoDB. 
It would be awesome if I could define the model in PynamoDB and have it exported to a YAML or JSON file that I could reference in my main serverless.yaml stack file (you can import external files into your config)

PR #367, opened 2017-10-07
Better handling of MapAttribute iteration and item assignment.

jpinner-lyft:
Now that we support nested element access in condition expressions using Python's `__getitem__` operator, Attributes are considered iterables (see PEP 234).

This commit prevents `iter()` calls on attributes (except for MapAttributes that are being used as AttributeContainers).

In addition, we add support for item assignment to MapAttributes to be symmetric with item access.

### DIFF
--- a/pynamodb/tests/test_attributes.py
+++ b/pynamodb/tests/test_attributes.py
@@ -606,6 +606,17 @@ class TestMapAttribute:
         assert item.map_attr['num'] == 3
         assert item.map_attr['nested']['nestedfoo'] == 'nestedbar'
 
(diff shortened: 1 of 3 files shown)


pynamodb/tests/test_attributes.py
```diff
@@ -606,6 +606,17 @@ def test_raw_set_attr(self):
         assert item.map_attr['num'] == 3
         assert item.map_attr['nested']['nestedfoo'] == 'nestedbar'
 
+    def test_raw_set_item(self):
```
garrettheel: Should we also have some tests for `iter`?

PR #370, opened 2017-10-12
 Fixed unwanted page read if limit is a multiple of page size.

jcbertin:
Also, added code fast path if no limit is given.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -608,8 +608,6 @@ class Connection(object):
         """
         Performs the ListTables operation
         """
-        if limit is not None and limit <= 0:
-            raise ValueError("Invalid 'limit' parameter")
         operation_kwargs = {}
         if exclusive_start_table_name:
             operation_kwargs.update({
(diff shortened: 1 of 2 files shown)


jcbertin:
Let me try to explain:
In Model.query, limit is checked before yielding so if limit is multiple of page size, the for loop terminates before limit testing. Then another query is done and the limit test bails the function.

Example flow:
```
page_size = None
limit = 1
data = query(limit=1) -> returns one item
for item in data.get(ITEMS):
    if limit == 0: -> nope
        return
    limit -= 1 -> limit == 0
    yield item -> for loop will terminate
while last_evaluated_key:
    data = query(limit=1) -> returns one item, *UNWANTED*
    for item in data.get(ITEMS):
        if limit == 0: -> *YES*, at least
            return
```

jpinner-lyft:
Result iteration has been re-worked as part of https://github.com/pynamodb/PynamoDB/pull/379 -- could you look at master and see if your concerns still apply?

jcbertin:
Issue addressed in #379.

PR #372, opened 2017-10-13
3.2.0

jpinner-lyft:


docs/release_notes.rst
```diff
@@ -1,34 +1,22 @@
 Release Notes
 =============
 
-v3.2.0rc2
```
jmphilli: these still exist... i think we need to leave them in

jpinner-lyft: I think we need to leave the tags -- I think it's fine to combine the release notes though since all the "release candidates" are still version `3.2.0`

jmphilli:
+1 as a stamp of approval. waiting for an rc test to pass first

Issue #207, opened 2016-12-08
Trying to Implementing Pagination using global secondary index


anush0247:
I am trying to implement pagination using last_evaluated_key & limit. I ran into below issues

```
from pynamodb.models import Model
from pynamodb.attributes import UnicodeAttribute, ListAttribute
from pynamodb.indexes import GlobalSecondaryIndex, AllProjection

class ServiceIdIndex(GlobalSecondaryIndex):
    """
    This class represents a global secondary index
    """
    class Meta:
        read_capacity_units = 1
        write_capacity_units = 1
        projection = AllProjection()
    service_id = UnicodeAttribute(hash_key=True)

class Channel(Model):
    """
    A DynamoDB User
    """
    class Meta:
        table_name = "dynamodb-user"
        host = "http://localhost:8000"
    name = UnicodeAttribute()
    channel_id = UnicodeAttribute(hash_key=True)
    service_id = UnicodeAttribute()
    members = ListAttribute()
    service_index = ServiceIdIndex()
    
for i in range(100):
    members = [
        {
            "username": "aneesh",
            "role": "admin",
        },
        {
            "username": "kumar",
            "role": "user",
        }
    ]
    channel2 = Channel(name="demo channel %d" % i, channel_id="Chn123%d" % i, service_id="Ser123", members=members)
    channel2.save()
    print channel2

```

1) How to get last_evaluated_key from channels generator 
2) index query giving query exception when we pass value for last_evaluated_key
```
channels = Channel.service_index.query("Ser123", scan_index_forward=False, last_evaluated_key=None, limit=10)
for each_channel in channels:
    print(each_channel)
```

Error Traceback: 

```
---------------------------------------------------------------------------
QueryError                                Traceback (most recent call last)
<ipython-input-66-2011f09bf20f> in <module>()
      1 channels = Channel.service_index.query("Ser123", scan_index_forward=False, last_evaluated_key="Serv123", limit=10)
----> 2 for each_channel in channels:
      3     print(each_channel)

/home/aneesh/.virtualenvs/chat/local/lib/python2.7/site-packages/pynamodb/models.pyc in query(cls, hash_key, consistent_read, index_name, scan_index_forward, conditional_operator, limit, last_evaluated_key, attributes_to_get, page_size, **filters)
    560         )
    561 
--> 562         data = cls._get_connection().query(hash_key, **query_kwargs)
    563         cls._throttle.add_record(data.get(CONSUMED_CAPACITY))
    564 

[...]

Issue #118, opened 2016-04-30
Need last_evaluated_key from last query


LiteWait:
I was wondering if there was a way to get the last_evaluated_key returned from the query method so I can implement paging on my Web client.

Edit: It seems query() can't take a last_evaluated_key either.  Hopefully someone has a solution or workaround before I decide to fork and doing something stupid.

Thanks.

danielhochman:
I think we can make a backwards compatible change to allow this functionality. We can have a `query_builder` method on the model which returns a `Query` object that exposes `last_evaluated_key` and other query arguments. We can have an `execute` method on `Query` that returns the result iterator. We can move all of the query logic to `Query` then have the current `query` method proxy to call `query_builder(*args, **kwargs).execute()`.

Then, to look at the `last_evaluated_key`, your code would look something like:

```
def get_page():
    q = MyModel.query_builder(hash_key='daniel', consistent_read=True)
    results = []
    for idx, result in enumerate(q.execute()):
        results.append(result)
        if idx == 99:
            return page, q.last_evaluated_key
    return page, None
```

Could also implement the `__iter__` and `next` methods directly and simplify this a bit. I'd have to think about it a bit more.

Thoughts?

LiteWait:
Great idea, sounds pretty clean.  Should you implement this happy to test it.  If not, I get the concept and when I get some time I'll give it a go.

danielhochman:
We should treat `scan` the same way, potentially with classes that inherit some behavior from a shared base class.

jlafon:
@danielhochman what about creating a `QueryResult` class with an `__iter__` method that is a generator (and thus backward compatible) but also has `last_evaluated_key` and other attributes?

danielhochman:
@jlafon yeah, I was leaning that way too. I might take a crack at this tomorrow.

jpinner-lyft:
Fixed in https://github.com/pynamodb/PynamoDB/pull/379

Issue #50, opened 2014-12-01
query with exclusive_start_key


mpresh:
Hello,
I am trying to run a table query by providing the last_evaluated_key int he following format 
exclusive_start_key': {'found': {'S': 'bla'}, 'id': {'S': u'67d54268-dab1-4c73-a424-dbe75d466bb5|subnet-69755941'}}
and I am receiving an error 
127.0.0.1 - - [01/Dec/2014 16:21:38] "GET /api/resources/?limit=25&env=67d54268-dab1-4c73-a424-dbe75d466bb5&last=67d54268-dab1-4c73-a424-dbe75d466bb5|subnet-69755941 HTTP/1.1" 500 -
....
QueryError: Failed to query items: {"__type":"com.amazon.coral.validate#ValidationException","message":"The provided starting key is invalid"}

Any idea if this is supported by Dynamodb. It works fine with scan but doesn't work with query.
I am tryign to figure out if this is a dynamodb issue, a pynamodb issue, or something with my code

Thanks for any help you can provide.
-Mike

jpinner-lyft:
Exclusive Start Key parameter requires hash and range keys of the item for both the table and the index -- it appears that `get_exclusive_start_key_map` doesn't handle this correctly.

jpinner-lyft:
Fixed in https://github.com/pynamodb/PynamoDB/pull/379

PR #379, opened 2017-10-17
Support pagination in high-level api query and scan methods.

jpinner-lyft:
This commit replaces the model's scan and query generator functions with an iterator object.
This address issues #50, #118, and #207.

The object exposes a `last_evaluated_key` property which allows the caller to perform pagination while using the high-level API:
```
results = []
started = False
last_evaluated_key = None
while not started or last_evaluated_key is not None:
    started = True
    results_iter = UserModel.query('foo', limit=50)
    results.append(list(results_iter))
    last_evaluated_key = results.last_evaluated_key
```

### DIFF
--- a/pynamodb/pagination.py
+++ b/pynamodb/pagination.py
@@ -1,0 +1,63 @@
+from pynamodb.constants import CAMEL_COUNT, ITEMS, LAST_EVALUATED_KEY
+
+
+class ResultIterator(object):
+    """
+    ResultIterator handles Query and Scan result pagination.
+
+    http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.Pagination
+    http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Scan.html#Scan.Pagination
+    """
+    def __init__(self, operation, args, kwargs, map_fn=None, limit=None):
+        self._operation = operation
+        self._args = args
+        self._kwargs = kwargs
+        self._map_fn = map_fn
+        self._limit = limit
+        self._needs_execute = True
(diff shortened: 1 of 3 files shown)


pynamodb/pagination.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if self._index == self._count:
+            raise StopIteration
+
+        item = self._items[self._index]
+        self._index += 1
+        if self._limit is not None:
+            self._limit -= 1
+        if self._map_fn:
+            item = self._map_fn(item)
```
garrettheel: This is great and sets us up to better handle the case where deserialization fails for a single row. Assume you did this intentionally and we can tackle that in a future PR. cc @danielhochman

pynamodb/pagination.py
```diff
@@ -0,0 +1,63 @@
+from pynamodb.constants import CAMEL_COUNT, ITEMS, LAST_EVALUATED_KEY
+
+
+class ResultIterator(object):
```
garrettheel: Do you think everything in here is adequately tested by everything we're already doing, or are there any gaps? I _think_ pretty much all of this should already be covered but just want to ask the question.

jpinner-lyft: I believe so -- we have test cases in model that handle different combinations of limits and result page sizes

pynamodb/pagination.py
jpinner-lyft: Yep, this approach also sets a foundation for providing access to capacity consumption during each iteration so we can use it as a baseline to build a more versatile rate limiting / throttle functionality.

PR #382, opened 2017-10-18
Added tags support.

jcbertin:


pynamodb/connection/base.py
```diff
@@ -634,6 +642,45 @@ def describe_table(self, table_name):
...
+        Performs the TagResource operation
+        """
+        tags_array = []
+        for key, value in tags.iteritems():
+            tags_array.append({KEY: key, VALUE: six.u(value)})
+        operation_kwargs = {RESOURCE_ARN: table_arn, TAGS: tags_array}
+        req_uuid = uuid.uuid4()
+
+        self.send_pre_boto_callback(TAG_RESOURCE, req_uuid, table_name)
+        self._make_api_call(TAG_RESOURCE, operation_kwargs)
```
jcbertin: Cannot use Connection.dispatch because it expects a table name in operation parameters.

PR #383, opened 2017-10-20
Fix settings

jcbertin:
This PR offers a cleaner API for settings monkey-patching.

Actually, the settings are set at loading time in the Model meta class by reading Model.Meta attributes. This is fine for rapid development but can become tedious when you have a lot of tables. Moreover, its quite difficult to implement a clean way to set those values at runtime (think about a lambda function which can be deployed in different regions and have to access dynamodb locally).

With this patch, settings are read from a pynamodb.Settings instance only at runtime when Connection class is instantiate. This allow users to monkey-patch settings before they are used. You can do this in two ways:

1. by adding a Model.Meta.settings attribute which must be an instance of Settings class.
2. by setting a default Setting instance globally in the DefaultSettingsDescriptor.settings class attribute.

I also included the thread-safe patch from AusIV.


jcbertin:
Unit tests for this feature is under way…

ButtaKnife:
This would be great to have if it could be done against #393 rather than #234.

PR #386, opened 2017-10-24
Deserialize hash key during table loads. Fixes #143

### LINKED ISSUE #143: Deserialize hash key while loading from json
While trying to load a table from a json dump, https://github.com/jlafon/PynamoDB/blob/devel/pynamodb/models.py#L764 does not deserialize the hash key. Hence, we run into an exception if the hash key is anything other than a UnicodeAttribute. 

For example, the following will lead to an exception.

```
    UnicodeAttribute, NumberAttribute
)

class Thread(Model):
    class Meta:
        table_name = 'Thread'

    forum_name = NumberAttribute(hash_key=True)
    subject = UnicodeAttribute(range_key=True)
    views = NumberAttribute(default=0)

# After the table has been populated

Thread.dump("thread_backup.json")
Thread.load("thread_backup.json")
```

jpinner-lyft:


pynamodb/tests/data.py
```diff
@@ -466,7 +466,7 @@
                     'N': '31'
                 },
                 'is_dude': {
-                    'N': '1'
+                    'BOOL': True
```
jmphilli: i think i may have been using this to test the old legacyboolattr, but if we remove that this is fine

jpinner-lyft: there are tests specific to legacy deserialization `TestLegacyBooleanAttribute`

PR #387, opened 2017-10-24
Model.loads() should deserialize using custom attribute names. Fixes #168

### LINKED ISSUE #168: Custom attribute names don't work properly with Model.loads()
In tests, custom attribute name (`attr_name` parameter) are tested only for hash key. But they won't work for other attributes, because `Model._deserialize()` looks up "custom" (pythonic) attr name and doesn't check `attr.attr_name`, in contrary to `Model._serialize()` which uses `aliased_attrs()` for lookup.

jpinner-lyft:


jmphilli:
thx

PR #395, opened 2017-10-27
Support limiting the total number of items evaluated by a query or scan operation.

jpinner-lyft:
Support limiting the total number of items evaluated by a query or scan operation (replaces #315).
This gives users better control over the capacity consumed by these operations, e.g.:
```
class PageLimitedIterator(object):

    def __init__(self, page_iterator, max_scanned_count):
        self._iter = page_iterator
        self._max_scanned_count = max_scanned_count

    def __iter__(self):
        return self

    def __next__(self):
        limit = self._max_scanned_count - self._iter.total_scanned_count
        if limit == 0:
            raise StopIteration()
        if limit < self._iter.page_size:
            self._iter.page_size = limit
        return next(self._iter)

    def next(self):
        return self.__next__()

result_iterator = Model.scan()
result_iterator.page_iter = PageLimitedIterator(result_iterator.page_iter, 10)
for item in result_iterator:
    # pass
```

### DIFF
--- a/pynamodb/pagination.py
+++ b/pynamodb/pagination.py
@@ -1,27 +1,75 @@
-from pynamodb.constants import CAMEL_COUNT, ITEMS, LAST_EVALUATED_KEY
+from pynamodb.constants import CAMEL_COUNT, ITEMS, LAST_EVALUATED_KEY, SCANNED_COUNT
+
+
+class PageIterator(object):
+    """
+    PageIterator handles Query and Scan result pagination.
+
+    http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.Pagination
+    http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Scan.html#Scan.Pagination
+    """
+    def __init__(self, operation, args, kwargs):
+        self._operation = operation
+        self._args = args
+        self._kwargs = kwargs
+        self._first_iteration = True
+        self._last_evaluated_key = None
+        self._total_scanned_count = 0
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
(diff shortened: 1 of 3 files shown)


pynamodb/pagination.py
```diff
@@ -16,14 +16,17 @@ def __init__(self, operation, args, kwargs, map_fn=None, limit=None):
...
         self._needs_execute = True
         self._total_count = 0
+        self._total_scanned_count = 0
 
     def _execute(self):
         data = self._operation(*self._args, **self._kwargs)
         self._count = data[CAMEL_COUNT]
         self._items = data.get(ITEMS)  # not returned if 'Select' is set to 'COUNT'
         self._last_evaluated_key = data.get(LAST_EVALUATED_KEY)
+        self._scanned_count = data[SCANNED_COUNT]
```
jmphilli: do we really need this as an instance var?

jpinner-lyft:
:lock: this still doesn't quite do all the things I want it to -- going to take another pass :lock:

pynamodb/pagination.py
```diff
@@ -1,27 +1,75 @@
-from pynamodb.constants import CAMEL_COUNT, ITEMS, LAST_EVALUATED_KEY
+from pynamodb.constants import CAMEL_COUNT, ITEMS, LAST_EVALUATED_KEY, SCANNED_COUNT
 
 
-class ResultIterator(object):
+class PageIterator(object):
```
jmphilli: you don't think people are using the class as a type hint or something in py3? we reserving this for 4.0?

jpinner-lyft: the `ResultIterator` class is still there -- the diff is just misleading -- look below

Issue #404, opened 2017-11-07
LegacyBooleanAttribute -> BooleanAttribute upgrade path needs data migration, work-arounds


scode:
*NOTE:* This issue will likely be edited further over the course of the next day or three.

The [release notes for 1.6.0](http://pynamodb.readthedocs.io/en/latest/release_notes.html#v1-6-0) contain instructions around the use of `LegacyBooleanAttribute` as a substitute for `BooelanAttribute` to allow a safe upgrade path despite a breaking serialization change: BooleanAttribute as of 1.6.0 stores booleans as a native boolean data type in DynamoDB, instead of an integer data type. The upgrade path was intended to be approximately:

* Upgrade to 1.6.0+ and switch to LegacyBooleanAttribute atomically. Roll that out everywhere. This causes readers to support both integers and booleans in DynamoDB.
* Switch to BooleanAttribute and roll that out everywhere. This still supports both integers and booleans in DynamoDB, and will write them as booleans.

At this point, you were supposed to be good to go. However, you still have a mix of data types in DynamoDB itself unless you actively re-write all objects. Further more, the `query()` and `scan()` APIs use of `filter_condition` breaks down and inappropriately filters out results. 

For example, given this model:

```
class UserModel(Model):
    class Meta:
        table_name = "dynamodb-users"
        host = "http://localhost:8000"
    username = UnicodeAttribute(hash_key=True)
    is_blue = LegacyBooleanAttribute()
```

And creating an object:

```
next_user = UserModel('janedoe', is_blue=True)
next_user.save()
```

Followed by moving to BooleanAttribute:

```
class UserModel(Model):
    class Meta:
        table_name = "dynamodb-users"
        host = "http://localhost:8000"
    username = UnicodeAttribute(hash_key=True)
    is_blue = LegacyBooleanAttribute()
```

And attempting to find the object:

```
for user in UserModel.query('janedoe', (UserModel.is_blue == True)):
    print('found nonpath', user.username, user.is_blue)
```

Will "miss" the object because `True != 1`. The work-around is to perform the scan as such (note that we both (a) switch to using `Path()`, and (b) cover both the True and 1 case; merely doing (b) without (a) is insufficient):

```
for user in UserModel.query('janedoe', (Path('is_blue') == True) | (Path('is_blue') == 1)):
    print('found path', user.username, user.is_blue)
```

We will provide some assistance for folks wanting to upgrade their data.

PR #405, opened 2017-11-07
Add migration support for LegacyBooleanAttribute.

scode:
This is in support of #404.

This also adds our first integration test with DynamoDB Local getting pulled in
by travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,12 @@ before_install:
 install:
   - pip install -r requirements-dev.txt
 
+before_script:
+  - wget http://dynamodb-local.s3-website-us-west-2.amazonaws.com/dynamodb_local_latest.tar.gz -O /tmp/dynamodb_local_latest.tar.gz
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -17,6 +17,12 @@ before_install:
 install:
   - pip install -r requirements-dev.txt
 
+before_script:
+  - wget http://dynamodb-local.s3-website-us-west-2.amazonaws.com/dynamodb_local_latest.tar.gz -O /tmp/dynamodb_local_latest.tar.gz
+  - tar -xzf /tmp/dynamodb_local_latest.tar.gz -C /tmp
+  - java -Djava.library.path=/tmp/DynamoDBLocal_lib -jar /tmp/DynamoDBLocal.jar -inMemory -port 8000 &
+  - sleep 2
```
scode: This is obviously flaky btw, but I felt it was ok. We could increase the number a bit, but I'm not too worried about having something here that's fundamentally flaky.

PR #410, opened 2017-11-10
Return the correct last_evaluated_key for limited queries/scans. Fixes #406

### LINKED ISSUE #406: Last Evaluated Key and Limited Queries/Scans
The `last_evaluated_key` value returns whatever was last evaluated for the current result page. When limiting the query but not setting the page size to the limit, the value does not represent the last item returned by the iterator.

jpinner-lyft:


pynamodb/pagination.py
```diff
@@ -100,7 +100,39 @@ def next(self):
...
+        # This can occur if the 'limit' is reached in the middle of a page.
+        item = self._items[self._index - 1]
+        last_evaluated_key = {}
+        if self.page_iter.last_evaluated_key:
+            # If the current page has a last_evaluated_key, use it to determine key attributes
+            for key in self.page_iter.last_evaluated_key.keys():
+                last_evaluated_key[key] = item[key]
+        else:
+            # Use the table meta data to determine the key attributes
+            table_connection = self.page_iter._operation.im_self
```
jmphilli: should we write a private method to expose this? do we need this other places in the file?

pynamodb/pagination.py
```diff
@@ -100,7 +100,39 @@ def next(self):
...
+            table_meta = table_connection.connection.get_meta_table(table_connection.table_name)
```
jmphilli: same

pynamodb/connection/base.py
```diff
@@ -95,6 +95,23 @@ def hash_keyname(self):
                     break
         return self._hash_keyname
 
+    def get_key_names(self, index_name=None):
+        """
+        Returns the names of the primary key attributes and index key attributes (if index_name is specified)
+        """
+        key_names = [self.hash_keyname]
+        range_keyname = self.range_keyname
```
jmphilli:
```python
if self.range_keyname:
    key_names.append(self.range_keyname)
```

pynamodb/connection/base.py
```diff
@@ -95,6 +95,23 @@ def hash_keyname(self):
...
+        if range_keyname is not None:
+            key_names.append(range_keyname)
+        if index_name is not None:
```
jmphilli: is there not more than one index?

pynamodb/connection/base.py
jpinner-lyft: that would require us to search for it twice if it doesn't exist

pynamodb/connection/base.py
jpinner-lyft: you need to get the keys on a per-index basis since they are different for each index

pynamodb/connection/base.py
jmphilli: search? twice? it's only used once in this method.

pynamodb/connection/base.py
jmphilli: i think that's what i'm asking... shouldn't there be a loop here?

jpinner-lyft: the function (as written) returns keys for a specific index (or for the table if no index is specified), not all combinations for all indexes

pynamodb/connection/base.py
jpinner-lyft: self.range_keyname is a property computed by scanning the metadata -- its memoized, but only if the value is not `None` -- if it is `None`, the if statement will search and then ....

actually you're right, if it's None you won't enter the if block -- updating

pynamodb/connection/base.py
```diff
@@ -95,6 +95,22 @@ def hash_keyname(self):
                     break
         return self._hash_keyname
 
+    def get_key_names(self, index_name=None):
```
jmphilli: can we add some tests for this

Issue #361, opened 2017-09-27
Support for inheritance between models


eapolinario:
Right now it is not possible to have a model `Foo` inherit from model `Bar`, i.e.:

```python
class Bar(Model):
    class Meta:
        table_name = 'bar_table'
...

class Foo(Bar):
    class Meta:
        table_name = 'foo_table'
...
```

Apparently the connection to the underlying tables is cached between these two models, which causes accesses to `Foo` to refer to the table 'bar_table' instead of 'foo_table'.

helveticafire:
@eapolinario have a look at https://github.com/pynamodb/PynamoDB/issues/164 which should shed some light on what you are suggesting.

PR #412, opened 2017-11-10
Support model class inheritence. Fixes #164, #361

### LINKED ISSUE #164: Inheriting from Model - should this work?
Hi,

Should it be possible to sub-class a pynamodb Model?  I'm getting the following error:

```
AttributeError: As of v1.0 PynamoDB Models require a `Meta` class with a table_name property.
                See https://pynamodb.readthedocs.io/en/latest/release_notes.html
```

MRE to reproduce (Python 2.7.12, pynamodb-1.5.3):

``` python

class SomeBase(Model):

    class Meta:
        table_name='table-one'
        region     = 'us-west-2'

    count = NumberAttribute(default=0)


class VariantA(SomeBase):

    def __init__(self, *args, **kwargs):
        super(VariantA, self).__init__(*args, **kwargs)


if __name__ == '__main__':
    some_a = VariantA()

    #print "Meta attribute present: {}".format(hasattr(some_a, 'Meta'))
    #print dir(some_a)

    some_a.email = 'user@example.com'
    some_a.count = 666
    some_a.save()
```

jpinner-lyft:
:lock: work in progress :lock:

`DefaultMeta` is removed so that a model's `Meta` class is inherited from the parent class.

Note: this is technically a **backwards-incompatible** change for those users who relied on dynamically setting attributes of their Meta class without creating it first.

The `_meta_table` value cached on the model is removed in favor of the value cached on the connection (which is cached per-table-name) to support model subclasses with different table names.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -11,7 +11,6 @@ import warnings
 from six import add_metaclass
 from pynamodb.exceptions import DoesNotExist, TableDoesNotExist, TableError
 from pynamodb.attributes import Attribute, AttributeContainer, AttributeContainerMeta, MapAttribute, ListAttribute
-from pynamodb.connection.base import MetaTable
 from pynamodb.connection.table import TableConnection
 from pynamodb.connection.util import pythonic
 from pynamodb.types import HASH, RANGE
@@ -145,21 +144,6 @@ class BatchWrite(ModelContextManager):
             unprocessed_items = data.get(UNPROCESSED_ITEMS, {}).get(self.model.Meta.table_name)
 
 
-class DefaultMeta(object):
-    pass
-
-
-class ResultSet(object):
-
-    def __init__(self, results, operation, arguments):
-        self.results = results
-        self.operation = operation
-        self.arguments = arguments
-
-    def __iter__(self):
-        return iter(self.results)
-
-
 class MetaModel(AttributeContainerMeta):
     """
     Model meta class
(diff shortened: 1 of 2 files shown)


pynamodb/models.py
```diff
@@ -145,21 +144,6 @@ def commit(self):
             unprocessed_items = data.get(UNPROCESSED_ITEMS, {}).get(self.model.Meta.table_name)
 
 
-class DefaultMeta(object):
-    pass
-
-
-class ResultSet(object):
```
jpinner-lyft: this is unused

pynamodb/models.py
```diff
@@ -145,21 +144,6 @@ def commit(self):
             unprocessed_items = data.get(UNPROCESSED_ITEMS, {}).get(self.model.Meta.table_name)
 
 
-class DefaultMeta(object):
```
jpinner-lyft: this is leftover from when default values were stored in this object

pynamodb/models.py
```diff
@@ -1248,9 +1228,7 @@ def _get_meta_data(cls):
         """
         A helper object that contains meta data about this table
         """
-        if cls._meta_table is None:
-            cls._meta_table = MetaTable(cls._get_connection().describe_table())
-        return cls._meta_table
+        return cls._get_connection().get_meta_table()
```
jpinner-lyft: returns the value cached on the base connection instead of the model

jpinner-lyft:
Open Question: say I have

```
ModelA(Model):
    class Meta:
        class table_name = 'foo'

ModelB(ModelA):
    pass
```
with this change, `ModelB.Meta.table_name == 'foo'`

Now if i do `ModelA.Meta.table_name = 'bar'` what should `ModelB.Meta.table_name` be? 
Or vice-versa, if i do `ModelB.Meta.table_name = 'bar'` should that change `ModelA`?

PR #413, opened 2017-11-16
Make model deserialization public

Dunedan:
Allows manual creation of model instances from DynamoDB JSON. Useful
e.g. when working with DynamoDB Streams.


jpinner-lyft:
:lock: we should fix / unify `_deserialize` and `_from_data` with MapAttribute's `_deserialize` first

Dunedan:
As long as the use case is understood (being able to generate PynamoDB model instances from DynamoDB-returned JSON, like it's the case when processing DynamoDB Streams) I don't mind how the solution looks to make it possible. 😃 
The approach outlined in this PR just seemed to be the easiest way to achieve it to me.

PR #419, opened 2017-11-21
Don't share connection when model inherits from another

kerrick-lyft:


jpinner-lyft:
:lock: pending model inheritance support for 3.3.0 :lock:

PR #421, opened 2017-11-28
Fix exclusive_start_key getting lost in PageIterator

drewisme:
The recent addition of PageIterator in #395 has muted the `last_evaluated_key` to query and scan methods.

This prevents manually getting results for a specific page.

### DIFF
--- a/pynamodb/pagination.py
+++ b/pynamodb/pagination.py
@@ -111,6 +111,7 @@ class ResultIterator(object):
     @property
     def last_evaluated_key(self):
         if self._first_iteration or self._index == self._count:
+            # Not started iterating yet: return `exclusive_start_key` if set, otherwise expect None; or,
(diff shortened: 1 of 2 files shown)


pynamodb/pagination.py
```diff
@@ -110,11 +110,7 @@ def next(self):
 
     @property
     def last_evaluated_key(self):
-        if self._first_iteration:
-            # Not started iterating yet: there cannot be a last_evaluated_key
-            return None
-
-        if self._index == self._count:
+        if self._first_iteration or self._index == self._count:
             # Entire page has been consumed: last_evaluated_key is whatever DynamoDB returned
```
jpinner-lyft: Can you update this comment to include why we fallback to the page iterator if `self._first_iteration` is true

drewisme:
That was crazy fast. Thanks!

PR #423, opened 2017-12-05
Add rate limiting parameters to migrate_boolean_attributes migration.

betamoo:


pynamodb/migration.py
```diff
@@ -5,6 +5,7 @@
 
 import logging
 
+from time import sleep
```
harleyk: I'd write this as just import time and do time.sleep instead.

jpinner-lyft:
preferable would be to remove rate_limited_scan entirely in favor of adding rate limiting to PageIterator

harleyk:
@jpinner-lyft That would be great but should that be above the migration?

jpinner-lyft:
@harleyk likely not if you all are blocked on this

pynamodb/migration.py
```diff
@@ -31,6 +32,9 @@ def _build_lba_filter_condition(attribute_names):
 def migrate_boolean_attributes(model_class,
                                attribute_names,
                                read_capacity_to_consume_per_second=10,
+                               page_size=None,
+                               limit=None,
+                               back_off=1,
```
jpinner-lyft: nit: if we've already published a release with `migrate_boolean_attributes` defined then move the new kwargs to the end

betamoo:
😽

pynamodb/migration.py
```diff
@@ -79,15 +83,21 @@ def migrate_boolean_attributes(model_class,
                                             meant to trigger the code path in boto, to allow us to unit test that
                                             we are jumping through appropriate hoops handling the resulting
                                             failure and distinguishing it from other failures.
+    :param page_size: Passed along to the underlying 'page_size'. Page size of the scan to DynamoDB.
+    :param limit: Passed along to the underlying 'limit'. Used to limit the number of results returned.
+    :param back_off: Number of seconds to sleep when exceeding capacity.
```
fandan-nyc: nit pick: 
IMHO, i think back_off is not a clear variable name. 
I personally suggest using something more descriptive such as `sec_sleep_when_exceed_cap` or anything which can clearly tell people what this variable is.

pynamodb/migration.py
```diff
@@ -79,15 +83,21 @@ def migrate_boolean_attributes(model_class,
...
     num_update_failures = 0
+    items_processed = 0
 
     for item in model_class.rate_limited_scan(_build_lba_filter_condition(attribute_names),
                                               read_capacity_to_consume_per_second=read_capacity_to_consume_per_second,
+                                              page_size=page_size,
+                                              limit=limit,
                                               allow_rate_limited_scan_without_consumed_capacity=allow_rate_limited_scan_without_consumed_capacity):
         actions = []
         condition = None
```
fandan-nyc: something super irrelevant that I wanna bring up here, correct me if you disagree. 
it feels the section between 103 to 139 can use some help of a helper function. 
separating the logic into smaller funcs always improve the readability.
since we are touching this piece of code anyway, do we want to spend a little bit more time to refactor this really quick ? 
it is really hard to read through and understand what is going on.

pynamodb/migration.py
```diff
@@ -119,10 +129,17 @@ def migrate_boolean_attributes(model_class,
                     if code == 'ConditionalCheckFailedException':
                         log.warn('conditional update failed (concurrent writes?) for object: %s (you will need to re-run migration)', item)
                         num_update_failures += 1
+                    elif code == 'ProvisionedThroughputExceededException':
```
fandan-nyc: it looks like the we have a group of code to check. let's take advantage of enum to avoid typo and the ease of reuse if ever.

pynamodb/migration.py
betamoo: I do agree that the code could use some comments, but I find the two parts of the main function : 1-build action list & 2- if actions, do them, are pretty minimal. Also this is -kinda of tmp- migration code, that we need to run soon on production machines, we could plan this for later maybe..

pynamodb/migration.py
betamoo: I am not sure if we have control over this (it seems like it's related to boto library/pynamo interaction, and/or code wide change) I think it's out of scope of this PR

pynamodb/migration.py
fandan-nyc: well, since this is tmp and will be deprecated soon ( I assume ), it is fine to let it be like this.

pynamodb/migration.py
fandan-nyc: well, since this is tmp migration code and will NOT be in prod anyway, i will let it be, so we can save some time and get this out asap

fandan-nyc:
+1 and good to go.

PR #425, opened 2017-12-07
Adding Rate limit for items updated per second 

betamoo:


betamoo:
PTAL @harleyk

betamoo:
PTAL @scode

PR #426, opened 2017-12-11
Externalizes AWS Credentials

nicysneiros:
Hi! I'm currently working with microservice using Zato and we have all of our config settings in a specific place (not the same as the default files expected by botocore). We have made our integration with Dynamo using boto3 (because it allows us to set the AWS credentials when we create a session) but we are looking forward to migrate to PynamoDB and setting the credentials on the fly is a requirement...

For this, I externalized the AWS credentials attributes in the Model's MetaClass and passed them along to the TableConnection to set the session credentials. This solution works for me and now I'm using this forked version. I decided to open this PR as a way to discuss if this is a valid solution or if we could improve it (I'm not a dynamo expert whatsoever). Let me know your thoughts! 😃

### DIFF
--- a/docs/awsaccess.rst
+++ b/docs/awsaccess.rst
@@ -12,5 +12,16 @@ policies as required, see the
 `DynamoDB <http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/UsingIAMWithDDB.html>`_ docs for more
 information.
 
+If for some reason you can't use conventional AWS configuration methods, you can set the credentials in the Model Meta class:
+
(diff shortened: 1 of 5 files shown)


edouardpoitras:
This is exactly what I was looking for! Thanks @nicysneiros

nicysneiros:
@garrettheel Updated PR with new tests and documentation for the new feature. Locally the tests are passing for all python versions required, but the Travis build is not passing and I'm not sure why... Can you help me?

PR #433, opened 2018-01-17
Fix proxies setting not being used

ubill:
I have found that class Model -> Meta:session_cls' proxy settings (proxies of the requests.Session) are not being taken into use.

Hence, when making a dynamodb request with the Model - the proxy is actually not being used.

This is a fix for that. Turn on debug level logging to confirm.


ubill:
The Travis CI tests failed for Python 3.3 with the same reason as https://github.com/pynamodb/PynamoDB/pull/426 . :(

False positive test result?

Issue #439, opened 2018-01-26
Supporting Model Inheritance in Pynamo


betamoo:
*Last Modified: 01/25/2018*

# Summary

[PynamoDB](https://github.com/pynamodb/PynamoDB) currently doesn’t -guarantee to- support inheritance. We aim at supporting two different types of inheritance in Pynamo: Abstract & Proxy inheritance (explained later) in order to allow more flexible ways of designing/describing models for Pynamo users.

# Goal

Support two types of inheritance Abstract & Proxy inheritance.

Non-goal: supporting Multi table inheritance.

# Context

[PynamoDB](https://github.com/pynamodb/PynamoDB) allows describing, creating and accessing tables in DynamoDB through pythonic styled methods. Creating tables in Pynamo is -mostly- declarative: in that by declaring python classes, the user can create & get access to DynamoDB tables and data.

Model inheritance is a feature that is supported by some ORMs, such as: [Django](https://www.djangoproject.com/)

The three types of inheritance supported in Django:

>**1. Abstract Inheritance**
>In which, the parent (inherited model) is an abstract model that doesn’t correspond to a table, but rather just a definition or an information holder for the children (inheriting models) to follow.  
>
>A use case of abstract inheritance is when more than one model share common field descriptions, the user can have an abstract class with the common fields and inherit it by different children, that way the user saves work that that otherwise would be redundant.

>**2. Proxy Inheritance**
>In this case, the inheriting child (or children) provides) only an alternate way to access the same table of the parent model. Both the proxies and their parent correspond to the same table but can have different configurations or different behaviours. 
>
>A use case for proxy inheritance is when a user needs to access the same table with two or more different configurations (different port or different RCU/WCU).  Another use case is to mask certain fields in the child proxy such that it only has access to a subset of  it’s parent fields.

>**3. Multi table inheritance**
>In which both the parent and each of the children represent different tables, but with links between the common fields and attributes.  That means that some fields of corresponding models are mirrored in the different tables and have to stay in sync (including for example cascade delete).
>
>A use case for multi table inheritance, is when the user needs to create a table that shadows the original table. The inheritance relation should take care of the data replication.

# Plan

As mentioned earlier, we are aiming only to support **abstract inheritance** and **proxy inheritance.** 

We are going to **skip** the multi table inheritance (at least for now) as it is almost impossible/will introduce many complications in a non relational key-value database like DynamoDB.

We are planning to approach the inheritance behaviour as follows:

[...]

PR #440, opened 2018-01-29
Adding support for absolute inheritance in Pynamo

betamoo:
Changes are described in https://github.com/pynamodb/PynamoDB/issues/439


pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
...
+                    setattr(meta_class, 'base_backoff_ms', get_settings_value('base_backoff_ms'))
+                if not hasattr(meta_class, 'max_retry_attempts'):
+                    setattr(meta_class, 'max_retry_attempts', get_settings_value('max_retry_attempts'))
+                if not hasattr(meta_class, ABSTRACT):
+                    setattr(meta_class, ABSTRACT, False)
+                is_abstract = getattr(meta_class, ABSTRACT, None)
+                if hasattr(meta_class, 'table_name'):
+                    has_table_name = True
+            else:
+                # TODO: decide what to do with DefaultMeta
```
betamoo: should we ban this behaviour? Currently I left it allowed.

pynamodb/exceptions.py
```diff
@@ -78,6 +78,20 @@ class DoesNotExist(PynamoDBException):
...
+    Raised when an operation is not allowed on abstract model
+    """
+    msg = "Operation is not allowed on abstract model"
+
+
+class InheritanceError(PynamoDBException):
+    """
+    Raised when violating model inheritance rules
+    """
+    msg = "Rules of model inheritance has been violated"
```
jpinner-lyft: nit: "have been violated"

pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
     """
     def __init__(cls, name, bases, attrs):
         super(MetaModel, cls).__init__(name, bases, attrs)
+
+        # We don't' allow inheritance from multiple direct parents
+        if (len(bases) > 1):
+            raise InheritanceError("Model can't inherit from multiple direct parents")
+
```
jpinner-lyft: This isn't a realistic restriction we can place on all classes that extend Model.
As far as I can tell this would break anyone using Mixins: e.g. `class MyModel(Model, Loggable)`

pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
     """
     def __init__(cls, name, bases, attrs):
         super(MetaModel, cls).__init__(name, bases, attrs)
+
+        # We don't' allow inheritance from multiple direct parents
+        if (len(bases) > 1):
+            raise InheritanceError("Model can't inherit from multiple direct parents")
```
jpinner-lyft: This isn't a feasible restriction we can place on uses of the Model class.
As far as I can tell this would break any code using Mixins: e.g. `class MyModel(Model, Loggable)`

pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
...
+
+        parent = None if len(bases) == 0 else bases[0]
+        # TODO: special parents as constants
+        parent_is_attribute_container = parent.__name__ == 'AttributeContainer'
+        parent_is_model = parent.__name__ == 'Model'
```
jpinner-lyft: seems like these should be done using `type()` checks

pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
...
+        parent_meta_class = None
+        is_parent_abstract = None
+        if parent  and not parent_is_attribute_container:
+            parent_meta_class = getattr(parent, META_CLASS_NAME, None)
+        if parent_meta_class and hasattr(parent_meta_class, ABSTRACT):
+            is_parent_abstract = getattr(parent_meta_class, ABSTRACT)
+
+        # Temporary check until implementing other inheritance types
```
jpinner-lyft: It seems like the thing to do here is to iterate through bases and check if any of them are `Model` types, and if they are, throw an exception if any of them don't have the `abstract=true` field in their `Meta` class

pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
...
+        if parent and not parent_is_attribute_container and not parent_is_model and not is_parent_abstract:
+            raise NotImplementedError("Inheritance from non abstract models is not supported now")
+
         if isinstance(attrs, dict):
+            meta_class = attrs.get(META_CLASS_NAME, None)
```
jpinner-lyft: open question now that we're starting to allow inheritance:
Do we allow models to inherit the `Meta` class from their parent class(es) or do we require each Model define their own `Meta` class? If the latter, how do we plan on handling `DefaultMeta` since we currently allow Models that do not define `Meta` at all?

pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
...
+
+            # Check special parents to avoid copying DefaultMeta
+            if parent and not parent_is_attribute_container and not parent_is_model:
+                if not meta_class:
+                    # Prepare an empty meta class
+                    meta_class = type(META_CLASS_NAME, (), {})
+                    setattr(cls, META_CLASS_NAME, meta_class)
+
+                if parent_meta_class:
+                        # Copying missing meta attributes from parent(s)
```
jpinner-lyft: If we are going to go this route, it seems like the easier thing would just be to rely on python's inheritance rules and access the meta class through the passed-in attributes.

pynamodb/models.py
jpinner-lyft: Note: if we ban this behavior, this would be a breaking change and we'd have to release a new major version of Pynamo.

pynamodb/models.py
```diff
@@ -169,40 +171,104 @@ class MetaModel(AttributeContainerMeta):
...
+
+            if is_abstract is True:
+                if has_table_name:
+                    raise cls.NotAllowedWhenAbstract("Abstract model can't have a table name")
+                if has_indexes:
+                    raise cls.NotAllowedWhenAbstract("Abstract model can't have an index")
+            elif is_abstract is False or is_abstract is None:
+                # Enforce that non abstract tables have table_name
+                if not has_table_name and not parent_is_attribute_container:
+                    raise InheritanceError("Non abstract models need a table name")
```
jpinner-lyft: This also seems like a breaking change -- I believe we allow dynamic reconfiguration of Models today, and as part of that allow Model's to overwrite or redefine their meta settings up until the point that we need to fetch the table metadata.

That is we can create a Model class without `table_name` and set it after class creation dynamically as long as we do so before `_get_connection` is ever called.

pynamodb/models.py
```diff
@@ -227,6 +293,9 @@ def __init__(self, hash_key=None, range_key=None, **attributes):
         :param range_key: Only required if the table has a range key attribute.
         :param attrs: A dictionary of attributes to set on this object.
         """
+        # Abstract models aren't allowed to have instance
+        self.__class__._check_not_abstract()
+
```
jpinner-lyft: It seems like this check should happen in the metaclass' `__new__` method

pynamodb/models.py
```diff
@@ -227,6 +293,9 @@ def __init__(self, hash_key=None, range_key=None, **attributes):
         :param range_key: Only required if the table has a range key attribute.
         :param attrs: A dictionary of attributes to set on this object.
         """
+        # Abstract models aren't allowed to have instance
+        self.__class__._check_not_abstract()
```
jpinner-lyft: It seems like this check should happen in the metaclass' __new__ method

pynamodb/models.py
```diff
@@ -313,9 +388,13 @@ def batch_write(cls, auto_commit=True):
                             passed here, changes automatically commit on context exit
                             (whether successful or not).
         """
+        cls._check_not_abstract()
         return BatchWrite(cls, auto_commit=auto_commit)
 
     def __repr__(self):
+        # Sanity check, as abstract models aren't allowed to have instance
```
jpinner-lyft: These sanity checks don't seem necessary.

pynamodb/models.py
```diff
@@ -822,13 +922,15 @@ def delete_table(cls):
...
         """
+        cls._check_not_abstract()
         return cls._get_connection().delete_table()
 
     @classmethod
     def describe_table(cls):
         """
         Returns the result of a DescribeTable operation on this model's table
         """
+        cls._check_not_abstract()
```
jpinner-lyft: Why is the check for `table_name` in `_get_connection()` not sufficient here?

jpinner-lyft:
I had a thought with what might be an alternate approach (though you all might have already considered and rejected this):

Given that we already have `AttributeContainer` as a base class for inheriting MapAttributes, and that a Model already extends it, do we perhaps already have a form of abstract inheritance already and we simply need to document / codify the rules around it?

```
SharedAttributes(AttributeContainer):
    foo = StringAttribute()

MyModelWithBar(Model, SharedAttributes):
    # inherits foo from SharedAttributes
    bar = NumberAttribute()

MyModelWithBaz(Model, SharedAttribtues):
    # inherits foo from SharedAttributes
    baz = NumberAttribtue((
```

pynamodb/models.py
Jamim: Hi @jpinner-lyft,
If you choose this approach please don't forget that model inheriting abstract model will also be abstract.
I believe you should handle it somehow.

bendog:
How is this PR going?

mojimi:
@garrettheel Can you review it please?

marioscience:
@jpinner-lyft @garrettheel any chance we can inherit from pynamodb.models.Model in a BaseModel that we can inherit from in all our custom models? That way it would be possible to write functionality in tandem with Model functionality. For example, right now if I want to write some code for the save function before doing super(...).save() I have to do it for every model and that is not very DRY. Any chance that this is supported in any way? 

The abstract class implementation in the PR above sounds like a good solution for my use case. I already tried just doing it with pynamodb out of the box but receive an error about the constructor not getting the right arguments. Any ideas or suggestions?

PR #442, opened 2018-02-02
prevent future datas

jmphilli:


pynamodb/attributes.py
```diff
@@ -768,7 +768,7 @@ def deserialize(self, values):
             if attr_class is None:
                 continue
             deserialized_value = None
-            if attr_value is not None:
+            if attr_value is not None and attr_class is not None:
```
scode: Isn't the `if attr_class is None` above already causing us never to get here if `attr_class` is None?

Should we have a test for this case?

scode:
Not sure I understand the intent here. The phrasing makes it sound like you want pynamo to ignore data seen in the database that we have no definition for. Outside of the comment I dropped (because I think it's already happening), is this really a behavior we want? Isn't it better to fail with an error than to silently "corrupt" data?

Keeping in mind that the data might be getting read with the purpose of then writing it back (e.g. the user calls `save()` and loses the data).

Apologies in advance if I'm missing something.

jpinner-lyft:
Echoing what @scode wrote, I don't think this change does anything. Pynamo already ignores unknown fields that are in the database.

The open question is if some or all `save()`s of an item should also ignore unknown fields. The current behavior is for an item save to remove any unknown fields.

PR #449, opened 2018-02-26
Improve error message where table_name is None

asottile:


pynamodb/models.py
```diff
@@ -1261,10 +1261,21 @@ def _get_connection(cls):
...
             raise AttributeError(
-                """As of v1.0 PynamoDB Models require a `Meta` class.
-                See https://pynamodb.readthedocs.io/en/latest/release_notes.html"""
+                'As of v1.0 PynamoDB Models require a `Meta` class.\n'
+                'Model: {}.{}\n'
+                'See https://pynamodb.readthedocs.io/en/latest/release_notes.html\n'.format(
+                    cls.__module__, cls.__name__,
+                ),
+            )
+        elif cls.Meta.table_name is None:
```
betamoo: ```elif not hasattr(cls.Meta, "table_name") or cls.Meta.table_name is None:```

betamoo: ```elif not hasattr(cls.Meta, "table_name") or cls.Meta.table_name is None:```

in case `Meta` doesn't have a `table_name`

asottile: While I agree: this isn't a regression here. Do you still want me to change that?

betamoo: I agree the current behaviour is errorness, but it would be nice to have this corrected as well. Appreciated!

asottile:
seriously, still supporting python2.6? it's been end of lifed for _[OVER 4 YEARS](https://www.python.org/dev/peps/pep-0361/)_?

/me begrudgingly downgrades format literals

pynamodb/models.py
asottile: done!

betamoo:
👍

harleyk:
@asottile We are planning to do a 4.0 release which drops 2.6 and early 3.x versions soon. And then 5.0 will be 3.x only. :)

pynamodb/models.py
betamoo: thanks!

PR #149, opened 2016-08-11
Add boolean attributes

jmphilli:

PR #450, opened 2018-02-27
Use attr.get_value(value) when deserialize

betamoo:
Change update & update_item to use attr.get_value(value) in deserialization.

The PR https://github.com/pynamodb/PynamoDB/pull/149 introduced ```attr.get_value(value)``` for backward compatibility between LBA and BA, and used it in ```model.from_raw_data ``` but not in ```model.update_item``` (nor ```model.update```) causing code like https://gist.github.com/betamoo/1fb15692d6fa74b9ddacfd4965632b26 to fail

### DIFF
--- a/pynamodb/tests/integration/test_migration.py
+++ b/pynamodb/tests/integration/test_migration.py
@@ -156,8 +156,9 @@ def test_migrate_boolean_attributes_skip_native_booleans(ddb_url):
     assert (0, 0) == migrate_boolean_attributes(BAModel, ['flag'], allow_rate_limited_scan_without_consumed_capacity=True)
 
 
+@pytest.mark.parametrize("flag_value",[True, False, None])
 @pytest.mark.ddblocal
-def test_legacy_boolean_attribute_deserialization_compatibility(ddb_url):
+def test_legacy_boolean_attribute_deserialization_in_update(ddb_url, flag_value):
     class BAModel(Model):
         class Meta:
             table_name = 'lba_deserialization_test'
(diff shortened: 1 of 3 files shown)


jeremy-lyft:
can you add some tests?

pynamodb/tests/integration/test_migration.py
```diff
@@ -154,3 +154,105 @@ class Meta:
...
+    # Create objects with a BooleanAttribute flag
+    BAModel('pkey1', flag=True, value = 'value0').save()
+    BAModel('pkey2', flag=False, value = 'value0').save()
+    BAModel('pkey3', flag=None, value = 'value0').save()
+
+    # Check we are able to read the flag with LegacyBooleanAttribute
+    assert True == LBAModel.get('pkey1').flag
+    assert False == LBAModel.get('pkey2').flag
+    assert None == LBAModel.get('pkey3').flag
+
```
betamoo: Without the change, update_item will throw `TypeError: expected string or buffer`

pynamodb/tests/integration/test_migration.py
```diff
@@ -154,3 +154,105 @@ class Meta:
...
+    # Update a value in the model causing LegacyBooleanAttribute to be deserialized
+    LBAModel.get('pkey1').update_item('value', 'value1', 'PUT')
+    LBAModel.get('pkey2').update_item('value', 'value1', 'PUT')
+    LBAModel.get('pkey3').update_item('value', 'value1', 'PUT')
+
+    # Check we are able to read the flag with LegacyBooleanAttribute
+    assert True == LBAModel.get('pkey1').flag
+    assert False == LBAModel.get('pkey2').flag
+    assert None == LBAModel.get('pkey3').flag
+
```
betamoo: Without the change, update will throw `TypeError: expected string or buffer`

pynamodb/tests/integration/test_migration.py
```diff
@@ -154,3 +154,105 @@ class Meta:
     BAModel.create_table(read_capacity_units=1, write_capacity_units=1)
     BAModel('pkey', flag=True).save()
     assert (0, 0) == migrate_boolean_attributes(BAModel, ['flag'], allow_rate_limited_scan_without_consumed_capacity=True)
+
+
+@pytest.mark.ddblocal
+def test_legacy_boolean_attribute_deserialization_compatibility(ddb_url):
```
jeremy-lyft: you can probably pytest.mark.parametrize those two tests

jeremy-lyft:
+1

pynamodb/tests/integration/test_migration.py
```diff
@@ -154,3 +154,105 @@ class Meta:
...
+    BAModel.get('pkey2').update_item('value', 'value1', 'PUT')
+    BAModel.get('pkey3').update_item('value', 'value1', 'PUT')
+
+    # Check we are able to read the flag with BooleanAttribute
+    assert True == BAModel.get('pkey1').flag
+    assert False == BAModel.get('pkey2').flag
+    assert None == BAModel.get('pkey3').flag
+
+    # Update a value in the model causing BooleanAttribute to be deserialized
+    BAModel.get('pkey1').update(actions=[BAModel.value.set('value2')])
```
jeremy-lyft: would it make sense to perform this set of updates on top of fresh LBAModel saves, instead of after update_item calls? technically update_item could correct the issue for the item, and the update calls would succeed even without the change.

betamoo:
@jmphilli

pynamodb/tests/integration/test_migration.py
betamoo: that makes sense

PR #452, opened 2018-03-06
Make get_attributes public

zackhsi:
Being able to dynamically introspect Model classes is valuable for tasks
such as validation. This commit converts the underscore-prefixed private
method `_get_attributes` into a public method.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -7,6 +7,7 @@ import json
 from base64 import b64encode, b64decode
 from copy import deepcopy
 from datetime import datetime
+import warnings
 from dateutil.parser import parse
 from dateutil.tz import tzutc
 from inspect import getargspec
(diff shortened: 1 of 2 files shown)


pynamodb/attributes.py
```diff
@@ -218,7 +218,7 @@ def __init__(self, **attributes):
         self._set_attributes(**attributes)
 
     @classmethod
-    def _get_attributes(cls):
+    def get_attributes(cls):
```
jpinner-lyft: while this isn't a public method i'm sure there's a lot of code calling it directly

should we keep the private signature that just calls the public method and mark it as deprecated?

zackhsi: You and your backwards-compatible API design, smh.

Yeah, sounds good 👍 .

PR #455, opened 2018-03-10
Skip null attributes post serialization for maps

### LINKED ISSUE #454: Null attributes post serialization not skipped in map
Related to #375; it was fixed so that null attributes would not be added when serializing a map. However, some attributes (like strings) serialize to null and are still being added post-serialization to the map.

jcomo:
Fixes #454.


jcomo:
@jpinner-lyft @jmphilli

jcomo:
Tests are passing now. Merge when ready

PR #457, opened 2018-03-13
Show that BinaryAttributes cannot be used with `update`.

RyanMcG:
The response data is double base64 encoded.


RyanMcG:
Handled by https://github.com/pynamodb/PynamoDB/commit/1399f912957b4ab5134b365703306375a35a8330

PR #469, opened 2018-03-21
Adding Enum Attribute to support discriminated union

betamoo:


pynamodb/attributes.py
```diff
@@ -386,6 +387,89 @@ def serialize(self, value):
...
+        else:
+            return self
+
+    def get_key_value(self):
+        return (self.key, self.value)
+
+    def get_value(self):
+        return self.value
+
+class EnumAttribute(Attribute):
```
jpinner-lyft: FYI we have lots of code that create `EnumAttribute` to represent python Enum either by name or value

pynamodb/attributes.py
```diff
@@ -386,6 +387,87 @@ def serialize(self, value):
             return six.u(value)
 
 
+class MultiKeyMap(object):
```
dmmiller: Comments.  What does a MultiKeyMap do?

showcase.py
```diff
@@ -0,0 +1,81 @@
...
+m1.save()
+print (MyModel.dumps())
+# Outputs:
+# [["id1", {"attributes": {"enum_attr": {"S": "{\"N\": \"2018\"}"}}}]]
+########################################
+
+# Different assignment patterns
+m2 = MyModel("id2")
+# All have the same effect
+m2.enum_attr["UnicodeAttribute"] = "President"
```
dmmiller: Is this multiple strings way to access things common?  "UnicodeAttribute", "S", "String"

showcase.py
```diff
@@ -0,0 +1,81 @@
...
+
+if MyModel.exists():
+    MyModel.delete_table()
+MyModel.create_table(read_capacity_units=2, write_capacity_units=2)
+
+
+############# Basic Usage ###############
+# Create an object
+m1 = MyModel("id1")
+m1.enum_attr['UnicodeAttribute'] = "Obama"
```
dmmiller: Could we end up with a . notation instead of the indexing notation?

Like m1.enum_attr.UnicodeAttribute = "Obama"

Or even better potentially infer it based on type?

betamoo: I am considering the dot notion, just need to get them to work...

showcase.py
betamoo: the multi-enum-value map theme is new. Usually you access a map like this `map_attr["attr_name"]` or `map_attr.attr_name`. Since I was careful in the design not to add arbitrary names for different values an enum can have, thus we need to access it with a value identifier. Choosing one identifier rather than the other seemed arbitrary to me; that's why I went with these different accessing scheme

showcase.py
dmmiller: :)

pynamodb/attributes.py
betamoo: yep, working on that

showcase.py
dmmiller: I'd argue for being opinionated in API design.  Leaving it open is going to leave you with multiple different ways of doing it and having to support it.  I think you should probably pick just one and make it the way.

betamoo: I need to mention though that "S", "String"  are in Pynamo (I am not introducing these)
They are used internally to map the type of the value, (as well as in boto). I have just exposed these identifiers

PR #480, opened 2018-03-28
Fix deserialization bug in BinarySetAttribute + better error msg when parsing json

betamoo:


pynamodb/attributes.py
```diff
@@ -328,8 +328,11 @@ def deserialize(self, value):
         """
         Returns a decoded string from base64
         """
-        if value and len(value):
-            return set([b64decode(val.encode(DEFAULT_ENCODING)) for val in value])
+        try:
+            if value and len(value):
+                return set([b64decode(val.decode(DEFAULT_ENCODING)) for val in value])
```
betamoo: decode following the scheme of `BinaryAttribute`

Issue #453, opened 2018-03-08
Indexes support a `.get` method


garrettheel:
It would be great if `Index`es supported a `.get`, like `Model`s do, as a convenience method.

The workaround is to do `.query(..., limit=1)`.

jpinner-lyft:
@garrettheel not sure what you want the return to be here -- `get` usually returns the item or raises an exception, but for indexes you are not guaranteed uniqueness so you'd either have to choose the "first" one (as in your example) or return an iterator

garrettheel:
Yep, this was a misunderstanding on my part. From the docs:

> In a DynamoDB table, each key value must be unique. However, the key values in a global secondary index do not need to be unique. To illustrate, suppose that a game named Comet Quest is especially difficult, with many new users trying but failing to get a score above zero. The following is some data that could represent this.

Closing this one, since it doesn't make sense.

PR #484, opened 2018-04-02
Add `get` support for indexes (#453)

GreenSmile:


pynamodb/tests/test_model.py
```diff
@@ -3305,6 +3305,33 @@ def test_index_queries(self):
             }
             self.assertEqual(req.call_args[0][1], params)
 
+        with patch(PATCH_METHOD) as req:
+            # equal to:
+            # for item in IndexedModel.email_index.query('foo', IndexedModel.numbers == [1, 2], limit=1):
+            for item in IndexedModel.email_index.get('foo', range_key=[1, 2]):
+                pass
```
betamoo: Can you please explain what happens here? why the for loop is empty? should we be checking the returned items?

GreenSmile: Nearby tests implement it like:
```
queried = []
for item in IndexedModel.email_index.query('foo', limit=2, user_name__begins_with='bar'):
    queried.append(item._serialize())
```
And then `queried` is never checked, so I decided not to create stub array in this case.

Anyway my implementation of .get() was wrong, as it should return single item, not iterable. Please check updated PR. Thanks.

garrettheel:
Hey @GreenSmile, sorry for the delay on this one. Happy to get this merged. Would you be able to rebase and add some type stubs?

PR #504, opened 2018-05-10
Attribute default must be a callable

ikonst:
Remove support for constant `Attribute.default` values, instead requiring them to be callable. This can reduce the occurrence of a particular class of bugs:

```py
class MyModel(Model):
  time = UTCDateTimeAttribute(default=datetime.utcnow())  # timestamp at class declaration time
```

and:

```py
class MyModel(Model):
  list = ListAttribute(default=[])

m1 = MyModel()
m1.list.append('foo')

m2 = MyModel()
assert m2.list == ['foo']  # :(
```


jpinner-lyft:
this has the potential to break a ton of existing code :(

dgrala:
We came across this issue in Lyft's avingest (with list attributes). 👍 for fixing this - can we ship it as a breaking change?

PR #507, opened 2018-05-30
Fix faulty documentation example

MHannila:
It is important that the default is not mutable, because that would cause the default value to be modified when the first instance of the object is modified.  

The current example results in following behaviour
```python
thread = Thread()
thread.notes.append('First note')
empty_thread = Thread()
empty_thread.notes
>>> ['First note']
```


garrettheel:
Good catch, definitely don't want to be promoting a bad practice like this. Thanks for fixing

PR #522, opened 2018-07-31
Add IntegerAttribute

ikonst:
The IntegerAttribute stores integers in an "N" attribute. To be effective, this change will be paired with a typeshed update.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -525,6 +525,25 @@ class NumberAttribute(Attribute):
         return json.loads(value)
 
 
+class IntegerAttribute(Attribute):
+    """
+    An integer attribute; unlike, NumberAttribute, truncates to an integer
+    """
+    attr_type = NUMBER
+
+    def serialize(self, value):
+        """
+        Encode integers as JSON
+        """
+        return json.dumps(int(value))
+
+    def deserialize(self, value):
+        """
+        Decode numbers from JSON
+        """
+        return int(json.loads(value))
+
+
 class UTCDateTimeAttribute(Attribute):
     """
     An attribute for storing a UTC Datetime
--- a/pynamodb/tests/test_attributes.py
+++ b/pynamodb/tests/test_attributes.py
@@ -17,7 +17,7 @@ from pynamodb.constants import UTC, DATETIME_FORMAT
 from pynamodb.models import Model
 
 from pynamodb.attributes import (
-    BinarySetAttribute, BinaryAttribute, NumberSetAttribute, NumberAttribute,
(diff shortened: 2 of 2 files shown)


pynamodb/attributes.py
```diff
@@ -525,6 +525,25 @@ def deserialize(self, value):
...
+        Encode integers as JSON
+        """
+        return json.dumps(int(value))
+
+    def deserialize(self, value):
+        """
+        Decode numbers from JSON
+        """
+        return int(json.loads(value))
+
```
betamoo: You may need to add a `get_value` method too

Example from Boolean attribute
```
     def get_value(self, value):
         # we need this for legacy compatibility.
         # previously, BOOL was serialized as N
         value_to_deserialize = super(BooleanAttribute, self).get_value(value)
         if value_to_deserialize is None:
             value_to_deserialize = json.loads(value.get(NUMBER_SHORT, '0'))
         return value_to_deserialize
```

`get_value` is called in various places  like `update` and the Integer behaviour should be consistent in all places

ikonst: Are you sure? In `Boolean` it's used as a migration path from `LegacyBoolean` which used a different underlying DynamoDB type. In my case, if someone replaces `NumberAttribute` with `IntegerAttribute`, they'll be using the same underlying type.

betamoo: I doubled checked, it seems all the places that uses `get_value`, pass it through `deserialize`. 
Example: https://github.com/pynamodb/PynamoDB/blob/57f5be33f9690bcf0c94b06efd6a7abd112edd03/pynamodb/models.py#L392
So I think you're right, it is not needed

betamoo:
👍

betamoo:
@harleyk looks good 👍

harleyk:
I'm not sure we want to be adding more non standard types to the core Pynamo interface which is my biggest concern. We are interested in the next iteration on this actually removing everything that isn't a Dynamo standard type from this and pushing stuff like this up to consumers.

ikonst:
@harleyk I see the value in being transparent about the underlying machinery, and we might want to create a pynamodb-attributes package containing things like `UTCDateTimeAttribute` (and maybe `UTCTimestampAttribute` that's an `N`?). At that point, we could also move `IntegerAttribute`, `LegacyBooleanAttribute`, etc. in one fell swoop.

Meanwhile, this gets rid of lots of `int(...)`s, `cast(...)`s and `# type: ignore`s that I see littering pretty much every codebase here in Lyft.

ikonst:
This will be introduced in a separate library instead.

PR #523, opened 2018-08-11
Update attributes.py

dgoldman916:
I noticed this issue when `_serialize` is called- while appending a dictionary with a list inside of it to a ListAttribute. Somewhere along the way the dictionary values are being returned as tuple, and is raising a ValueError- `ValueError: Unknown value: <class 'tuple'>` inside of `_get_class_for_serialize`, since it is not in `SERIALIZE_CLASS_MAP`. Folding tuple behavior right into list behavior seems like a catch-all, but it would be nice if there was a way to handle this tedious exception.


ikonst:
Can you provide a code sample reproducing the problem?

Issue #529, opened 2018-08-26
PynamoDB broken with botocore >= 1.11.0


ButtaKnife:
This change breaks Connection._make_api_call(): https://github.com/boto/botocore/commit/818d23505620f5091e5a4114020b66b5e6d7aa7b#diff-054181c44857a41f4e27bab9c8a6ba6e

In particular, https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/connection/base.py#L357 is trying to access a `proxies` property which is no longer included on the botocore Endpoint class.

anovis:
got same issue. See the stack trace below. 

```
File "/usr/local/lib/python3.6/site-packages/pynamodb/models.py", line 485, in get
    hash_key, range_key = cls._serialize_keys(hash_key, range_key)
  File "/usr/local/lib/python3.6/site-packages/pynamodb/models.py", line 1372, in _serialize_keys
    hash_key = cls._hash_key_attribute().serialize(hash_key)
  File "/usr/local/lib/python3.6/site-packages/pynamodb/models.py", line 1219, in _hash_key_attribute
    hash_keyname = cls._get_meta_data().hash_keyname
  File "/usr/local/lib/python3.6/site-packages/pynamodb/models.py", line 1262, in _get_meta_data
    cls._meta_table = MetaTable(cls._get_connection().describe_table())
  File "/usr/local/lib/python3.6/site-packages/pynamodb/connection/table.py", line 263, in describe_table
    return self.connection.describe_table(self.table_name)
  File "/usr/local/lib/python3.6/site-packages/pynamodb/connection/base.py", line 654, in describe_table
    tbl = self.get_meta_table(table_name, refresh=True)
  File "/usr/local/lib/python3.6/site-packages/pynamodb/connection/base.py", line 499, in get_meta_table
    data = self.dispatch(DESCRIBE_TABLE, operation_kwargs)
  File "/usr/local/lib/python3.6/site-packages/pynamodb/connection/base.py", line 313, in dispatch
    data = self._make_api_call(operation_name, operation_kwargs)
  File "/usr/local/lib/python3.6/site-packages/pynamodb/connection/base.py", line 357, in _make_api_call
    proxies=self.client._endpoint.proxies,
AttributeError: 'Endpoint' object has no attribute 'proxies'
```

toshke:
gettings same error with botocore version `1.11.1`
```
/pynamodb/connection/base.py", line 357, in _make_api_call
[...]

PR #531, opened 2018-08-27
Fix botocore > 1.11.0 compatibility

Moutix:
Related to #529 

I don't think it's a good idea to rely on protected parameters here (but was already the case before)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,13 @@ python:
   - "pypy"
 
 env:
-  - AWS_SECRET_ACCESS_KEY=fake_key AWS_ACCESS_KEY_ID=fake_id UPDATE_BOTOCORE=0
-  - AWS_SECRET_ACCESS_KEY=fake_key AWS_ACCESS_KEY_ID=fake_id UPDATE_BOTOCORE=1
+  - AWS_SECRET_ACCESS_KEY=fake_key AWS_ACCESS_KEY_ID=fake_id
 
 before_install:
   - pip install six==1.9.0
 
 install:
   - pip install -r requirements-dev.txt
-  - if [ "${UPDATE_BOTOCORE}" = "1" ]; then pip install botocore -U; fi
 
 before_script:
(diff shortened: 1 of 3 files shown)


ningirsu:
Like updating the version of botocore to 1.11.0 in the requirements-dev.txt?

garrettheel:
Yep, happy for tests to track a new-ish botocore

ningirsu:
@garrettheel Should be good now

.travis.yml
```diff
@@ -10,13 +10,15 @@ python:
   - "pypy"
 
 env:
-  - AWS_SECRET_ACCESS_KEY=fake_key AWS_ACCESS_KEY_ID=fake_id
+  - AWS_SECRET_ACCESS_KEY=fake_key AWS_ACCESS_KEY_ID=fake_id UPDATE_BOTOCORE=0
```
garrettheel: Would you mind pinning a newer `botocore` (e.g 1.11.4) in `requirements-dev.txt` instead of the out-of-band update? This will help avoid random test failures

pynamodb/connection/base.py
```diff
@@ -351,10 +351,14 @@ def _make_api_call(self, operation_name, operation_kwargs):
 
             response = None
             try:
+                proxies = getattr(self.client._endpoint, "proxies", None)
```
garrettheel: A comment regarding the change of internal behavior in `1.11.0` might be helpful here

.travis.yml
ningirsu: As you wish, it was to be sure to catch in advance if a new botocore version make another breaking change

pynamodb/connection/base.py
ningirsu: :+1:

.travis.yml
garrettheel: Yep, understand the motivation there. I think in practice we'll know about breaking changes about as quickly anyway and not floating here would avoid disruption to other ongoing development. Long-term I certainly agree that it'd be best to not be so dependent on internal behavior :)

PR #537, opened 2018-08-30
Add type stubs

garrettheel:
This is mostly copied from [typeshed](https://github.com/python/typeshed/tree/master/third_party/2and3/pynamodb).

The bulk of new typing is in `operand.pyi` and adding some new expression args to `Connection` and `TableConnection`. This doesn't cover everything, but it's a start!


pynamodb/expressions/operand.pyi
```diff
@@ -0,0 +1,58 @@
...
+    def serialize(self, placeholder_names: Dict[str, str], expression_attribute_values: Dict[str, str]) -> str: ...
+
+
+class _NumericOperand(_Operand): ...
+
+
+class _ListAppendOperand(_Operand): ...
+
+
+class _ConditionOperand(_Operand): ...
```
ikonst: Is this needed?
```
    def __eq__(self, other: Any) -> condition.Comparison: ...  # type: ignore
    def __ne__(self, other: Any) -> condition.Comparison: ...  # type: ignore
    def __lt__(self, other: Any) -> condition.Comparison: ...
    def __le__(self, other: Any) -> condition.Comparison: ...
    def __gt__(self, other: Any) -> condition.Comparison: ...
    def __ge__(self, other: Any) -> condition.Comparison: ...
    def between(self, lower: Any, upper: Any) -> condition.Between: ...
    def is_in(self, *values: Any) -> condition.In: ...
```

garrettheel: Added these and some more

Issue #539, opened 2018-09-25
DeprecationWarning: inspect.getargspec() is deprecated


junejie:
Environment:
```
platform linux -- Python 3.6.5, pytest-3.8.1, py-1.6.0, pluggy-0.7.1
rootdir: /home/x/projects/x, inifile: setup.cfg
plugins: flake8-1.0.0, env-0.6.2, cov-2.5.1
```

Error:
```
/home/x/.virtualenvs/x/lib/python3.6/site-packages/pynamodb/attributes.py:632: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect
.getfullargspec()
  attribute_args = getargspec(Attribute.__init__).args[1:]

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

This is a warning from `pytest`

tom-dalton-fanduel:
For anyone else waiting for the above PR to get merged/released, you can work around this by adding a warning filter to pytest's config like so:

```
[pytest]
filterwarnings =
	# PynamoDB issue: https://github.com/pynamodb/PynamoDB/issues/539
	# Should be fixed by https://github.com/pynamodb/PynamoDB/pull/541
	ignore:inspect.getargspec\(\) is deprecated:DeprecationWarning:pynamodb.attributes
```

Ref https://docs.pytest.org/en/latest/warnings.html and https://docs.python.org/3/library/warnings.html#describing-warning-filters

TomGoBravo:
Hey @ikonst, this is the source of the one warning when I run tests in my repo. Please may you merge PR #541 (or something like it) to master?

ikonst:
Not fixed in #701?

TomGoBravo:
@ikonst Yes, I tried running my tests against bf582948901649f3bcc6ec94ec367be83b7db504 and the warnings went away :-) Please make a new pypi release and mark this issue as fixed.

ikonst:
Released 4.1.0 with the fix.

PR #541, opened 2018-09-29
fix for DeprecationWarning: inspect.getargspec() is deprecated

junejie:
ref: https://github.com/pynamodb/PynamoDB/issues/539


ikonst:
Same as #701?

garrettheel:
Fixed in #701

PR #543, opened 2018-10-05
Remove invocation of botocore's vendor code

gbshankar:
Seems like the use of vendor code is causing exceptions like these:
OpenSSL/SSL.py", line 1993, in shutdown self._raise_ssl_error(self._ssl, result)
OpenSSL/SSL.py", line 1631, in _raise_ssl_error
raise SysCallError(errno, errorcode.get(errno))
SysCallError: (32, 'EPIPE')

botocore no longer uses this vendor code. So, probably a good time to remove from pynamodb too.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -6,6 +6,7 @@ from __future__ import division
 import logging
 import math
 import random
+import requests
 import time
 import uuid
 import warnings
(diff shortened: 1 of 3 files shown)


pynamodb/connection/base.py
```diff
@@ -6,6 +6,7 @@
 import logging
 import math
 import random
+import requests
```
asottile: you'll want to add `requests` to `setup.py`'s `install_requires`

gbshankar: Ya, that's a blunder. Fixed it.

garrettheel:
Requests are made through `botocore`, instead of `requests`, as of 4.0

PR #546, opened 2018-10-16
Don't break on serialization of optional and null map attributes

azatoth:


ikonst:
What scenario are you fixing here?

PR #559, opened 2018-11-06
Replaces vendored botocore Requests with "normal" Requests.

### LINKED ISSUE #558: Vendored Botocore Requests
Hello All!

Botocore has begun un-vendoring the Requests library, and replacing it with urllib3: https://github.com/boto/botocore/pull/1495.

PynamoDB is making use of the vendored botocore Requests library here: https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/connection/base.py#L19-L20

As such, this results in breakage when unit testing with the latest version of moto (https://github.com/spulec/moto), which addresses this change.

Is there a major level of effort required to swap out the vendored requests? I would be happy to submit a PR, but I don't have a ton of knowledge on the lower-level logic of PynamoDB. I would imagine that this would will need to be done, since botocore is migrating away from making use of the vendored Requests library.

mikegrima:
See #558 for details.

Fixes #558.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -47,7 +47,7 @@ The default AWS region to connect to.
 session_cls
 -----------
 
-Default: ``botocore.vendored.requests.Session``
+Default: ``requests.Session``
 
 A class which implements the Session_ interface from requests, used for making API requests
 to DynamoDB.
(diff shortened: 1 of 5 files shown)


asottile:
This is a dupe of #543

garrettheel:
We use `botocore` to make requests as of `4.0`

PR #571, opened 2018-12-16
Pass null datetimes through deserializer

AdamNiederer:
This prevents a crash when a UTCDateTimeAttribute is nullable and is explicitly
set to None through Model.update

### DIFF
--- a/pynamodb/tests/test_attributes.py
+++ b/pynamodb/tests/test_attributes.py
@@ -151,6 +151,9 @@ class TestUTCDateTimeAttribute:
         attr = UTCDateTimeAttribute()
         assert attr.deserialize(tstamp.strftime(DATETIME_FORMAT)) == tstamp
 
+    def test_utc_date_time_desrialize_none(self):
+        assert UTCDateTimeAttribute().deserialize(None) == None
+
     def test_dateutil_parser_fallback(self):
         """
         UTCDateTimeAttribute.deserialize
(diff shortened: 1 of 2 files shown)


aventurella:
I encountered a similar issue with Nullable JSONAttributes , Thanks for the unit test I had initially forgotten to do that in my Pull Request https://github.com/pynamodb/PynamoDB/pull/581

pynamodb/tests/test_attributes.py
```diff
@@ -151,6 +151,9 @@ def test_utc_date_time_deserialize(self):
         attr = UTCDateTimeAttribute()
         assert attr.deserialize(tstamp.strftime(DATETIME_FORMAT)) == tstamp
 
+    def test_utc_date_time_desrialize_none(self):
```
ikonst:
```suggestion
    def test_utc_date_time_deserialize_none(self):
```

ikonst:
This is not specific to UTCDateTimeAttribute, so I don't addressing it for a specific attribute type is correct. Furthermore, this is not a problem when you `save()` so this pretty strongly indicates that `deserialize` is not expected to handle `None`s.

I guess you're doing something like:
```py
my_model.update({
    'my_nullable_date': {'value': None, 'action': 'PUT'},
})
```
while the correct way would be:
```
my_model.update({
    'my_nullable_date': {'action': 'DELETE'},
})
```

PR #576, opened 2018-12-30
Add rudimentary tests for type stubs

ikonst:
This suite tests some of the more tricky type stubs.

I'm intentionally using the same convention (`# E: expected error`) as mypy's data-driven test suites. Unfortunately mypy's own testsuite engine [cannot be leveraged](https://github.com/python/mypy/issues/6115) at this point.


ikonst:
cc @garrettheel can take a 👀?

ikonst:
Switching to a (surprisingly simpler!) implementation where we "reconstruct" the test program with the "# E: ..." comments, so that the `pytest -vv` diff is exactly what needs to be changed:
```
E           class MyModel(Model):
E         -     my_list = ListAttribute(of=MyMap)
E         +     my_list = ListAttribute(of=MyMap)  # E: bar
E         ?                                      ++++++++++
```

ikonst:
> It's unfortunate that there's not a more direct way to use the mypy API to reduce some of the setup/string types here

In fact, mypy's own data-driven tests are using the exact same pattern. I think mypy's implementation (which keeps the "programs" in a separate .ini file) allows the 'expected' portion (the "# E: ..." comments) to be automatically updated if mypy decides to change the output format.



> I'll keep an eye out on the pytest plugin to help simplify this for future.

I agree. Improving [pytest-mypy](https://github.com/dbader/pytest-mypy) might be one avenue. I've started [mypy issue 6115](https://github.com/python/mypy/issues/6115) to discuss this.

garrettheel:
Thanks!

PR #580, opened 2019-01-04
Stop using requests in favour of botocore & urllib3

garrettheel:
Using botocore's vendored `requests` library is becoming increasingly problematic. This moves to using `urllib3` instead via `botocore`, rather than managing the `requests` version directly and further diverging from the `botocore` implementation.

Breaking changes include:
 * Dropping python 2.6 support
 * No more `session_cls` or `requests_timeout_seconds` config
   * Added `connect_timeout_seconds`, `read_timeout_seconds` and `max_pool_connections` config
 * Exceptions raised should remain the same, however they will no longer wrap `requests` exceptions.
   * For example where the `exc.cause` was previously a `requests.exceptions.ReadTimeout`, it would now be a `botocore.exceptions.ReadTimeoutError`.

Improvements:
 * Implements the `before-send` botocore hook to support testing via moto

Related: #558, #566, #559, #543, #569

Also related: https://github.com/boto/botocore/issues/1486

## Benchmarks

Here are some simple benchmarks using a dynalite server. Seems to be ~10% faster for the read/write path.

```
# (kill-requests)$ python -m timeit -s 'import benchmark; benchmark.setup()' -n 2000 -r 10 -- "benchmark.test()"
2000 loops, best of 10: 1.63 msec per loop
# (master)$ python -m timeit -s 'import benchmark; benchmark.setup()' -n 2000 -r 10 -- "benchmark.test()"
2000 loops, best of 10: 1.82 msec per loop

from pynamodb.attributes import NumberAttribute, UnicodeAttribute


class UserModel(Model):
    class Meta:
        table_name = 'UserModel'

    user_name = UnicodeAttribute(hash_key=True, attr_name='user_name')
    user_id = UnicodeAttribute(range_key=True)
    zip_code = NumberAttribute(null=True)
    email = UnicodeAttribute()


def setup():
    UserModel.create_table(read_capacity_units=1, write_capacity_units=1, wait=True)
    UserModel("test", user_id="1234", email="test@gmail.com", zip_code=1234).save()


def test():
    UserModel.get("test", "1234")
```


pynamodb/connection/base.py
```diff
@@ -338,12 +344,24 @@ def _make_api_call(self, operation_name, operation_kwargs):
         1. It's faster to avoid using botocore's response parsing
         2. It provides a place to monkey patch requests for unit testing
         """
+
+        """
+        Garrett TODO:
+         * why was botocore's response parsing slow? is that still the case?
```
danielhochman: botocore uses an internal construct called "shapes" to validate responses meet the expected format. shapes are made up of other shapes. shapes are defined in JSON model files, which can generically be shared and updated among across their various SDKs. because of this, botocore does a bunch of recursive expansion and parsing to ultimately validate the response. we don't need any of the validation though.

botocore is designed for compatibility and maintainability rather than efficiency or speed.

pynamodb/connection/base.py
```diff
@@ -338,12 +344,24 @@ def _make_api_call(self, operation_name, operation_kwargs):
...
+
         operation_model = self.client._service_model.operation_model(operation_name)
         request_dict = self.client._convert_to_request_dict(
             operation_kwargs,
             operation_model
         )
-        prepared_request = self._create_prepared_request(request_dict, operation_model)
+
+        # TODO: add an option for switching between urllib3/requests? will need
+        # to document changes. especially things like exceptions
```
danielhochman: these are internal methods, so i'm not sure it warrants a compatibility release before deprecation. i would just cut over.

if you want to do a compatibility release that just changes exceptions to standardize on internal exceptions first you could do that, then release the cutover to urllib3.

docs/settings.rst
```diff
@@ -9,13 +9,20 @@ Settings reference
...
+
 
-request_timeout_seconds
+read_timeout_seconds
 -----------------------
 
 Default: ``60``
 
-The default timeout for HTTP requests in seconds.
+The time till a timeout exception is thrown when attempting to read from a connection.
```
ikonst:
```suggestion
The time in seconds until a URLError is raised when attempting to read from a connection.
```
(assuming it's `URLError`)

garrettheel: I think these: https://github.com/boto/botocore/blob/cd7fd229c4dd3637bceac440a0b31a142774754a/botocore/exceptions.py#L109

pynamodb/connection/base.py
```diff
@@ -489,7 +525,22 @@ def client(self):
         # if the client does not have credentials, we create a new client
         # otherwise the client is permanently poisoned in the case of metadata service flakiness when using IAM roles
         if not self._client or (self._client._request_signer and not self._client._request_signer._credentials):
-            self._client = self.session.create_client(SERVICE_NAME, self.region, endpoint_url=self.host)
+            # TODO(garrett): should we set the timeout directly on the
+            # httpsession instead? this represents a change to existing
+            # behavior, where we set `timeout` on the requests session which
+            # would have been an _overall_ timeout of 60 seconds. this would
+            # change behavior to be 60 seconds each for connect and read. there
+            # is no "total" option available in botocore config
```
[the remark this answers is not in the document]
garrettheel: I currently defaulted to 60, which is the same as botocore. Do you think we should diverge and set it to something more reasonable?

pynamodb/connection/base.py
```diff
@@ -489,7 +525,22 @@ def client(self):
...
+
+            # note that `max_pool_connections` defaults to 10 and can be
+            # configured using the config object below.
+            # TODO: check what the requests default is and possibly expose this
+            # as a top-level option via the API. note that this would invalidate
```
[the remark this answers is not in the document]
garrettheel: Done. Need to consider how we'll support adding headers to requests with the removal of `session_cls`

docs/settings.rst
ikonst: Either one?

docs/release_notes.rst
```diff
@@ -1,6 +1,23 @@
...
+v4.0.0a1
+--------
+
+:date: 2019-04-10
+
+This is major release and contains breaking changes. Please read the notes below carefully.
+
+Given that `botocore` has moved to using `urllib3` directly for making HTTP requests, we'll be doing the same (via `botocore`). This means the following:
+
+ * The `session_cls` option is no longer supported. <TODO: discuss replacement opts>
```
ikonst: TODO

docs/release_notes.rst
```diff
@@ -1,6 +1,23 @@
...
+ * The `request_timeout_seconds` parameter is no longer supported. `connect_timeout_seconds` and `read_timeout_seconds` are now available instead.
+
+Other changes in this release:
+
+ * Fixes moto support through implementing the botocore "before-send" hook. Other botocore hooks remain unimplemented.
```
ikonst: Will address https://github.com/pynamodb/PynamoDB/issues/569 ?

docs/settings.rst
```diff
@@ -9,13 +9,20 @@ Settings reference
 
 Here is a complete list of settings which control default PynamoDB behavior.
 
+connect_timeout_seconds
+-----------------------
+
+Default: ``60``
+
+The time till a timeout exception is thrown when attempting to make a connection.
```
ikonst:
```suggestion
The time in seconds until a <class> is raised when attempting to make a connection.
```

docs/settings.rst
```diff
@@ -44,15 +51,13 @@ Default: ``"us-east-1"``
...
------------
+max_pool_connections
+--------------------
 
-Default: ``botocore.vendored.requests.Session``
+Default: ``10``
 
-A class which implements the Session_ interface from requests, used for making API requests
-to DynamoDB.
+TODO
```
ikonst: TODO

docs/release_notes.rst
garrettheel: Correct, should fix this

pynamodb/connection/base.py
garrettheel: I'm just going to cut over. I don't think the effort to maintain both is worth it

pynamodb/connection/base.py
```diff
@@ -222,27 +227,28 @@ class Connection(object):
...
         else:
-            self.session_cls = get_settings_value('session_cls')
+            self._connect_timeout_seconds = get_settings_value('connect_timeout_seconds')
 
-        if request_timeout_seconds is not None:
-            self._request_timeout_seconds = request_timeout_seconds
+        if read_timeout_seconds is not None:
+            self._read_timeout_seconds = read_timeout_seconds
         else:
-            self._request_timeout_seconds = get_settings_value('request_timeout_seconds')
```
ikonst: Will there be any warning about this setting no longer being applied?

pynamodb/connection/base.py
```diff
@@ -338,35 +338,55 @@ def _make_api_call(self, operation_name, operation_kwargs):
         1. It's faster to avoid using botocore's response parsing
         2. It provides a place to monkey patch requests for unit testing
         """
+
```
ikonst: boop

pynamodb/connection/base.py
```diff
@@ -338,35 +338,55 @@ def _make_api_call(self, operation_name, operation_kwargs):
...
+                    # to reset it before attempting to send the request again.
+                    # This will ensure that we resend the entire contents of the
+                    # body.
+                    prepared_request.reset_stream()
+
+                    # Create a new request when retried (including a new signature).
+                    prepared_request = self._create_prepared_request(request_dict, operation_model)
+
+                # Implement the before-send event from botocore
+                event_name = 'before-send.dynamodb.%s' % operation_model.name
```
ikonst: I think percent-formatting is deprecated now?

pynamodb/connection/base.py
```diff
@@ -338,35 +338,55 @@ def _make_api_call(self, operation_name, operation_kwargs):
...
-            except (requests.RequestException, ValueError) as e:
+                if i > 0:
+                    # If there is a stream associated with the request, we need
+                    # to reset it before attempting to send the request again.
+                    # This will ensure that we resend the entire contents of the
+                    # body.
+                    prepared_request.reset_stream()
+
+                    # Create a new request when retried (including a new signature).
+                    prepared_request = self._create_prepared_request(request_dict, operation_model)
```
ikonst: Can we de-indent (and then remove line 351) ?

pynamodb/connection/base.pyi
```diff
@@ -22,13 +25,21 @@ class Connection:
...
+        connect_timeout_seconds: Optional[float] = ...,
+        read_timeout_seconds: Optional[float] = ...,
+        max_retry_attempts: Optional[int] = ...,
+        base_backoff_ms: Optional[int] = ...,
+        max_pool_connections: Optional[int] = ...,
+        extra_headers: Optional[MutableMapping[Text, Text]] = ...,
+    ) -> None: ...
     def dispatch(self, operation_name, operation_kwargs): ...
     @property
     def session(self): ...
```
ikonst: curious, this is not typed? hard to understand what "session" means

pynamodb/connection/base.py
ikonst: Read on, 💯

pynamodb/settings.py
```diff
@@ -8,11 +8,13 @@
...
 default_settings_dict = {
-    'request_timeout_seconds': 60,
+    'connect_timeout_seconds': 60,
+    'read_timeout_seconds': 60,
     'max_retry_attempts': 3,
     'base_backoff_ms': 25,
     'region': 'us-east-1',
-    'session_cls': requests.Session,
+    'max_pool_connections': 10,
+	'extra_headers': None,
```
ikonst: stray tab

pynamodb/connection/base.py
garrettheel: Yep, took this from botocore which supports 2.6 still. I'll switch to `.format` now that we've dropped support for that.

pynamodb/connection/base.py
garrettheel: Good call - updated.

pynamodb/connection/base.pyi
garrettheel: Added a type for it

pynamodb/connection/base.py
garrettheel: removed (:

PR #581, opened 2019-01-09
Fix deserializing a None value for a JSONAttribute

aventurella:
If another user has written a JSONAttribute from a separate language/lib: `{NULL: true}` to a table PynamoDB also queries then it will fail to deserialize the JSON value is `None` was passed in.

This adds a conditional check for None and acts accordingly for JSONAttribute

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -436,6 +436,9 @@ class JSONAttribute(Attribute):
         """
         Deserializes JSON
         """
+        if value is None:
+            return None
+
         return json.loads(value, strict=False)
 
 


ikonst:
I'm not sure what's the exact scenario. Are you defining a `JSONAttribute(null=False)` and expect it to handle a null value?

JSONAttribute is based by an S-typed attribute, and DynamoDB won't store a null in it -- the only way to make it a "null" is to delete the attribute. Given the following model:
```py
class MyModel(Model):
    class Meta:
        table_name = f'pynamodb-{__name__}'
        read_capacity_units = 10
        write_capacity_units = 10

    key = UnicodeAttribute(hash_key=True)
    value = JSONAttribute(null=True)
```
this code:
```py
MyModel._get_connection().put_item(uuid_key, attributes={
    'value': {'S': None},
})
```
produces an 'invalid_type' DynamoDB error.

Can you take this test stub and change it to reproduce the bug?
```py
import pytest
from pynamodb.attributes import UnicodeAttribute, JSONAttribute


class MyModel(Model):
    class Meta:
        table_name = f'pynamodb-{__name__}'
        read_capacity_units = 10
        write_capacity_units = 10

    key = UnicodeAttribute(hash_key=True)
    value = JSONAttribute(null=True)


@pytest.fixture(scope='module', autouse=True)
def create_table():
    MyModel.create_table()


def test_bug(uuid_key):
    MyModel._get_connection().put_item(uuid_key, attributes={
        'value': {'S': None},
    })

    # verify underlying storage
    item = MyModel._get_connection().get_item(uuid_key)
    assert 'value' not in item['Item']

    # verify deserialization
    model = MyModel.get(uuid_key)
    assert model.value is None
```

Thanks!

ikonst:
You're right, we seem to be mishandling `NULL` attributes. For that matter, `NullAttribute` also doesn't seem to be resulting in `{"NULL": true}` puts.

Here's a repro:
```py
import json
import subprocess
from uuid import uuid4
from pynamodb.attributes import UnicodeAttribute, JSONAttribute, NullAttribute
from pynamodb.exceptions import TableError

endpoint_url = 'http://localhost:8000'
table_name = 'pynamodb-test-nullable-model'

class MyModel(Model):
    class Meta:
        host = endpoint_url
        table_name = table_name
        read_capacity_units = 10
        write_capacity_units = 10

    key = UnicodeAttribute(hash_key=True)
    value = JSONAttribute()


def test():
    try:
        MyModel.delete_table()
    except TableError:
        pass
    MyModel.create_table()

    key = str(uuid4())
    item_data = {"key": {"S": key}, "value": {"NULL": True}}
    subprocess.call([
        'aws', 'dynamodb',
        'put-item',
        '--endpoint-url', endpoint_url,
        '--table-name', table_name,
        '--item', json.dumps(item_data),
    ])

    # verify deserialization
    model = MyModel.get(key)
    assert model.value is None

test()
```

PR #583, opened 2019-01-10
Add tests for typing of untyped lists

ikonst:


ikonst:
cc @betamoo mostly to prove that I'm not making things worse :)

PR #587, opened 2019-01-16
Fix type stubs for Model.count

garrettheel:


pynamodb/models.py
```diff
@@ -546,6 +546,7 @@ def count(cls,
         :param filter_condition: Condition used to restrict the query results
         :param consistent_read: If True, a consistent read is performed
         :param index_name: If set, then this index is used
+        :param rate_limit:
```
ikonst: Let's add docs while at it? Or are you just trying to make PyCharm happy? :)

pynamodb/tests/test_mypy.py
```diff
@@ -9,6 +9,18 @@
 from .mypy_helpers import assert_mypy_output  # noqa
 
 
+def test_model():
```
ikonst: `test_model_count`?

garrettheel: I figured I'd add more model-related tests in here later as necessary

PR #594, opened 2019-02-12
Add a sleep delay after a commit to avoid ProvisionedThroughputExceededException

augustincouette:


pynamodb/models.py
```diff
@@ -45,9 +45,10 @@ class ModelContextManager(object):
 
     """
 
-    def __init__(self, model, auto_commit=True):
+    def __init__(self, model, auto_commit=True, delay=0):
```
ikonst: You add it to the base class but it should only apply to a specific subclass of `BatchWrite`?

ikonst:
@garrettheel is there no rate-limiting mechanism for batch-writing? Does only batch-reading benefit from rate limiting?

garrettheel:
@ikonst not currently. I think I'd prefer that we use the same interfaces as much as possible if implementing write rate limiting

PR #598, opened 2019-02-27
Allow set max_operations in meta class for batch writes

DKbyo:


ikonst:
Could you explain the rationale? Per [DynamoDB docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html):

> A single BatchWriteItem operation can contain up to 25 PutItem or DeleteItem requests. The total size of all the items written cannot exceed 16 MB.

Do you want to set it to less than 25? Also, wouldn't it useful to be configurable as a parameter to the `batch_write` method rather than a single value per model?

DKbyo:
Well, if you have capacity units limited for small tables/projects, you need to limit batch writes and put some delay between them. And yes... will be more useful to set it as a parameter in the `batch_write`  method if you change the capacity units dinamically.

ikonst:
`BatchWrite` doesn't perform any throttling though (i.e. there's no `time.sleep` or anything like that). Let's say you reduce the number from 25 to 15, you'd simply be making HTTP calls more often, but your capacity utilization would still be the same. From the docs:

> Parallel processing reduces latency, but each specified put and delete request consumes the same number of write capacity units whether it is processed in parallel or not.

Would this work for your case?
```py
from more_itertools import chunked

batch_writer = MyModel.batch_write()
for models_chunk in chunked(models, 15):
   for model in models_chunk:
      batch_writer.save(model)
   batch_writer.commit()
   time.sleep(1)
```

PR #604, opened 2019-03-26
Add ability to disable SSL certificate verification

CaptObvious:
When working with a local DynamoDB clone it's not possible to disable SSL verification right now.


ikonst:
Interesting, since we actually do this regularly:
```py
class Meta:
    host = "http://localhost:8080"
```

How do you end up using the `Connection` class directly?

garrettheel:
I don't think this particular PR is relevant since we moved off requests in 4.0, but would be interested to hear the use-case if you want something similar for botocore. Closing for now, but please reopen if necessary!

Issue #89, opened 2015-11-05
Transaction support


sidtaduri:
Heya @jlafon, 

Any plans to support transactions in future releases?
https://aws.amazon.com/blogs/aws/dynamodb-transaction-library/

jlafon:
Hi @sidtaduri, I had read through the code in that library when it was released to see if PynamoDB could provide a similar feature. It's not a trivial thing to implement, and I don't have quite enough time to dedicate to it (I maintain PynamoDB in my spare time). However, pull requests are always welcome :)

jmphilli:
@sidtaduri do you still feel you need this functionality?

sidtaduri:
@jmphilli - not at the moment.

wobeng:
@jmphilli I do, do you have it?

jhesed:
Hi @jmphilli,
Is there any plan on implementing transactions in the near future?

vinodmehra:
https://aws.amazon.com/blogs/aws/new-amazon-dynamodb-transactions/?sc_channel=sm&sc_publisher=TWITTER&sc_country=Global&sc_geo=GLOBAL&sc_outcome=awareness&trk=AWS_reInvent_2018_launch_DynamoDB_Transactions_TWITTER&sc_content=AWS_reInvent_2018_launch_DynamoDB_Transactions&sc_category=Amazon+DynamoDB&linkId=60250973

jhesed:
wow that's a gem @vinodmehra.  Thanks!

batmop:
Would love this implemented

disrael:
This is a must have. Please assign and give time frame or give us pointers on how we would pull request it in.

Surgo:
I think it have following tasks

- [ ] Share connection between tables (maybe backward imcompatible) 
    - [`TransactGetItems`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactGetItems.html) and [`TransactWriteItems`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html) can updating multiple tables at once
   - Like [django's global db connection](https://github.com/django/django/blob/2.1.5/django/db/__init__.py#L16)?
- [ ] Add `TransactGetItems` `TransactWriteItems` 
[...]

PR #615, opened 2019-04-28
adding support for transactions

hallie:
#89


pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,56 @@
...
+
+transact_write = TransactWrite(host=cfg.DYNAMODB_HOST, region='us-east-1')
+User(1).save(in_transaction=transact_write)
+BankStatement(1).save(condition=(BankStatement.user_id.does_not_exist()), in_transaction=transact_write)
+transact_write.commit()
+
+transact_get = TransactGet(host=cfg.DYNAMODB_HOST, region='us-east-1')
+user = User.get(1, in_transaction=transact_get)
+statement = BankStatement.get(1, in_transaction=transact_get)
+transact_get.commit()
```
hparanjape-lyft: This test is asserting that the transactions do not fail. But, should it also assert that the values are correct? Also, would like to see negative test cases in integration test as well.

pynamodb/connection/base.py
```diff
@@ -419,6 +418,7 @@ def _make_api_call(self, operation_name, operation_kwargs):
                     raise VerboseClientError(botocore_expected_format, operation_name, verbose_properties)
                 except VerboseClientError as e:
                     if is_last_attempt_for_exceptions:
+                        print(botocore_expected_format)
```
jmphilli: i'll assume the prints go away when the tests come

pynamodb/connection/base.py
```diff
@@ -948,26 +952,44 @@ def update_item(self,
...
+            operation_kwargs[CONDITION_EXPRESSION] = condition_expression
+        if name_placeholders:
+            operation_kwargs[EXPRESSION_ATTRIBUTE_NAMES] = self._reverse_dict(name_placeholders)
+        if return_values:
+            operation_kwargs.update(self.get_return_values_map(return_values))
+        operation_kwargs[TABLE_NAME] = table_name
+        operation_kwargs.update(self.get_identifier_map(table_name, hash_key, range_key))
+        return operation_kwargs
+
+    def get_operation_kwargs_for_put_item(self,
```
Aidanie: same here

pynamodb/connection/base.py
```diff
@@ -1080,15 +1126,12 @@ def batch_get_item(self,
...
-    def get_item(self,
-                 table_name,
-                 hash_key,
-                 range_key=None,
-                 consistent_read=False,
-                 attributes_to_get=None):
-        """
-        Performs the GetItem operation and returns the result
-        """
+    def get_operation_kwargs_for_get_item(self,
```
Aidanie: ditto

pynamodb/connection/transaction.py
```diff
@@ -0,0 +1,180 @@
...
+from pynamodb.constants import (
+    TABLE_NAME, PROJECTION_EXPRESSION, UPDATE, GET, PUT, DELETE, TRANSACT_ITEMS, CLIENT_REQUEST_TOKEN, CONDITION_CHECK,
+    CONDITION_EXPRESSION, EXPRESSION_ATTRIBUTE_NAMES, EXPRESSION_ATTRIBUTE_VALUES, UPDATE_EXPRESSION,
+    RETURN_VALUES_ON_CONDITION_FAILURE, ITEM, KEY, RETURN_VALUES, RESPONSES
+)
+
+PUT = PUT.lower().capitalize()
+DELETE = DELETE.lower().capitalize()
+
+TRANSACT_ITEM_LIMIT = 10
```
Aidanie: nit: maybe link the necessity of the magic number (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-dynamodb-transactions)

pynamodb/models.py
```diff
@@ -344,9 +357,14 @@ def delete(self, condition=None, conditional_operator=None, **expected_values):
             kwargs.update(expected=self._build_expected_values(expected_values, DELETE_FILTER_OPERATOR_MAP))
         kwargs.update(conditional_operator=conditional_operator)
         kwargs.update(condition=condition)
+
+        if in_transaction is not None:
+            operation_kwargs = self._get_connection().get_operation_kwargs_for_delete_item(*args, **kwargs)
+            in_transaction.add_delete_item(operation_kwargs)
+            return True
```
Aidanie: The return statements here (and for other operation results) seem kinda odd - we're really entering a promise state, rather than a straightforward result return. I'm not sure returning a default is a good eventual form

pynamodb/models.py
```diff
@@ -442,15 +466,20 @@ def update(self, attributes=None, actions=None, condition=None, conditional_oper
...
+            data = self._get_connection().update_item(*args, **kwargs)
+            for name, value in data[ATTRIBUTES].items():
+                attr_name = self._dynamo_to_python_attr(name)
+                attr = self.get_attributes().get(attr_name)
+                if attr:
+                    setattr(self, attr_name, attr.deserialize(attr.get_value(value)))
+            return data
+
+        operation_kwargs = self._get_connection().get_operation_kwargs_for_update_item(*args, **kwargs)
+        in_transaction.add_update_item(operation_kwargs)
```
Aidanie: no return here, unlike the others? 🤔

pynamodb/models.py
```diff
@@ -481,25 +515,40 @@ def get(cls,
...
             range_key=range_key,
             consistent_read=consistent_read,
             attributes_to_get=attributes_to_get
         )
-        if data:
-            item_data = data.get(ITEM)
-            if item_data:
-                return cls.from_raw_data(item_data)
-        raise cls.DoesNotExist()
+        in_transaction.add_get_item(cls, hash_key, range_key, operation_kwargs)
```
Aidanie: ditto on the no return

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,257 @@
...
+
+def get_error_code(error):
+    return error.cause.response['Error'].get('Code')
+
+
+@pytest.mark.usefixtures('setup')
+@pytest.mark.ddblocal
+class TestTransaction:
+
+    def test_transact_write__error__idempotent_parameter_mismatch(self):
```
Aidanie: DynamoDB doesn't seem to require a clientRequestToken for transactions - we should probably maintain pairity with the interface as offered - https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html#API_TransactWriteItems_RequestParameters

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,257 @@
...
+    request.addfinalizer(finalizer=fin)
+
+
+def get_error_code(error):
+    return error.cause.response['Error'].get('Code')
+
+
+@pytest.mark.usefixtures('setup')
+@pytest.mark.ddblocal
+class TestTransaction:
```
Aidanie: There's a lot of model setup going on per test, perhaps moving that to a common setup/breakdown would be helpful.

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,257 @@
...
+        user2 = transaction.from_results(User, 2)
+        statement1 = transaction.from_results(BankStatement, 1)
+        statement2 = transaction.from_results(BankStatement, 2)
+
+        assert user1.user_id == 1
+        assert statement1.user_id == 1
+        assert statement1.balance == 0
+        assert user2.user_id == 2
+        assert statement2.user_id == 2
+        assert statement2.balance == 100
```
Aidanie: nit: could have common func that ensures user/bank/statement models are equivalent

pynamodb/tests/integration/test_transaction_integration.py
hallie: it doesn't require you to pass your own in, but it does require one in general (it'll generate it if you don't) :0

So if you try to double commit with the same transaction connection, it'll throw this error

pynamodb/models.py
hallie: I'm still trying to figure out the best way for handling mypy issues that come with the fact that transactions don't return anything

pynamodb/tests/integration/test_transaction_integration.py
hallie: the original idea was to use a `setup` fixture with `scope="class"` to create before/delete after. it works locally, but is failing in CI, so I'm gonna see if moving it to `setup_class`/`teardown_class` fixes it

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,263 @@
...
+    user_id = NumberAttribute(hash_key=True)
+    balance = NumberAttribute(default=0)
+    active = BooleanAttribute(default=True)
+
+
+class LineItem(Model):
+
+    class Meta:
+        region = 'us-east-1'
+        table_name = 'line-item'
```
betamoo: I think you need the host here

hallie: the host is set here: https://github.com/pynamodb/PynamoDB/pull/615/files#diff-6bc097399e79abc4d8f5f7d0bc88da30R85

pynamodb/tests/integration/test_transaction_integration.py
betamoo: I am pretty sure we need to pass this token if we want item idempotency. I couldn't find anything in the doc you linked about it generating a token?

pynamodb/models.py
ikonst: fwiw, overloaded return types might actually be possible with mypy literals: https://mypy.readthedocs.io/en/latest/literal_types.html
e.g.
```py
def delete(..., in_transaction=Literal[None], ...) -> ...:
```

It is somewhat odd but 🤷‍♂ I guess that's why Python is not Java?

pynamodb/tests/integration/test_transaction_integration.py
hallie: I mean, I don't generate/add an idempotency key to the transaction by default, but the idempotency error still raises if i attempt to commit the same request twice 🤷‍♀

pynamodb/models.py
hallie: ok, i will attempt to overload :0

ikonst: Sorry, of course I meant:
```py
@overload
def delete(..., in_transaction: None, ...) -> ...:
@overload
def delete(..., in_transaction: Transaction, ...) -> ...:
```
(`:` and not `=`)

PR #618, opened 2019-05-07
Add support for transactions

### LINKED ISSUE #89: Transaction support

hallie:
Creating an interface for working with [Transactions](https://aws.amazon.com/blogs/aws/new-amazon-dynamodb-transactions/) in DynamoDB

Includes support for
- [TransactGetItems](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactGetItems.html)
  - Get
- [TransactWriteItems](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html#API_TransactWriteItems_RequestSyntax)
  - ConditionCheck
  - Delete
  - Put
  - Update

Closes #89


pynamodb/models.py
```diff
@@ -460,6 +490,11 @@ def save(self, condition=None, conditional_operator=None, **expected_values):
             kwargs.update(expected=self._build_expected_values(expected_values, PUT_FILTER_OPERATOR_MAP))
         kwargs.update(conditional_operator=conditional_operator)
         kwargs.update(condition=condition)
+
+        if in_transaction is not None:
```
emise: An issue we're running into with this kind of implementation-

What will happen if we try to overload the `save` method with additional validation logic?

1st save: Validate on non-existence of a field, then we add the field for saving. Because it's a transaction, we don't actually commit.
2nd save via transaction: Validate on non-existence of a field, but this field now exists. Validation fails.

emise: ^ can ignore this now; as @hallie pointed out, it's actually caused by a manual retry. Can use the retries on the transaction to implement retry logic.

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,218 @@
...
+    @pytest.mark.ddblocal
+    def test_transact_get(self):
+        # making sure these entries exist, and with the expected info
+        User(1).save()
+        BankStatement(1).save()
+        User(2).save()
+        BankStatement(2, balance=100).save()
+
+        # get users and statements we just created and assign them to variables
+        with TransactGet() as transaction:
```
garrettheel: In general I like this interface, though I'm worried that either:
 1) People will accidentally try to use properties of these models before the transaction is committed
 2) A single item will hit some error condition [1] and then the proxy object will be invalid

Did you consider returning some _other_ object which the hydrated model can be fetched? That other object would raise exceptions if things are not right (the transaction is still open, the item doesn't exist, etc.)

cc @ikonst for opinions


[1]  https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactGetItems.html
> If a requested item could not be retrieved, the corresponding ItemResponse object is Null, or if the requested item has no projected attributes, the corresponding ItemResponse object is an empty Map.

pynamodb/models.py
```diff
@@ -345,8 +372,20 @@ def delete(self, condition=None, conditional_operator=None, **expected_values):
...
+            operation_kwargs = self._get_connection().get_operation_kwargs_for_delete_item(*args, **kwargs)
+            in_transaction.add_delete_item(model=self, operation_kwargs=operation_kwargs)
+            return {}
         return self._get_connection().delete_item(*args, **kwargs)
 
+    def update_item_with_raw_data(self, data):
```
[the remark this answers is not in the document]
hallie: Can write a test, but the reason it isn't private is because its used in the `transactions` module to update the model with the fetched data

pynamodb/constants.py
```diff
@@ -245,6 +252,53 @@
     BETWEEN: 'between'
 }
 
+# each transaction can only support 10 actions at a time
+# https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-dynamodb-transactions
+TRANSACT_ITEMS_LIMIT = 10
```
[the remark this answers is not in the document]
hallie: Sure :0 Should I also remove the client request token validation (that its a string/less than 37 characters)? I originally added that because my team got pretty confused about the source of a write failure that ultimately ended up being that the request token they were using was too long

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,218 @@
...
+                m.delete_table()
+
+    @staticmethod
+    def get_error_code(error):
+        return error.cause.response['Error'].get('Code')
+
+    @pytest.mark.ddblocal
+    def test_transact_write__error__idempotent_parameter_mismatch(self):
+
+        transaction = TransactWrite(host=DDB_URL)
```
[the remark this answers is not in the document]
hallie: On a high level, this is mostly testing that reusing the same client request token causes the second transaction to fail. I can rewrite the test to make this clearer

pynamodb/tests/integration/test_transaction_integration.py
hallie: Do you mean like making a class that inherits from `Model` that will raise an error if some attribute (`transaction_in_progress = True`?) indicates that it hasn't completed and the user attempts to access any properties before the commit? Then updates the model with all of the attributes at the end?

I guess my general worry would be how do we make it clear to the user that the models returned by a transaction aren't the same model classes that they'd have in a normal `get`

ikonst: Something like:
```py
with TransactGet() as transaction:
    user1_promise = User.get(1, in_transaction=transaction)

user1 = user1_promise.resolve()
```
?

I suppose this could be typed as:
```
_T = TypeVar('_T', bound='Model')

class _ModelPromise[_T]:
    def resolve() -> _T: ...

class Model(metaclass=MetaModel):
    ...
    @override
    @classmethod
    def get(cls: Type[_T], hash_key: KeyType, range_key: Optional[KeyType] = ..., consistent_read: bool = ..., in_transaction: Transaction) -> _ModelPromise[_T]: ...

    @override
    @classmethod
    def get(cls: Type[_T], hash_key: KeyType, range_key: Optional[KeyType] = ..., consistent_read: bool = ...) -> _T: ...
```

pynamodb/constants.py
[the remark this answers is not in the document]
garrettheel: If the message that the API gives is unclear then happy to have validation here. This is just something that could easily change in a future DynamoDB release, so better if we don't hardcode validation logic.

pynamodb/tests/integration/test_transaction_integration.py
hallie: I guess the primary downside to this is that you can't do `user1: User = User.get(...)`, and if you wanted `User` typing on the variable, you'd have to reassign it outside of the transaction like:
```
with TransactWrite() as t:
    _user1_promise = User.get(1, in_transaction=t)
    _user2_promise = User.get(2, in_transaction=t)
user1: User = _user1_promise.resolve()
user2: User = _user2_promise.resolve()
```

ikonst: To clarify, you don't need the `: User` part - mypy should be smart enough.

pynamodb/constants.py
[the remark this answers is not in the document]
hallie: That's a fair point :0 I don't think its confusing enough to make it a blocker if dynamo changes in the future

pynamodb/tests/integration/test_transaction_integration.py
ikonst: @jpinner-lyft says he'd rather
```py
with TransactWrite() as t:
    user1_future = t.get(User, 1)
    ...
user1 = user1_future.get()
```

(a) let's call it a "future" rather than a "promise" as per:
> A future is a placeholder object for a result that does not yet exist. A promise is a writable, single-assignment container, which completes a future. Promises can complete the future with a result to indicate success, or with an exception to indicate failure.
> ...
> Promises can be thought of as a future with a public set method which the caller (or anybody else) can use to set the value of the future.

Since in our case the caller (i.e. user code) cannot set the value of the future, then the caller is merely given a "future".

(b) how about we rename "resolve" to "get" as per
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#get()
?

garrettheel: Yep, I'm on board with the future idea. I think we should move back to having the `.get` and `.delete`, etc. as methods on the `Transaction` object rather than overloading the ones on `Model`. 

e.g
```python
with TransactGet() as t:
  user_future = t.get(User, hash_key, range_key)

try:
  user = user_future.get()
except User.DoesNotExist:
  pass
except ...
```

ikonst: Let's `resolve()` —> `get()`, or maybe `value` (property). I've made the mistake of suggesting `resolve()` but it's actually a method of a _promise_ which should get the value as an argument (i.e. `promise.resolve(to_value)`).

garrettheel: @ikonst Yep, you're right there. I haven't worked with promise semantics for a while now so forgot some of the specifics :)

hallie: So, to be clear on the ask, we want `TransactGet` to have a completely different interface from `TransactWrite`, in that, instead of overloading the `Model.get` function, we create a `TransactGet.get`, that works like:
```python
with TransactGet() as t:
    user_future: _ModelFuture = t.get(model=User, hash_key=1)
    ride_future: _ModelFuture = t.get(model=Ride, hash_key=1, range_key=2)

try:
    user = user_future.get()
    ride = ride_future.get()
except DoesNotExist:
    pass
```

?

pynamodb/models.pyi
```diff
@@ -127,3 +131,11 @@ class BatchWrite(Generic[_T], ModelContextManager[_T]):
     def __exit__(self, exc_type, exc_val, exc_tb) -> None: ...
     pending_operations: Any
     def commit(self) -> None: ...
+
+class _ModelFuture:
+    _model_cls: Type[_T]
+    _model: Optional[_T]
```
hallie: so i keep getting mypy errors on `_model_cls: Type[_T]` and `_model: Optional[_T]`: `error: Invalid type "pynamodb.models._T"`. Would y'all happen to know why? @garrettheel @ikonst

pynamodb/tests/integration/test_transaction_integration.py
garrettheel: > instead of overloading the Model.get function, we create a TransactGet.get, that works like ...
Yep, that's the suggestion.

> we want TransactGet to have a completely different interface from TransactWrite
Nope! I would suggest that we adopt a similar interface with `TransactWrite` for consistency. Happy to talk through what that should look like

pynamodb/models.pyi
ikonst: you need to make `_ModelFixture` a generic class:
```
class _ModelFuture(Generic[_T]):
```

hallie: thank you! that did it :)

pynamodb/connection/base.pyi
```diff
@@ -58,6 +58,7 @@ class Connection:
     def get_item_collection_map(self, return_item_collection_metrics): ...
     def get_exclusive_start_key_map(self, table_name: Text, exclusive_start_key): ...
 
+    def get_operation_kwargs_for_delete_item(self, *args, **kwargs) -> Dict: ...
```
ikonst: Are those `get_operation_kwargs_for_...` becoming part of the API?

pynamodb/connection/transactions.py
```diff
@@ -0,0 +1,124 @@
...
+from pynamodb.constants import (
+    RETURN_VALUES_ON_CONDITION_FAILURE, ITEM, RETURN_VALUES, RESPONSES,
+    TRANSACTION_CONDITION_CHECK_REQUEST_PARAMETERS, TRANSACTION_DELETE_REQUEST_PARAMETERS,
+    TRANSACTION_GET_REQUEST_PARAMETERS, TRANSACTION_PUT_REQUEST_PARAMETERS, TRANSACTION_UPDATE_REQUEST_PARAMETERS
+)
+from pynamodb.models import _ModelFuture
+
+
+class Transaction(object):
+
```
ikonst: Reminder - docstring

pynamodb/connection/transactions.py
```diff
@@ -0,0 +1,124 @@
...
+        self._return_consumed_capacity = return_consumed_capacity
+
+    def _commit(self):
+        raise NotImplementedError()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._commit()
```
ikonst: We commit even when there was an exception?

pynamodb/connection/transactions.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self, *args, **kwargs):
+        super(TransactGet, self).__init__(*args, **kwargs)
+        self._get_items = []
+        self._proxy_models = []
+
+    def get(self, model_cls, hash_key, range_key=None):
+        self._hash_model(model_cls, hash_key, range_key)
+        operation_kwargs = model_cls.get_operation_kwargs_for_get_item(hash_key, range_key=range_key)
+        get_item = self._format_request_parameters(TRANSACTION_GET_REQUEST_PARAMETERS, operation_kwargs)
+        proxy_model = _ModelFuture(model_cls)
```
ikonst:
```suggestion
        model_future = _ModelFuture(model_cls)
```
?

pynamodb/connection/transactions.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class TransactGet(Transaction):
+    _get_items = None
+    _proxy_models = None
+
+    def __init__(self, *args, **kwargs):
+        super(TransactGet, self).__init__(*args, **kwargs)
+        self._get_items = []
+        self._proxy_models = []
```
ikonst:
```suggestion
        self._futures = []
```

pynamodb/connection/transactions.pyi
```diff
@@ -0,0 +1,46 @@
+from typing import Set, Tuple, TypeVar, Type, Any, List, Optional, Dict, Union, Text
+
+from pynamodb.models import Model, _ModelFuture
+
+from pynamodb.connection import Connection
+
+
+_T = TypeVar('_T', bound=Model)
```
ikonst: Maybe `ModelType` if we have `KeyType`?

pynamodb/models.py
```diff
@@ -250,6 +251,16 @@ def __init__(self, hash_key=None, range_key=None, **attributes):
             attributes[self._range_keyname] = range_key
         super(Model, self).__init__(**attributes)
 
+    def get_hash_key(self):
+        key_name = self._hash_keyname
+        return getattr(self, key_name)
+
+    def get_range_key(self):
+        key_name = self._range_keyname
+        if key_name is not None and hasattr(self, key_name):
```
```suggestion
        if key_name is not None:
            getattr(self, key_name, None)
```

iirc, `hasattr(self, key_name)` is just a `getattr(self, key_name, None)` that discards the result: https://hynek.me/articles/hasattr/

pynamodb/models.py
```diff
@@ -250,6 +251,16 @@ def __init__(self, hash_key=None, range_key=None, **attributes):
             attributes[self._range_keyname] = range_key
         super(Model, self).__init__(**attributes)
 
+    def get_hash_key(self):
```
ikonst: @jpinner-lyft is planning to introduce
```py
    @property
    def _hash_key(self):
        return getattr(self, self._hash_keyname)

    @property
    def _range_key(self):
        return getattr(self, self._range_keyname) if self._range_keyname else None
```
in this #644.

Maybe you should introduce the same. Note that his version introduces private methods: are you OK with private methods?

pynamodb/models.py
```diff
@@ -324,15 +335,42 @@ def __repr__(self):
...
+            condition=condition,
+        )
+        in_transaction.add_condition_check_item(
+            model_cls=cls,
+            hash_key=hash_key,
+            range_key=range_key,
+            operation_kwargs=operation_kwargs
+        )
+
+    def delete(self, condition=None, in_transaction=None):
```
ikonst: `get`s and `put`s are methods of the transaction, but `delete`s still use the `in_transaction`-parameter API?

pynamodb/models.py
```diff
@@ -324,15 +335,42 @@ def __repr__(self):
...
         kwargs.update(condition=condition)
+
+        if in_transaction is not None:
+            operation_kwargs = self._get_connection().get_operation_kwargs_for_delete_item(*args, **kwargs)
+            in_transaction.add_delete_item(model=self, operation_kwargs=operation_kwargs)
+            return {}
         return self._get_connection().delete_item(*args, **kwargs)
 
-    def update(self, actions, condition=None):
+    def _update_item_with_raw_data(self, data):
```
ikonst: @jpinner-lyft does this mesh with your work in #644?

pynamodb/models.py
```diff
@@ -324,15 +335,42 @@ def __repr__(self):
...
+        for name, value in data.items():
+            attr_name = self._dynamo_to_python_attr(name)
+            attr = self.get_attributes().get(attr_name)
+            if attr:
+                setattr(self, attr_name, attr.deserialize(attr.get_value(value)))
+
+    def update(self, actions, condition=None, in_transaction=None):
```
ikonst: ditto re: `in_transaction`

pynamodb/models.py
```diff
@@ -975,3 +1030,23 @@ def _serialize_keys(cls, hash_key, range_key=None):
         if range_key is not None:
             range_key = cls._range_key_attribute().serialize(range_key)
         return hash_key, range_key
+
+
+class _ModelFuture:
+
```
ikonst: Is this space for a docstring?

pynamodb/models.py
```diff
@@ -975,3 +1030,23 @@ def _serialize_keys(cls, hash_key, range_key=None):
...
+        self._resolved = False
+
+    def update_with_raw_data(self, data):
+        if data:
+            self._model = self._model_cls.from_raw_data(data=data)
+        self._resolved = True
+
+    def get(self):
+        if not self._resolved:
+            raise Exception('Cannot resolve until request has been processed')
```
```suggestion
            raise RuntimeException('Cannot get until resolved')
```
unless we want the language to be specifically about transactions, which might be OK-ish

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,247 @@
...
+from pynamodb.models import Model
+
+IDEMPOTENT_PARAMETER_MISMATCH = 'IdempotentParameterMismatchException'
+PROVISIONED_THROUGHPUT_EXCEEDED = 'ProvisionedThroughputExceededException'
+RESOURCE_NOT_FOUND = 'ResourceNotFoundException'
+TRANSACTION_CANCELLED = 'TransactionCanceledException'
+TRANSACTION_IN_PROGRESS = 'TransactionInProgressException'
+
+
+DDB_URL = os.getenv('PYNAMODB_INTEGRATION_TEST_DDB_URL', 'http://localhost:8000')
```
ikonst: Why not use the `ddb_url` fixture?

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,247 @@
...
+    DifferentRegion,
+    LineItem,
+    User,
+]
+
+
+CONNECTION = Connection(host=DDB_URL)
+
+
+class TestTransactionIntegration:
```
ikonst: Can we have the tests as functions rather than class methods? This is how modern pytest tests should be written. For setup and teardown that's per-module, do:
```
@pytest.fixture(scope='module', autouse=True)
def create_tables(ddb_url):
    for m in TEST_MODELS:
        m.Meta.host = ddb_url
        m.create_table(
            read_capacity_units=10,
            write_capacity_units=10,
            wait=True
        )
    yield
    for m in TEST_MODELS:
        if m.exists():
            m.delete_table()
```

btw, 'create_table' already checks for 'exists'

pynamodb/connection/base.pyi
garrettheel: Can you add all of the `get_operation_kwargs..` functions here?

pynamodb/connection/base.pyi
```diff
@@ -93,8 +105,25 @@ class Connection:
         return_item_collection_metrics: Optional[Any] = ...
     ) -> Dict: ...
 
+    def transact_write_items(
+        self,
+        condition_check_items: List[Dict],
+        delete_items: List[Dict],
+        put_items: List[Dict],
+        update_items: List[Dict],
+        client_request_token: Optional[str] = ...,
```
garrettheel:
```suggestion
        client_request_token: Optional[Text] = ...,
```

pynamodb/connection/transactions.py
```diff
@@ -0,0 +1,124 @@
...
+    def _format_request_parameters(valid_parameters, operation_kwargs):
+        if RETURN_VALUES in operation_kwargs.keys():
+            operation_kwargs[RETURN_VALUES_ON_CONDITION_FAILURE] = operation_kwargs.pop(RETURN_VALUES)
+        return {
+            key: value for key, value in operation_kwargs.items() if key in valid_parameters
+        }
+
+
+class TransactGet(Transaction):
+    _get_items = None
```
garrettheel: What's the purpose of these?

pynamodb/connection/transactions.py
```diff
@@ -0,0 +1,124 @@
...
+    RETURN_VALUES_ON_CONDITION_FAILURE, ITEM, RETURN_VALUES, RESPONSES,
+    TRANSACTION_CONDITION_CHECK_REQUEST_PARAMETERS, TRANSACTION_DELETE_REQUEST_PARAMETERS,
+    TRANSACTION_GET_REQUEST_PARAMETERS, TRANSACTION_PUT_REQUEST_PARAMETERS, TRANSACTION_UPDATE_REQUEST_PARAMETERS
+)
+from pynamodb.models import _ModelFuture
+
+
+class Transaction(object):
+
+    _connection = None
```
garrettheel: Same here -- why the class variables?

pynamodb/models.py
hallie: im fine with updating to be less specific

pynamodb/connection/transactions.py
hallie: do you mean why are they put outside of the `__init__`, or why do they exist at all?

pynamodb/connection/base.pyi
hallie: ye

pynamodb/models.py
```diff
@@ -257,7 +257,8 @@ def get_hash_key(self):
 
     def get_range_key(self):
         key_name = self._range_keyname
-        if key_name is not None and hasattr(self, key_name):
+        if key_name is not None:
+            return getattr(self, key_name, None)
             return getattr(self, key_name)
```
ikonst: of course remove this line :) github doesn't let you make suggestions that replace multiple lines

hallie: heh

pynamodb/models.py
hallie: yeah, we just changed it for `get` because we're returning something, but I didn't think we'd agreed to make all of them methods on the transaction? cc @garrettheel

pynamodb/connection/transactions.py
hallie: was mostly copying what happens in batches: https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/models.py#L99

pynamodb/connection/base.pyi
hallie: do you have a different suggestion? i thought it would be easier to have specific functions so that people don't pass parameters that don't work with that operation

pynamodb/tests/integration/test_transaction_integration.py
hallie: sure :0

pynamodb/tests/integration/test_transaction_integration.py
hallie: i think i made this change while debugging some stuff but can put it back as `ddb_url`

pynamodb/models.py
hallie: Currently I'm accessing these within the `Transaction` object, so they wouldn't be used privately :0

pynamodb/models.py
hallie: i can actually undo making this function because im not using it the way i originally intended to

pynamodb/models.py
hallie: What about `in_transaction`?

pynamodb/connection/base.pyi
hallie: actually, now that i've made the function `get_operation_kwargs` I can delete/replace all of these operation-specific ones

pynamodb/connection/transactions.py
hallie: updated to only commit if there are no exception args passed in

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,254 @@
...
+    DifferentRegion,
+    LineItem,
+    User,
+]
+
+
+CONNECTION = None
+
+
+def _get_connection(url):
```
ikonst:
```
@pytest.fixture(scope='module')
def connection(dburl):
    return Connection(host=url)
```

hallie: `ddb_url` is scoped on the function level, so it can't be used with a `module` level fixture, which is why i made the helper :/

hallie: nvm, can just update its fixture

ikonst: Maybe change `ddb_url` to be `session='scope'`? It doesn't really change.

hallie: ye, that's what i did

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,248 @@
...
+        if m.exists():
+            m.delete_table()
+
+
+def get_error_code(error):
+    return error.cause.response['Error'].get('Code')
+
+
+@pytest.mark.ddblocal
+def test_transact_write__error__idempotent_parameter_mismatch(ddb_url, connection):
```
ikonst:
```suggestion
def test_transact_write__error__idempotent_parameter_mismatch(connection):
```
here and there

pynamodb/connection/transactions.py
```diff
@@ -0,0 +1,134 @@
...
+        self._client_request_token = client_request_token
+        self._return_item_collection_metrics = return_item_collection_metrics
+        self._condition_check_items = []
+        self._delete_items = []
+        self._put_items = []
+        self._update_items = []
+
+    def add_condition_check_item(self, model_cls, hash_key, range_key, operation_kwargs):
+        condition_item = self._format_request_parameters(TRANSACTION_CONDITION_CHECK_REQUEST_PARAMETERS, operation_kwargs)
+        self._hash_model(model_cls, hash_key, range_key)
```
garrettheel: Would it not be valid to have multiple condition checks on the same item? I think we only care that it's not one of the items modified in the transaction

hallie: I could be wrong, but don't _think_ so? (Since you can add conditions to all of the write operations on their own, and the whole transaction fails if any of them fail)

But, at the same time, maybe it doesn't matter for us to validate that on this end, and we can just let people try/let the request failure error tell them what to fix

hallie: removing!

pynamodb/models.py
ikonst: I think it was @jpinner-lyft 's strong opinion that methods should preferably have two separate return types based on a parameter

hallie: o, i meant that we're returning something _new_. Like, `Model.get` normally returns an instance of that model, populated with the data returned from the request. But, because we don't have that data until after a transaction completes, we made the decision to return something *completely* different (a `ModelFuture`) to make it clearer to the user that something has to happen before they're able to access the complete model.

The same isn't necessarily true for any of the other operations (except for maybe the `Model.update` because it normally ends with populating the model with the updated data)

ikonst:
```python
if in_transaction is not None:
    ...
    return {}  # empty dict
return self._get_connection().delete_item(*args, **kwargs)  # actual API response
```

so, yeah, the type technically is the same, but the nature is arguably not the same (and arguably that empty dict should've been a `None`)

ikonst: For sake of argument, let's say this function would be typed to return `Optional[dict]` - that'd be suboptimal, since its parameter `is_transaction` controls which one it'd be, so we again come to:
```python
@overload
def delete(..., in_transaction: Transaction) -> None: ...
@overload
def delete(..., in_transaction: None) -> dict: ...
```

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -145,6 +146,20 @@ def test_transact_write__error__transaction_cancelled(connection):
         assert get_error_code(e) == TRANSACTION_CANCELLED
 
 
+@pytest.mark.ddblocal
+def test_transact_write__error__multiple_operations_on_same_record(connection):
+    BankStatement(1).save()
+
+    # attempt to do a transaction with multiple operations on the same record
+    try:
```
ikonst: with pytest.raises(PutError) as exc_info:
    ...

assert get_error_code(exc_info.value) == VALIDATION_EXCEPTION

hallie: tru

pynamodb/models.py
hallie: im fine with conditionally returning `None` :0

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -120,14 +118,13 @@ def test_transact_write__error__idempotent_parameter_mismatch(connection):
 
 @pytest.mark.ddblocal
 def test_transact_write__error__different_regions(connection):
-    try:
+    with pytest.raises(PutError) as exec_info:
```
ikonst: `exc_info` (`exc`eption) — cf. `sys.exc_info()`

pynamodb/models.pyi
```diff
@@ -36,14 +39,23 @@ class Model(metaclass=MetaModel):
...
+    def condition_check(cls, hash_key: KeyType, in_transaction: TransactWrite, condition: Any, range_key: Optional[KeyType] = ...): ...
+    @overload
+    def delete(self, condition: Optional[Any] = ..., in_transaction: None = ...) -> Any: ...
+    @overload
+    def delete(self, condition: Optional[Any] = ..., in_transaction: TransactWrite = ...) -> None: ...
+    @overload
+    def update(self, actions: Optional[List[Any]] = ..., condition: Optional[Any] = ..., in_transaction: None = ...) -> Any: ...
+    @overload
+    def update(self, actions: Optional[List[Any]] = ..., condition: Optional[Any] = ..., in_transaction: TransactWrite = ...) -> None: ...
+    @overload
```
hallie: hello @ikonst im getting this mypy error in the python 3.7 test run and its not very clear to me why its failing (it looks like all the other ones, which are passing), and i was wondering if you knew what was wrong:
```
pynamodb/models.pyi:53: error: Overloaded function signatures 1 and 2 overlap with incompatible return types
```

ikonst: hmpf, we have a pretty’s recent mypy version... I wonder what that that could be.

ikonst: Well, `Any` and `None` are compatible :) But I wish this would work.

ikonst: Maybe try and define the 2nd one without the defaulting?
```python
def save(self, condition: Optional[Any] = ..., in_transaction: TransactWrite) -> None: ...
```

hallie: that seemed to fix it locally :P

pynamodb/models.pyi
```diff
@@ -52,7 +52,7 @@ class Model(metaclass=MetaModel):
     @overload
     def save(self, condition: Optional[Any] = ..., in_transaction: None = ...) -> Dict[str, Any]: ...
     @overload
-    def save(self, condition: Optional[Any] = ..., in_transaction: TransactWrite = ...) -> None: ...
+    def save(self, in_transaction: TransactWrite, condition: Optional[Any] = ...) -> None: ...
```
ikonst: Uh, the reordering makes it wrong, though.
Maybe
```python
@overload
def save(self, condition: Optional[Any], in_transaction: TransactWrite) -> None: ...
@overload
def save(self, *, condition: Optional[Any] = ..., in_transaction: TransactWrite) -> None: ...
```

Although, maybe we'll just follow the pattern where all those methods are part of the `TransactWrite`?

pynamodb/models.pyi
ikonst: added https://github.com/python/mypy/issues/7047

pynamodb/models.pyi
hallie: If we're going to follow the same pattern of all the methods being on `TransactWrite`, I would rather us go ahead and make that decision, rather than continue to go back and forth on it. :/

I personally prefer the currently implemented method, as I think its a lot cleaner. As I stated well over a month ago at this point, my team's goal was to have this change merged into pynamodb by the end of this month, and I would greatly prefer to not have to completely rewrite this if there's no real reason to do so.

hallie: (also, the `*, in_transaction: TransactWrite` suggestion worked, so its passing mypy again)

ikonst: Hallie, it's erroneous with just 2 overloads, though, since `foo.save(condition, my_transaction)` is not unsupported. I intentionally split it into
1. a signature with a non-defaulted `condition`
2. a signature with no `condition` where you must pass `my_transaction` by keyword

(Thankfully we have just 2 arguments here, otherwise it'd quickly get unwieldy.)

hallie:
@ikonst @garrettheel @jpinner-lyft 
I've updated the `TransactWrite` class with the following methods:
```python
def condition_check(self, model_cls: Type[ModelType], hash_key: KeyType, range_key: Optional[KeyType] = ..., condition: Condition = ...) -> None: ...
    def delete(self, model: ModelType, condition: Optional[Condition] = ...) -> None: ...
    def save(self, model: ModelType, condition: Optional[Condition] = ..., return_values: Optional[Any] = ...) -> None: ...
    def update(self, model: ModelType, actions: List[Any], condition: Optional[Condition] = ..., return_values: Optional[Any] = ...) -> None: ...
```

so now the way to perform an operation is as follows:
```python
with TransactWrite() as transaction:
    transaction.save(User(1))
    transaction.update(User(2), actions=[User.name.set('Hallie')])
    transaction.condition_check(User, 3, condition=(User.name != 'Hallie'))
```

pynamodb/models.pyi
hallie: changed the implementation to be on the `TransactWrite` class

pynamodb/connection/base.py
```diff
@@ -780,38 +800,83 @@ def get_exclusive_start_key_map(self, table_name, exclusive_start_key):
...
+                             return_values_on_condition_failure=None):
         self._check_condition('condition', condition)
 
-        operation_kwargs = {TABLE_NAME: table_name}
-        operation_kwargs.update(self.get_identifier_map(table_name, hash_key, range_key))
+        operation_kwargs = {}
         name_placeholders = {}
         expression_attribute_values = {}
 
+        if table_name is not None and hash_key is not None:
```
garrettheel: The method name made it seem as though this was used to build kwargs for all operations, so I initially thought that this line would be a bug for operations that don't involve a `hash_key` (such as `DeleteTable`). While this does reduce code, I think this method makes things slightly harder to understand.

Assuming the intent is DRY, consider breaking out chunks of this instead (such as `_get_transaction_kwargs`).

pynamodb/connection/base.py
```diff
@@ -873,34 +924,72 @@ def put_item(self,
...
-        if return_values:
-            operation_kwargs.update(self.get_return_values_map(return_values))
-        if name_placeholders:
-            operation_kwargs[EXPRESSION_ATTRIBUTE_NAMES] = self._reverse_dict(name_placeholders)
-        if expression_attribute_values:
-            operation_kwargs[EXPRESSION_ATTRIBUTE_VALUES] = expression_attribute_values
+        try:
+            return self.dispatch(TRANSACT_WRITE_ITEMS, operation_kwargs)
+        except BOTOCORE_EXCEPTIONS as e:
+            raise PutError("Failed to write transaction items: {0}".format(e), e)
```
garrettheel: Did you consider creating a new type of error for these? `PutError` doesn't map well to deletes or updates

pynamodb/models.py
```diff
@@ -975,3 +1022,28 @@ def _serialize_keys(cls, hash_key, range_key=None):
...
+        self._resolved = False
+
+    def update_with_raw_data(self, data):
+        if data:
+            self._model = self._model_cls.from_raw_data(data=data)
+        self._resolved = True
+
+    def get(self):
+        if not self._resolved:
+            raise Exception('Cannot get until resolved')
```
garrettheel: I think it's worth creating a new exception in pynamodb for this. You might want to model it off https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.InvalidStateError, which is pretty similar. See https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.result

pynamodb/tests/integration/test_transaction_integration.py
```diff
@@ -0,0 +1,252 @@
...
+    # create a users and a bank statements for them
+    User(1).save()
+    BankStatement(1).save()
+
+    # attempt to do this as a transaction with the condition that they don't already exist
+    with pytest.raises(PutError) as exc_info:
+        with TransactWrite(connection=connection) as transaction:
+            transaction.save(User(1), condition=(User.user_id.does_not_exist()))
+            transaction.save(BankStatement(1), condition=(BankStatement.user_id.does_not_exist()))
+    assert get_error_code(exc_info.value) == TRANSACTION_CANCELLED
```
garrettheel: Can we check that the sub-code was `ConditionalCheckFailed` here too? I'd also be interested in seeing another test for the `TransactionConflict` case

pynamodb/connection/base.py
hallie: ye, i can call it `TransactError`?

pynamodb/connection/base.py
hallie: yeah, i could have this part be something like `_get_identifiers` and then the rest of it for building kwargs

pynamodb/connection/base.py
garrettheel: Sounds good!

pynamodb/connection/base.py
```diff
@@ -873,34 +924,72 @@ def put_item(self,
...
+
+    def transact_get_items(self, get_items, return_consumed_capacity=None):
+        """
+        Performs the TransactGet operation and returns the result
+        """
+        operation_kwargs = self.get_operation_kwargs(return_consumed_capacity=return_consumed_capacity)
```
betamoo: just curious, in case of multiple tables, is the returned consumed capacity the total sum?

hallie: Unsure, but I can check :0

pynamodb/connection/base.py
```diff
@@ -463,10 +465,15 @@ def _handle_binary_attributes(data):
                 for attr in six.itervalues(item):
                     _convert_binary(attr)
         if RESPONSES in data:
-            for item_list in six.itervalues(data[RESPONSES]):
```
scode: Could you elaborate on the purpose of this change? Probably worth adding comments to the code as well.

Also, is there a chance that in the future the type here won't be concretely `list`? Checking for a concrete list type like this looks superficially dangerous without understanding full reasoning.

pynamodb/models.py
```diff
@@ -975,3 +1022,28 @@ def _serialize_keys(cls, hash_key, range_key=None):
...
+    For example: when performing a TransactGet request, this is a stand-in for a model that will be returned
+    when the operation is complete
+    """
+    def __init__(self, model_cls):
+        self._model_cls = model_cls
+        self._model = None
+        self._resolved = False
+
+    def update_with_raw_data(self, data):
+        if data:
```
scode: Should this be `if data is not None` or is it intentionally trying to capture some falsy case other than `None`? If so, it's not apparent to me what it would be.

pynamodb/connection/base.py
hallie: batch operations return `Responses` as a dict (which is why it iterates over the `itervalues`) whereas transaction operations return them as a list. Is there some less specific type check you'd prefer I do?

pynamodb/models.py
hallie: it can just be `if data is not None`

pynamodb/connection/base.py
scode: No specific preference without investigating boto, but the potential concern is just that relying on a concrete type could potentially cause breakage in the future if all boto is guaranteeing is iterable for example. My comment was more of a general concern around relying on a concrete type, than a claim that it's actually wrong. If it seems safe that this won't break with future boto upgrades, we're good.

*edit*:  Looking at the implementation, I guess this is just the raw JSON response without any boto steps in between, so this shouldn't break unless DynamoDB changes its behavior so it's fine. I commented before realizing this was a parse of a raw response.

pynamodb/connection/base.py
hallie: looks like this:
```
{
    'ConsumedCapacity': [
        {'TableName': 'different-region', 'CapacityUnits': 4.0, 'WriteCapacityUnits': 4.0},
        {'TableName': 'statement', 'CapacityUnits': 4.0, 'WriteCapacityUnits': 4.0},
        {'TableName': 'user', 'CapacityUnits': 4.0, 'WriteCapacityUnits': 4.0}
    ]
}
```

pynamodb/tests/integration/test_transaction_integration.py
hallie: so the error returned by amazon is kinda dumb, and there are [other people](https://github.com/aws/aws-sdk-js/issues/2464) who have worked around it by parsing the string. should we do that to make it more helpful?

an example is, if one condition fails out of 3, the error message is:
```python
{
    'Error': {
        'Message': 'Transaction cancelled, please refer cancellation reasons for specific reasons [None, ConditionalCheckFailed, None]',
        'Code': 'TransactionCanceledException'
    }
}
```

Since the list of errors is kept in request order, we could theoretically return the identifier map with the parsed out error and store it on the response

pynamodb/connection/transactions.py
```diff
@@ -14,13 +15,18 @@ def __init__(self, connection, return_consumed_capacity=None):
...
+        return error.cause.response['Error'].get('Code')
+
     def _commit(self):
         raise NotImplementedError()
 
     def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        print(exc_type, exc_val)
```
ikonst: don't forget to remove

pynamodb/models.py
hallie: im actually not using these anymore so im deleting them from my pr

pynamodb/tests/integration/test_transaction_integration.py
garrettheel: Ah I see, that is unfortunate! I think it's probably OK for the test to just do a `.contains` match to see that it failed for the expected reason, but I don't feel too strongly so whichever way you'd prefer

hallie: I think I'll hold off on adding the reasons to the response, and create a separate pr for that :0 for now, i can just add a check that the message contains the expected reason within the string

pynamodb/connection/base.py
```diff
@@ -899,34 +947,86 @@ def put_item(self,
...
+                             put_items,
+                             update_items,
+                             client_request_token=None,
+                             return_consumed_capacity=None,
+                             return_item_collection_metrics=None):
+        """
+        Performs the TransactWrite operation and returns the result
+        """
+        transact_items = []
+        transact_items.extend([
```
garrettheel: nit: the explicit `[]` wrapping should be unnecessary -- extend can take an iterable. i.e

```python
transact_items.extend(
  {...} for foo in bar
)
```

pynamodb/connection/base.pyi
```diff
@@ -56,9 +56,27 @@ class Connection:
     def get_identifier_map(self, table_name: Text, hash_key, range_key: Optional[Any] = ..., key: Any = ...): ...
     def get_consumed_capacity_map(self, return_consumed_capacity): ...
     def get_return_values_map(self, return_values): ...
+    def get_return_values_on_condition_failure_map(self, return_values_on_condition_failure): ...
```
garrettheel: can you add a type for `return_values_on_condition_failure`?

pynamodb/connection/base.pyi
```diff
@@ -56,9 +56,27 @@ class Connection:
...
+        table_name: Text,
+        hash_key: Any,
+        range_key: Optional[Any] = ...,
+        key: Text = ...,
+        attributes: Optional[Any] = ...,
+        attributes_to_get: Optional[Any] = ...,
+        actions: Optional[Sequence[Action]] = ...,
+        client_request_token: Optional[Text] = ...,
+        condition: Optional[Condition] = ...,
+        consistent_read: bool = ...,
```
garrettheel: should this be `Optional[bool]`?

pynamodb/models.py
```diff
@@ -336,12 +337,13 @@ def delete(self, condition=None):
         kwargs.update(condition=condition)
         return self._get_connection().delete_item(*args, **kwargs)
 
-    def update(self, actions, condition=None):
+    def update(self, actions, condition=None, in_transaction=None):
```
garrettheel: was this meant to be removed?

tests/test_transaction.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_commit__not_implemented(self, mocker):
+        t = Transaction(connection=mocker.MagicMock())
+        with pytest.raises(NotImplementedError):
+            t._commit()
+
+
+class TestTransactGet:
+
+    def setup(self):
+        self.mock_model_cls = MagicMock(__name__='MockModel')
```
garrettheel: Lots of mocking here! Is it possible to switch out some of these mocks for real objects so that we're testing more throughly?

hallie: replaced some things!

tests/test_transaction.py
```diff
@@ -1,57 +1,97 @@
...
-            t._put_items = [{}]
-            t._update_items = [{}]
-
-        mock_connection.transact_write_items.assert_called_once_with(
+        connection = Connection()
+        mock_connection_transact_write = mocker.patch.object(connection, 'transact_write_items')
+        with patch(PATCH_METHOD) as req:
+            req.return_value = MOCK_TABLE_DESCRIPTOR
+            with TransactWrite(connection=connection) as t:
+                t._condition_check_items = [{}]
```
garrettheel: Could you use the real methods on `TransactWrite` to add values instead? Would be nice to have a unit test also exercise through the operations kwargs building logic

hallie: okidokies :0

hallie: added!

PR #620, opened 2019-05-16
make MapAttribute value can be StringSet

supercoderhawk:
If value in MapAttribute is StringSet, exception will occur. Now, This commit will fix this issue.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -13,7 +13,8 @@ from dateutil.tz import tzutc
 from inspect import getargspec
 from pynamodb.constants import (
     STRING, STRING_SHORT, NUMBER, BINARY, UTC, DATETIME_FORMAT, BINARY_SET, STRING_SET, NUMBER_SET,
-    MAP, MAP_SHORT, LIST, LIST_SHORT, DEFAULT_ENCODING, BOOLEAN, ATTR_TYPE_MAP, NUMBER_SHORT, NULL, SHORT_ATTR_TYPES
+    MAP, MAP_SHORT, LIST, LIST_SHORT, DEFAULT_ENCODING, BOOLEAN, ATTR_TYPE_MAP, NUMBER_SHORT, NULL, SHORT_ATTR_TYPES,
+    STRING_SET_SHORT
 )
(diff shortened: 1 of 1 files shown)


ikonst:
Good catch. Maybe add to this test?
https://github.com/pynamodb/PynamoDB/blob/86e9bbfa38cb9d84e082389558d3ca2225f62a8e/pynamodb/tests/test_attributes.py#L50-L52

supercoderhawk:
Unittest has been added

ikonst:
Shouldn't you also change `test_map_overridden_attrs_serialize`? I thought the crash happened when you serialize.

supercoderhawk:
Sorry, I didn't read the unittest code carefully. I have fixed it as your advice.

ikonst:
@supercoderhawk I tried reverting 96081a1bafcb9caebc304c780d11556804675fa5 on your branch and the newly added tests are still passing.

I didn't look into this yet, but superficially I see that you changed only `DESERIALIZE_CLASS_MAP` (but not `SERIALIZE_CLASS_MAP`) and change tests only for serialization.

supercoderhawk:
Dynamodb record example
```json
{"item_id": {
    "S": "1234"
  },
"interventions": {
    "L": [
      {
        "M": {
          "names": {
            "SS": [
              "A",
              "B",
              "C",
            ]
          },
        }
      }
    ]
  }
}
```

code will raise exception
```python
from pynamodb.attributes import UnicodeAttribute, ListAttribute
class ModelExample(Model):
    class Meta:
        table_name = 'test'
    item_id = UnicodeAttribute(has_key=True)
    interventions = ListAttribute(null=True)
```

ikonst:
In your sample code, you're using `ListAttribute` and not `MapAttribute`. Is this intentional?

supercoderhawk:
> In your sample code, you're using `ListAttribute` and not `MapAttribute`. Is this intentional?

But the item in list is a Map and this map item will cause the exception.

garrettheel:
@supercoderhawk I tried a few combinations (e.g list of maps containing a SS) and can't recreate the exception. Would you be able to write a test case reproducing it?

PR #628, opened 2019-05-24
Exclude tests from bdist

### LINKED ISSUE #593: Will not run under python 3.5 since 3.3.2
After upgrading to 3.3.3 from 3.3.0, found that my app was broken running under python 3.5 which should not have happened for a point release. After checking the releases, it seems due to the file `mypy_helpers.py` in the test directory using python 3.6 code on line 15:

    with open(f'{tempdirname}/__main__.py', 'w') as f:

Thanks

ikonst:
~This should ensure that tests for Python 3.6 features won't break installation on old Python versions.~


setup.py
```diff
@@ -31,7 +31,7 @@ def find_stubs(package):
 setup(
     name='pynamodb',
     version=__import__('pynamodb').__version__,
-    packages=find_packages(),
+    packages=find_packages(exclude=('pynamodb.tests',)),
```
garrettheel:
```suggestion
    packages=find_packages(exclude=('pynamodb.tests', 'pynamodb.tests.*')),
```

https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages suggests that we'll need this to also omit the integration test package

ikonst: From manually testing this, excluding the parent package was enough to exclude all tests in the resulting dist.

ikonst: @asottile says that this will exclude from sdist too. He also thinks we should `mv pynamodb/tests tests` and then `exclude=('tests*',)`.

And, regardless, we should dig into #593 some more, since it's a hack w.r.t #593.

garrettheel: I don't feel strongly about exactly where tests live, but if you've tested this and it works for subpackages too then happy to 🚢

asottile:
I think we should investigate the original issue further before pursuing this -- I can't think of a circumstance where an application would try and import pynamo's tests -- and even then it would trip over the test-only libraries being used just the same as syntax

let's get an original reproduction and make sure we're actually fixing something before merging :+1:

PR #631, opened 2019-05-31
Add TTL to pynamodb

irhkang:
Leverage DynamoDB's TTL support by creating a TTLAttribute.  This also adds a `default_for_new` to prevent people from creating TTLs on old objects during migrations.

Note: this alone does not delete items past the TTL.  You must first ensure your table has the correct IAM policies and enable TTLs for your dynamo table


pynamodb/models.py
```diff
@@ -1177,6 +1182,17 @@ def _get_indexes(cls):
                     cls._indexes[pythonic(LOCAL_SECONDARY_INDEXES)].append(idx)
         return cls._indexes
 
+    def _set_ttl_field(self):
+        """
+        Sets the TTL field specified in the Meta if the field is not set.
+        """
+        cls = type(self)
+        if cls.Meta.ttl_field and cls.Meta.ttl_seconds:
```
ikonst: Should we error preemptively if the TTL attribute's type is not `N`?

pynamodb/models.py
```diff
@@ -195,6 +195,10 @@ def __init__(cls, name, bases, attrs):
                         setattr(attr_obj, 'aws_access_key_id', None)
                     if not hasattr(attr_obj, 'aws_secret_access_key'):
                         setattr(attr_obj, 'aws_secret_access_key', None)
+                    if not hasattr(attr_obj, 'ttl_field'):
```
```suggestion
                    if not hasattr(attr_obj, 'ttl_attribute'):
```

pynamodb/models.py
irhkang: Ah yes.. I was thinking about that.  I wanted to make sure the approach was ok

pynamodb/models.py
```diff
@@ -195,6 +195,10 @@ def __init__(cls, name, bases, attrs):
...
+                        setattr(attr_obj, 'ttl_field', get_settings_value('ttl_field'))
```
ikonst: Do we want a global "ttl field" override?

pynamodb/models.py
```diff
@@ -195,6 +195,10 @@ def __init__(cls, name, bases, attrs):
...
+                    if not hasattr(attr_obj, 'ttl_seconds'):
+                        setattr(attr_obj, 'ttl_seconds', get_settings_value('ttl_seconds'))
```
ikonst: ditto

pynamodb/models.py
```diff
@@ -1177,6 +1182,17 @@ def _get_indexes(cls):
                     cls._indexes[pythonic(LOCAL_SECONDARY_INDEXES)].append(idx)
         return cls._indexes
 
+    def _set_ttl_field(self):
+        """
+        Sets the TTL field specified in the Meta if the field is not set.
```
mkuchenbecker: To be clear, this sets the field if it is not set. So enabling this on a model and performing a read-modify-write on all records would set the TTL for all records?

ikonst: Maybe call it "set_default" to mirror the existing `setdefault` pattern in Python?

pynamodb/models.py
irhkang: Yes I deliberately do this.  I do not want it at the attribute level because a table should only have 1 TTL field (ie: dynamo expects 1 TTL field).  Since a table should have 1 TTL field, it should only have 1 TTL second as well.

ikonst: I understand why it's one per model. I'm referring to `get_settings_value` which means "if it's not set on the model, I'll go to a project-level default".

pynamodb/models.py
irhkang: Correct.. The plan is to correctly set the TTLs for existing records before doing this (Although this can be quite difficult..)

ikonst:
I believe there's an initial "describe-table" API call when using PynamoDB. Can we also perform a "describe-time-to-live" API call to discover the configured `TimeToLiveDescription.AttributeName`? Otherwise we might have a mismatch between our assumed attribute name and the actual attribute name.

Similarly, should we perform an "update-time-to-live" during Create Table?

ikonst:
While we're not too deep yet, how about an alternative API such as:
```
class MyModel(Model):
   my_ttl = TTLAttribute(default=timedelta(days=3))
```

1) `MetaModel.__init__` would ensure there's no more than one `TTLAttribute` in a model.
2) The resulting instance attribute could be assigned either a `datetime` or a `timedelta`. If a `timedelta` is assigned, it's implicitly from now. This will be especially useful for `default=timedelta(...)`.
3) The attribute would internally be an `N`-typed attribute but deserialized into (i.e. readable as) a `datetime`. The type stub for `TTLAttribute` would reflect that.
4) The `Model.create_table` could be updated to call `update-time-to-live` API if the model has a `TTLAttribute`.
5) Optional: `Model._get_meta_data` could be updated to call the `describe-time-to-live` API to ensure that there's no mismatch, though that might be too much. (We don't seem to perform a similar check for other properties of a table.)

irhkang:
I didn't want to create a TTLAttribute because 

1. I didn't have any strict way of enforcing that 1 table only had 1 TTLAttribute.  I suppose I can check in the init function..
2. The current Pynamo attributes roughly map to a DynamoDB type.  Dynamo has no TTL data type.  TTL is just a concept Dynamo has.

That being said, I don't feel strongly about either way so I'm willing to change implementations.

irhkang:
I'm most worried about point 4) because it bypasses AWS's UI that lets you see what items will be dropped.  Someone can easily mis-set a TTL and drop all their data.  I'd rather someone explicitly set that flag.

ikonst:
> I'm most worried about point 4) because it bypasses AWS's UI that lets you see what items will be dropped. Someone can easily mis-set a TTL and drop all their data. I'd rather someone explicitly set that flag.

I'm not suggesting updating the TTL of an existing table, only to assign TTL in a newly created table. This is analogous to how we create GSIs when creating tables (in `create_table`), but we have no methods sync GSIs (i.e. add and remove as needed) of existing tables.

pynamodb/attributes.py
```diff
@@ -525,6 +535,36 @@ def deserialize(self, value):
...
+        Return the epoch representation (in seconds).
+        """
+        if value is None:
+            return None
+        elif isinstance(value, int):
+            return value
+        elif isinstance(value, timedelta):
+            return int(time.time() + value.total_seconds())
+        elif isinstance(value, datetime):
+            return int((value - datetime(1970,1,1)).total_seconds())
```
ikonst: I think it'll break with tz-aware datetimes:
```
TypeError: can't subtract offset-naive and offset-aware datetimes
```

In Python 3 we have `datetime.timestamp`, but for compatibility I think we'd have to do:
```
import calendar

return calendar.timegm(value.timetuple())
```

From testing:
```py
>>> import calendar 
>>> import time
>>> from datetime import datetime
>>> time.time() - calendar.timegm(datetime.utcnow().timetuple())
0.7275009155273438
>>> time.time() - calendar.timegm(datetime.now(tz=tzutc()).timetuple())
0.18059110641479492
```
(first one is naive, second is tz-aware, in both cases the delta is less than 1 second)

ikonst:
```suggestion
            return calendar.timegm(value.timetuple())
```

pynamodb/attributes.py
```diff
@@ -525,6 +535,36 @@ def deserialize(self, value):
...
+        else:
+            raise ValueError("TTLAttribute value must be an int, timedelta. or datetime.")
+
+    def deserialize(self, value):
+        """
+        Decode the epoch representation into a datetime
+        """
+        return datetime.utcfromtimestamp(value)
```
ikonst: Since UTCDateTimeAttribute returns a tz-aware datetime, maybe
```suggestion
        return datetime.utcfromtimestamp(value).replace(tzinfo=tzutc())
```

pynamodb/attributes.py
irhkang: Correct.. timestamp() breaks in python2.  That's why I didn't use it.  I'll play around with something..

pynamodb/attributes.py
irhkang: Didn't think it was necessary to have tz aware datetime but I'll add a tzinfo

pynamodb/attributes.py
irhkang: Also I'm fairly certain clendar.timegm only works for.. gm (ie: utc time)

Just verified.. 
```
>>> calendar.timegm(datetime.utcnow().timetuple()) - time.time()
-0.058157920837402344
>>> calendar.timegm(datetime.now().timetuple()) - time.time()
-25200.54909801483
```

ikonst: `datetime.now()` returns a naive timezone in your local timezone, but it's impossible to tell that it's not UTC (i.e. the results of `datetime.now()` and `datetime.utcnow()` are indistinguishable).

ikonst: You're right, though!
```
import calendar
import time
from datetime import datetime
from dateutil import tzutc, tzlocal
```
then
```
>>> time.time() - calendar.timegm(datetime.now(tz=tzlocal()).utctimetuple())
25200.258514881134
```

With `utctimetuple` it seems to be OK, though:
```
>>> time.time() - calendar.timegm(datetime.now(tz=tzlocal()).utctimetuple())
0.7876191139221191
>>> time.time() - calendar.timegm(datetime.now(tz=tzutc()).utctimetuple())
0.14149904251098633
```
```

File: pynamodb/attributes.py (hunk 7 above)
Comment (ikonst) [reply pynamodb/attributes.py]: Just for symmetry with UTCDateTimeAttribute.

File: pynamodb/connection/base.py [hunk 8]
```diff
@@ -616,6 +617,23 @@ def create_table(self,
...
+        """
+        operation_kwargs = {
+            TABLE_NAME: table_name,
+            TIME_TO_LIVE_SPECIFICATION: {
+                ATTR_NAME: ttl_attribute_name,
+                ENABLED: True,
+            }
+        }
+        try:
+            data = self.dispatch(UPDATE_TIME_TO_LIVE, operation_kwargs)
```
ikonst: nit -
```suggestion
            return self.dispatch(UPDATE_TIME_TO_LIVE, operation_kwargs)
```

irhkang: thought it was bad practice to return in a try catch?

Will change..

pynamodb/models.py
```diff
@@ -203,6 +204,10 @@ def __init__(cls, name, bases, attrs):
                     if attr_obj.attr_name is None:
                         attr_obj.attr_name = attr_name
 
+            ttl_attributes = [name for name, attr_obj in attrs.items() if isinstance(attr_obj, TTLAttribute)]
+            if len(ttl_attributes) > 1:
+                raise ValueError("1 TTLAttribute per model is allowed. {} are TTLAttributes".format(ttl_attributes))
```
```suggestion
                raise ValueError("The model has more than one TTL attribute: {}".format(", ".join(ttl_attr_names)))
```

pynamodb/models.py
```diff
@@ -203,6 +204,10 @@ def __init__(cls, name, bases, attrs):
                     if attr_obj.attr_name is None:
                         attr_obj.attr_name = attr_name
 
+            ttl_attributes = [name for name, attr_obj in attrs.items() if isinstance(attr_obj, TTLAttribute)]
```
```suggestion
            ttl_attr_names = [name for name, attr_obj in attrs.items() if isinstance(attr_obj, TTLAttribute)]
```

pynamodb/models.py
```diff
@@ -897,12 +906,16 @@ def create_table(cls, wait=False, read_capacity_units=None, write_capacity_units
...
-                        return
+                        break
                     else:
                         time.sleep(2)
                 else:
                     raise TableError("No TableStatus returned for table")
 
+        ttl_attribute = cls._ttl_attribute()
+        if ttl_attribute:
+            cls._get_connection().update_time_to_live(cls.Meta.table_name, ttl_attribute.attr_name)
```
ikonst: 👏

pynamodb/models.py
```diff
@@ -232,12 +237,16 @@ class Model(AttributeContainer):
     _index_classes = None
     DoesNotExist = DoesNotExist
 
-    def __init__(self, hash_key=None, range_key=None, **attributes):
+    def __init__(self, hash_key=None, range_key=None, previously_saved=False, **attributes):
```
ikonst: Maybe use a name that has a lower chance of being used by existing code, e.g. `__previously_saved`?

pynamodb/models.py
```diff
@@ -232,12 +237,16 @@ class Model(AttributeContainer):
...
         """
         :param hash_key: Required. The hash key for this object.
         :param range_key: Only required if the table has a range key attribute.
         :param attrs: A dictionary of attributes to set on this object.
         """
+        # This is for the default_for_new attribute value.  We assume an attribute
+        # is "new" if it has not been previously saved.  previously_saved will be set to true
+        # when deserializing JSON that dynamoDB returns
+        self._previously_saved = previously_saved
```
ikonst: Where is this used? Is this different from `is_new_object`?

irhkang: you're right.. I only use it in the constructor.  This is a vestige from an earlier implementation

pynamodb/connection/base.py
ikonst: Never thought it a bad practice -- but I'm willing to be convinced otherwise. If anything, I usually clean up variables that add indirection without making code clearer:
e.g.
```
result = foo()
return result
```
to
```
return foo()
```

pynamodb/models.py
irhkang: that makes it a class var and it makes tests fail.. changing back

ikonst: A ... class var?!

irhkang: I was too vague.. https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-a-single-and-a-double-underscore-before-an-object-name

ikonst:
```py
class Foo:
    def __init__(self, abc, __xyz):
        self.abc = abc
        self.xyz = __xyz

foo = Foo('abc', 'xyz')
print(foo.abc)
print(foo.xyz)
```
prints:
```
abc
xyz
```

irhkang: yea but I removed the self.xyz part.  I don't think 2 underscores are necessary.. I don't think there will be collisions.  It's just a parameter like `is_new_object`

ikonst: The point is, if someone currently has:
```
class MyModel(Model):
   is_new_object = UnicodeAttribute()
   ...
```
and does:
```
my_model = MyModel(is_new_object="hello world")
```
then before your change, their code is valid, and after your change, it has unexpected side effects. That's why I'm recommending a parameter name that would have lower chances of colliding with an existing attribute, especially since we're only using it as a means of internal signaling.

pynamodb/attributes.py
irhkang: I think this is almost right..

```
>>> calendar.timegm(datetime(2019, 6,5, tzinfo=pytz.timezone('America/New_York')).utctimetuple()) - 1559707200 # 1559707200 is 9PM 6/4/2019 SF time
3600
```

pynamodb/__init__.py
```diff
@@ -7,4 +7,4 @@
 """
 __author__ = 'Jharrod LaFon'
 __license__ = 'MIT'
-__version__ = '4.0.0b1'
+__version__ = '4.1.0b1'
```
garrettheel: Please don't bump the version here -- releases are managed separately

pynamodb/attributes.py
```diff
@@ -212,14 +220,14 @@ def _initialize_attributes(cls):
...
-    def __init__(self, **attributes):
+    def __init__(self, _previously_saved = False, **attributes):
         # The `attribute_values` dictionary is used by the Attribute data descriptors in cls._attributes
         # to store the values that are bound to this instance. Attributes store values in the dictionary
         # using the `python_attr_name` as the dictionary key. "Raw" (i.e. non-subclassed) MapAttribute
         # instances do not have any Attributes defined and instead use this dictionary to store their
         # collection of name-value pairs.
         self.attribute_values = {}
-        self._set_defaults()
+        self._set_defaults(_previously_saved)
```
ikonst:
```suggestion
        self._set_defaults(_previously_saved=_previously_saved)
```
to be safe

pynamodb/models.py
```diff
@@ -247,7 +252,7 @@ def __init__(self, hash_key=None, range_key=None, **attributes):
                     "This table has no range key, but a range key value was provided: {0}".format(range_key)
                 )
             attributes[self._dynamo_to_python_attr(range_keyname)] = range_key
-        super(Model, self).__init__(**attributes)
+        super(Model, self).__init__(_previously_saved, **attributes)
```
ikonst: to be safe:
```suggestion
        super(Model, self).__init__(_previously_saved=_previously_saved, **attributes)
```

pynamodb/models.py
```diff
@@ -897,12 +902,21 @@ def create_table(cls, wait=False, read_capacity_units=None, write_capacity_units
...
-                        return
+                        break
                     else:
                         time.sleep(2)
                 else:
                     raise TableError("No TableStatus returned for table")
 
+        ttl_attribute = cls._ttl_attribute()
+        if ttl_attribute:
+            # Some dynamoDB implementations (eg: dynalite) do not support updating TTLs so
```
garrettheel: I'd suggest a kwarg to ignore ttl failures (defaults to `False`) rather than compromising legitimate production usage of TTL because of a local dev server is missing a feature

pynamodb/attributes.py
```diff
@@ -525,6 +536,53 @@ def deserialize(self, value):
...
+    attr_type = NUMBER
+
+    def __set__(self, instance, value):
+        """
+        Force a TTLAttribute to have a UTC datetime value
+        """
+        if not instance or self._is_map_attribute_class_object(instance):
+            return
+        if isinstance(value, timedelta):
+            value = int(time.time() + value.total_seconds())
```
garrettheel: `time.time()` is a bug as it's timezone-dependent. Please fix and add tests to catch this case

irhkang: How is it timezone dependent? I've never heard about that  -  https://docs.python.org/2/library/time.html#time.time

ikonst: @garrettheel `time.time()` is seconds since epoch and the epoch is unambiguously defined in GMT

pynamodb/attributes.py
```diff
@@ -212,14 +219,14 @@ def _initialize_attributes(cls):
 @add_metaclass(AttributeContainerMeta)
 class AttributeContainer(object):
 
-    def __init__(self, **attributes):
+    def __init__(self, _previously_saved = False, **attributes):
```
ikonst: @garrettheel can you think of any more elegant way of letting the model know whether it's a newly constructed one or a deserialized one?

The tricky part is that it needs to be known during initialization, since that's where defaults are established, and we thought of introducing a concept of `default_for_new` i.e. defaults that only apply during initialization (of ostensibly new models) but don't get applied to existing models that are going through a load-save cycle:
```proto
class MyModel(Model):
    time_to_live = TTLAttribute(default_for_new=timedelta(days=366))
```

Unlike what was previously claimed, this feature cannot be re-implemented as a custom initializer -- see why:
https://gist.github.com/ikonst/ebc4b7f54079afedd7aba452544c1c71
Instead, you'd have to implement it in application code (rather than in the model):
```python
def my_app_handler(foo, bar):
    my_model = MyModel()
    my_model.id = str(uuid4())
    my_model.foo = foo
    my_model.bar = bar
    my_model.time_to_live = timedelta(days=366)
    my_model.save()
```

pynamodb/attributes.py
garrettheel: Yep, my bad here. I misunderstood the docs, sorry for the misdirection.

pynamodb/attributes.py
```diff
@@ -552,11 +551,12 @@ def __set__(self, instance, value):
         """
         if not instance or self._is_map_attribute_class_object(instance):
             return
+        current_timestamp_s = datetime.utcnow().timestamp()
```
garrettheel: I led you down the wrong track here. If we use a `datetime` instance then we'd need to make it tz aware for this to work correctly. Also I don't think that `timestamp()` is available in py2, so `time.time()` is indeed better here.

pynamodb/attributes.py
irhkang: In your gist.. Get calls the constructor of the model (so the value of the attribute becomes 42). https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/models.py#L404 calls https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/models.py#L438.

Refresh does not call the constructor - it just gets the values from dynamo and overwrites the model's attributes https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/models.py#L380.  Hence why the value is 7.

ikonst: Yeah, my case isn't even a good simulation of the actual scenario we're concerned about. Here's something more resembling our concern:
https://gist.github.com/ikonst/66de10eb86721ac343da581aed24ff82

irhkang: I think your gists basically validate the concern that putting something in the constructor of the model will not work given the current pynamo codebase

pynamodb/connection/base.py
```diff
@@ -613,6 +614,22 @@ def create_table(self,
...
+            TABLE_NAME: table_name,
+            TIME_TO_LIVE_SPECIFICATION: {
+                ATTR_NAME: ttl_attribute_name,
+                ENABLED: True,
+            }
+        }
+        try:
+            return self.dispatch(UPDATE_TIME_TO_LIVE, operation_kwargs)
+        except BOTOCORE_EXCEPTIONS as e:
+            raise TableError("Failed to update TTL on table: {0}".format(e), e)
```
```suggestion
            raise TableError("Failed to update TTL on table: {}".format(e), e)
```

ikonst:
@garrettheel you mean `update_time_to_live`, `update_ttl`?

garrettheel:
@ikonst I just noticed `create_table`, didn't check the rest yet!

pynamodb/attributes.pyi
```diff
@@ -125,6 +125,8 @@ class TTLAttribute(Attribute[datetime]):
     def __get__(self: _A, instance: None, owner: Any) -> _A: ...
     @overload
     def __get__(self, instance: Any, owner: Any) -> datetime: ...
+    def serialize(self, value: Any) -> Any: ...
```
ikonst: I don't think we need to repeat this, since the signature is the same as in the base class.

pynamodb/models.pyi
```diff
@@ -94,7 +94,16 @@ class Model(metaclass=MetaModel):
...
     def describe_table(cls): ...
     @classmethod
-    def create_table(cls: Type[_T], wait: bool = ..., read_capacity_units: Optional[Any] = ..., write_capacity_units: Optional[Any] = ...): ...
+    def create_table(
+        cls: Type[_T],
+        wait: bool = ...,
+        read_capacity_units: Optional[Any] = ...,
+        write_capacity_units: Optional[Any] = ...,
+        billing_mode: Optional[Any] = ...,
+        ignore_update_ttl_errors: Optional[bool] = ...
```
ikonst: not just `: bool`?

pynamodb/models.pyi
```diff
@@ -94,7 +94,16 @@ class Model(metaclass=MetaModel):
...
+    ): ...
+    @classmethod
+    def update_ttl(cls, ignore_update_ttl_errors): ...
```
ikonst:
```suggestion
    def update_ttl(cls, ignore_update_ttl_errors: bool): ...
```

pynamodb/models.pyi
irhkang: because you can't have required params after optional ones

ikonst: You're thinking of default params. "Optional" just means that it can be None.

irhkang: You're right - I'll change

ikonst:
@irhkang are there any other missing signatures?
@garrettheel any more comments before we merge this into the beta?

pynamodb/models.pyi
```diff
@@ -29,7 +29,7 @@ class Model(metaclass=MetaModel):
     _hash_keyname: Optional[str]
     _range_keyname: Optional[str]
     _connection: Optional[TableConnection]
-    def __init__(self, hash_key: Optional[KeyType] = ..., range_key: Optional[Any] = ..., **attrs) -> None: ...
+    def __init__(self, hash_key: Optional[KeyType] = ..., range_key: Optional[Any] = ..., _user_instantiated: bool, **attrs) -> None: ...
```
ikonst: @irhkang since it's an internal parameter, maybe it **shouldn't** be part of the typeshed?

PR #632, opened 2019-06-06
Remove deprecated conditional APIs

garrettheel:
The APIs in https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LegacyConditionalParameters.html have been deprecated for quite a long time now and they make for a lot of complexity in PynamoDB. While there's some migration effort, it's probably about time we dropped this and slimmed down.


pynamodb/connection/base.py
```diff
@@ -1127,8 +1103,6 @@ def rate_limited_scan(self,
                           attributes_to_get=None,
                           page_size=None,
                           limit=None,
-                          conditional_operator=None,
```
jpinner-lyft: I think we can just remove `rate_limited_scan` in favor of the paginators.

ikonst: TIL!

pynamodb/models.py
```diff
@@ -256,11 +254,6 @@ def has_map_or_list_attributes(cls):
                 return True
         return False
 
-    @classmethod
-    def _conditional_operator_check(cls, conditional_operator):
-        if conditional_operator is not None and cls.has_map_or_list_attributes():
```
jpinner-lyft: you should also be able to delete `has_map_or_list_attributes`

pynamodb/models.py
```diff
@@ -335,112 +328,34 @@ def __repr__(self):
...
-        if len(expected_values):
-            kwargs.update(expected=self._build_expected_values(expected_values, DELETE_FILTER_OPERATOR_MAP))
-        kwargs.update(conditional_operator=conditional_operator)
-        kwargs.update(condition=condition)
-        return self._get_connection().delete_item(*args, **kwargs)
 
-    def update_item(self, attribute, value=None, action=None, condition=None, conditional_operator=None, **expected_values):
-        """
-        Updates an item using the UpdateItem operation.
+        # TODO: raise exception if old args passed?
```
jpinner-lyft: python should have you covered:
`TypeError: delete() got an unexpected keyword argument 'conditional_operator'`

pynamodb/models.py
```diff
@@ -335,112 +328,34 @@ def __repr__(self):
...
-        for name, value in data.get(ATTRIBUTES).items():
-            attr_name = self._dynamo_to_python_attr(name)
-            attr = self.get_attributes().get(attr_name)
-            if attr:
-                setattr(self, attr_name, attr.deserialize(attr.get_value(value)))
-        return data
+        return self._get_connection().delete_item(*args, **kwargs)
 
-    def update(self, attributes=None, actions=None, condition=None, conditional_operator=None, **expected_values):
+    def update(self, actions=None, condition=None):
```
jpinner-lyft: Curious about your thoughts on this API. What I wanted was `def update(self, *actions, condition=None)` but I believe that syntax isn't supported in python 2.

pynamodb/models.py
```diff
@@ -335,112 +328,34 @@ def __repr__(self):
...
         """
         Updates an item using the UpdateItem operation.
 
-        :param attributes: A dictionary of attributes to update in the following format
-                            {
-                                attr_name: {'value': 10, 'action': 'ADD'},
-                                next_attr: {'value': True, 'action': 'PUT'},
-                            }
+        :param actions:
```
jpinner-lyft: should probably add `condition`

pynamodb/models.py
jpinner-lyft: At a minimum we should drop the `=None` make `actions` an arg and not a kwarg.

pynamodb/models.py
```diff
@@ -335,112 +328,34 @@ def __repr__(self):
...
         """
-        if attributes is not None and not isinstance(attributes, dict):
-            raise TypeError("the value of `attributes` is expected to be a dictionary")
         if actions is not None and not isinstance(actions, list):
```
jpinner-lyft: `actions` should not be None -- just check that it is a list and add a check that the list it isn't empty

pynamodb/models.py
```diff
@@ -554,10 +464,9 @@ def count(cls,
         :param consistent_read: If True, a consistent read is performed
         :param index_name: If set, then this index is used
         :param rate_limit: If set then consumed capacity will be limited to this amount per second
-        :param filters: A dictionary of filters to be used in the query. Requires a hash_key to be passed.
         """
         if hash_key is None:
-            if filters:
+            if filter_condition is not None:
                 raise ValueError('A hash_key must be given to use filters')
             return cls.describe_table().get(ITEM_COUNT)
```
jpinner-lyft: probably not an issue right now, but this seems broken -- IMHO the right thing to do here is to provide either a filtered query or a filtered scan -- not to use the (what i think is several hours delayed?) table description

pynamodb/models.py
```diff
@@ -633,13 +532,10 @@ def query(cls,
         :param limit: Used to limit the number of results returned
         :param scan_index_forward: If set, then used to specify the same parameter to the DynamoDB API.
             Controls descending or ascending results
-        :param conditional_operator:
         :param last_evaluated_key: If set, provides the starting point for query.
         :param attributes_to_get: If set, only returns these elements
         :param page_size: Page size of the query to DynamoDB
-        :param filters: A dictionary of filters to be used in the query
```
jpinner-lyft: should add `rate_limit`

pynamodb/models.py
```diff
@@ -689,116 +575,28 @@ def query(cls,
...
 
         :param filter_condition: Condition used to restrict the scan results
         :param segment: If set, then scans the segment
         :param total_segments: If set, then specifies total segments
         :param limit: Used to limit the number of results returned
-        :param conditional_operator:
         :param last_evaluated_key: If set, provides the starting point for scan.
         :param page_size: Page size of the scan to DynamoDB
-        :param filters: A list of item filters
         :param consistent_read: If True, a consistent read is performed
```
jpinner-lyft: should add `index_name` and `rate_limit` params

pynamodb/models.py
garrettheel: Yep, I think we'll have to wait for py3 for that interface. For now I'll make it an arg.

pynamodb/models.py
garrettheel: Yep, this is definitely a problem and a common confusion when using `.count`. Happy to consider changing this API for the 4.0.0 release or we wait until a future one.

pynamodb/connection/base.py
```diff
@@ -928,39 +847,13 @@ def update_item(self,
             operation_kwargs.update(self.get_item_collection_map(return_item_collection_metrics))
         if return_values:
             operation_kwargs.update(self.get_return_values_map(return_values))
-        if not actions and not attribute_updates:
-            raise ValueError("{0} cannot be empty".format(ATTR_UPDATES))
+        if not actions:
+            raise ValueError("'actions' cannot be empty")
         actions = actions or []
```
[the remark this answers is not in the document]
garrettheel: technically not from `Model`, but it's still taken as a kwarg and defaults to `None` so it might be better to keep this for now (short of a larger refactor)

pynamodb/connection/base.py
```diff
@@ -1442,91 +1143,17 @@ def query(self,
...
-                condition = ~condition
-            if condition_expression is None:
-                condition_expression = condition
-            elif conditional_operator == AND:
-                condition_expression = condition_expression & condition
-            else:
-                condition_expression = condition_expression | condition
-        return condition_expression.serialize(name_placeholders, expression_attribute_values)
-
     def _get_condition(self, table_name, attribute_name, operator, *values):
```
jpinner-lyft: we can probably drop this method and inline the one place we create an `__eq__` expression

pynamodb/connection/base.py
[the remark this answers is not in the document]
jpinner-lyft: the unnecessary line is `actions = actions or []`

README.rst
```diff
@@ -108,27 +108,21 @@ first name begins with 'D':
 
 .. code-block:: python
 
-    for user in UserModel.query("John", first_name__begins_with="D"):
+    for user in UserModel.query("John", filter_condition=(UserModel.first_name.startswith("D")):
```
jpinner-lyft: I think you should be able to drop the `filter_condition=` part -- at least that was my goal at one point when writing this:
`UserModel.query("John", UserModel.first_name.startswith("D"))`

README.rst
```diff
@@ -108,27 +108,21 @@ first name begins with 'D':
...
         print(user.first_name)
 
 Examples of ways to query your table with filter conditions:
 
 .. code-block:: python
 
-    for user in UserModel.query("John", filter_condition= (UserModel.email=="djohn@company.org")):
+    for user in UserModel.query("John", filter_condition=(UserModel.email=="djohn@company.org")):
```
jpinner-lyft: same as above: `UserModel.query("John", UserModel.email=="djohn@company.org")`
i know it's just my opinion but i think it reads better

README.rst
jpinner-lyft: another nit: this example is actually backwards -- `John` should be the range_key and `Denver` should be the hash_key

pynamodb/connection/base.py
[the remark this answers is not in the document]
garrettheel: oh duh, missed the exception above. removing.

README.rst
garrettheel: re `filter_condition`: this would require moving the kwarg to an arg too, right? are you suggesting we do that now?

fixed the example for you. it's probably not the most intuitive example we could give here, will consider rewriting

jpinner-lyft: it shouldn't -- the way this is supposed to work is that it without the keyword it will get assigned to the `range_key_condition` argument -- then the table will parse it and realize that it's actually a filter_condition since it doesn't match the range_key for the table / index and move it

garrettheel: Ah cool, didn't realize that case was handled. I'll update the example

PR #635, opened 2019-06-08
Drop python 2.6 compatibility code.

jpinner-lyft:


pynamodb/compat.py
```diff
@@ -1,79 +1,8 @@
...
-        except AttributeError:
-            continue
-
+    def predicate(value):
         value_cls = getattr(value, '__class__', None)
-        if value_cls and issubclass(value_cls, classinfo):
-            results.append((key, value))
-    results.sort()
-    return results
+        return value_cls and issubclass(value_cls, classinfo)
```
jpinner-lyft: @garrettheel do you have any idea why we essentially check `issubclass(value.__class__, classinfo)` instead of say `isinstance(value, classinfo)`?

As far as I can tell the use of `issubclass` is inherited from the initial work on the model api:
https://github.com/pynamodb/PynamoDB/blob/37bfe00/pynamodb/models.py#L116-L121

Any idea when this wouldn't be equivalent to calling `isinstance`? something, something, "old-style classes"?

On another note any idea why we do `getattr(value, '__class__', None)` instead of `value.__class__`? Looks like this was changed for "python 2 compatibility" back in https://github.com/pynamodb/PynamoDB/commit/c86e8fe . something, something, "old-style classes" again? cc/ @asottile

asottile: from what I can tell, this function is being used to looking for attributes which are `Attribute` descriptors -- pretty sure you can more safely simplify this to `isinstance(obj, classinfo)`

`isinstance` also works fine with old-style classes, though I don't think they are at play /  relevant here

there's technically another tricking thing which isn't being handled here, the `__class__` attribute could itself be a descriptor that raises an exception -- but I don't think this is important to handle

asottile:
If you want to remove even more python2.6 compat code, you can also run [pyupgrade](https://github.com/asottile/pyupgrade) -- the default mode is "remove python2.6 compatibility"

pynamodb/compat.py
jpinner-lyft: @asottile do you mind reviewing https://github.com/pynamodb/PynamoDB/pull/635/commits/82eef8f ?

PR #590, opened 2019-01-29
Adding billing mode. (Used #574 as reference)

leobarcellos:
- This fixes #574 large diffs and add tests

PR #637, opened 2019-06-09
Fix BillingMode compatibility

ikonst:
dynamodb-local doesn't seem to support the BillingMode property when creating tables. For “provisioning” billing mode, it's safe to omit it since it's the default.

Follow-up to #590


pynamodb/connection/base.py
```diff
@@ -568,9 +568,10 @@ def create_table(self,
 
         if billing_mode not in AVAILABLE_BILLING_MODES:
             raise ValueError("incorrect value for billing_mode, available modes: {0}".format(AVAILABLE_BILLING_MODES))
-        operation_kwargs[BILLING_MODE] = billing_mode
```
ikonst: This was simply unneeded.

ikonst:
It's a mystery how the code was passing tests, though, since we're using dynamodb-local for our tests.

ikonst:
You mean, about #590? (this is just a small fix on top of it)

PR #638, opened 2019-06-10
Remove MapAttributeMeta

jpinner-lyft:


jpinner-lyft:
More cleanup for the 4.0 release.

PR #639, opened 2019-06-10
Replace Model._get_meta_data with hash and range key attributes

jpinner-lyft:


pynamodb/models.py
```diff
@@ -224,7 +227,8 @@ class Model(AttributeContainer):
 
     # These attributes are named to avoid colliding with user defined
     # DynamoDB attributes
-    _meta_table = None
+    _hash_keyname = None
```
ikonst: Should we name this something like `_hash_python_keyname`?

PR #641, opened 2019-06-10
raise better error message when index does not exist

conjmurph:
`Connection.query` raises `ValueError("No hash key attribute for index: {}".format(index_name))` if it cannot find the hash key name for a given index. I interpret this error as the index exists, but there is an issue with the hash key. What is really happening is either 1. the index does not exist 2. the index exists, but it does not have a hash key attribute. In case 1, we should raise an exception that is explicit about the fact the index does not exist.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -121,15 +121,8 @@ class MetaTable(object):
         """
         global_indexes = self.data.get(GLOBAL_SECONDARY_INDEXES)
         local_indexes = self.data.get(LOCAL_SECONDARY_INDEXES)
-        indexes = []
-        if local_indexes:
-            indexes += local_indexes
-        if global_indexes:
-            indexes += global_indexes
-        for index in indexes:
-            if index.get(INDEX_NAME) == index_name:
-                return True
-        return False
+        indexes = (global_indexes or []) + (local_indexes or [])
+        return any(index.get(INDEX_NAME) == index_name for index in indexes)
 
     def get_index_hash_keyname(self, index_name):
         """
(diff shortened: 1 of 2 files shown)


pynamodb/connection/base.py
```diff
@@ -115,6 +115,22 @@ def get_key_names(self, index_name=None):
                 key_names.append(index_range_keyname)
         return key_names
 
+    def has_index_name(self, index_name):
+        """
+        Returns True if the base table has a global or local secondary index with index_name
+        """
+        global_indexes = self.data.get(GLOBAL_SECONDARY_INDEXES)
+        local_indexes = self.data.get(LOCAL_SECONDARY_INDEXES)
+        indexes = []
```
rowillia:
```python
indexes = (global_indexes or []) + (local_indexes or [])
```

pynamodb/connection/base.py
```diff
@@ -115,6 +115,22 @@ def get_key_names(self, index_name=None):
...
+        if local_indexes:
+            indexes += local_indexes
+        if global_indexes:
+            indexes += global_indexes
+        for index in indexes:
```
rowillia:
```python
return any(index.get(INDEX_NAME) == index_name for index in indexes)
```

pynamodb/connection/base.py
```diff
@@ -1083,6 +1099,8 @@ def query(self,
         if tbl is None:
             raise TableError("No such table: {}".format(table_name))
         if index_name:
+            if not tbl.has_index_name(index_name):
+                raise ValueError("Table {0} has no index: {1}".format(table_name, index_name))
```
```suggestion
                raise ValueError("Table {} has no index: {}".format(table_name, index_name))
```

conjmurph:
Anything left on this @rowillia @ikonst ?

ikonst:
Thanks for the contribution @conjmurph!

PR #643, opened 2019-06-11
Remove LegacyBooleanAttribute

jpinner-lyft:


pynamodb/tests/test_attributes.py
```diff
@@ -520,7 +480,7 @@ def test_boolean_deserialize(self):
         """
         attr = BooleanAttribute()
         assert attr.deserialize('1') is True
-        assert attr.deserialize('0') is True
+        assert attr.deserialize('0') is False
```
jpinner-lyft: @garrettheel this seems like it might be a real bug in master

jmphilli:
i'm for this - i wanted to do this back when we went from 2->3 i think

pynamodb/tests/test_attributes.py
```diff
@@ -519,8 +479,6 @@ def test_boolean_deserialize(self):
         BooleanAttribute.deserialize
         """
         attr = BooleanAttribute()
-        assert attr.deserialize('1') is True
-        assert attr.deserialize('0') is True
         assert attr.deserialize(True) is True
```
betamoo: should we leave the tests of of 1 and 0 too?

jpinner-lyft: /shrug i deleted the backwards-compatibility so they won't work anyways

betamoo: this will only be in 4.x right? So that current user should not be affect

betamoo:
+1

pynamodb/tests/test_attributes.py
jpinner-lyft: correct -- this is for the `4.x` release so anyone who has followed along with the migration code we released and done the upgrade should not be affected

Issue #627, opened 2019-05-24
JSONAttribute(null=True) field TypeError: the JSON object must be str, bytes or bytearray, not 'NoneType' 


monkut:
> pynamodb 3.3.3
> Local testing done with Localstack: image: localstack/localstack

I have the following model:

```
class CollectionStateIndex(GlobalSecondaryIndex):

    class Meta:
        index_name = settings.DYNAMODB_COLLECTION_STATE_INDEXNAME
        read_capacity_units = 2
        write_capacity_units = 1

        projection = AllProjection()
        host = settings.DYNAMODB_ENDPOINT

    collection_id = UnicodeAttribute(hash_key=True)
    state = UnicodeAttribute(range_key=True)


class S3ItemModel(Model):
    s3_uri = UnicodeAttribute()
    collection_id = UnicodeAttribute()
    request_id = UnicodeAttribute(hash_key=True)
    state = UnicodeAttribute()
    created_at_timestamp = NumberAttribute()
    updated_at_timestamp = NumberAttribute()
    result = JSONAttribute(null=True)
    errors = JSONAttribute(null=True)

    collections_state_index = CollectionStateIndex()

    class Meta:
        table_name = settings.DYNAMODB_S3ITEMMODEL_TABLENAME
        region = settings.AWS_REGION
        host = settings.DYNAMODB_ENDPOINT
        read_capacity_units = 2
        write_capacity_units = 1
```

And it appears that when performing a query for a result where the `errors` value is Null, the query fails to properly decode the item with the exception:

> TypeError: the JSON object must be str, bytes or bytearray, not 'NoneType'
```
item = next(results)
File "/var/task/pynamodb/pagination.py", line 183, in __next__
item = self._map_fn(item)
File "/var/task/pynamodb/models.py", line 529, in from_raw_data
kwargs[attr_name] = attr.deserialize(attr.get_value(value))
File "/var/task/pynamodb/attributes.py", line 439, in deserialize
return json.loads(value, strict=False)
File "/var/lang/lib/python3.6/json/__init__.py", line 348, in loads
'not {!r}'.format(s.__class__.__name__))
TypeError: the JSON object must be str, bytes or bytearray, not 'NoneType'
```

Query:
```
        results = S3ItemModel.query(
            hash_key=request_id,
        )
        item = next(results)
```

DynamoDB Record:
```
{
  "collection_id": {
    "S": "collection:11"
  },
  "created_at_timestamp": {
    "N": "1558593743"
[...]

Issue #424, opened 2017-12-07
After upgrading from 3.2.0 to 3.2.1 attributes with default values get reset to None after a refresh


canoon:

ikonst:
cc @jpinner-lyft any idea what went wrong w.r.t default and refresh?

jpinner-lyft:
The default values only get set during AttributeContainer initialization, not during refresh. I'm not sure if this is a bug or expected behavior. The linked PR fixes "refresh" to set all attribute values to whatever is in the database (as opposed to ignoring unknown attribute values). I would maybe ask @garrettheel what he thinks the behavior of defaults should be after a refresh where the attribute isn't in the database.

garrettheel:
@jpinner-lyft given the inability to store `null` in the database, I'd agree that this is indeed unexpected behavior and warrants a fix like the one you propose

samopus1io:
Will this ever be addressed? 

We rely on adding new BoolAttributes with default=True It has always worked  (getting True when no value was explicitly set), but we were surprised to hit a production issue where using refresh() would reset the value to None and then throwing an exception on subsequent save because None was not allowed.

PR #644, opened 2019-06-12
Always reset all model attributes when loading from dynamo.

jpinner-lyft:
An extension of the fix for issue #166: whenever loading data from dynamo, override all model data.

This is a breaking change, especially for attribute "defaults". Previously any model loading except for refresh (scan, query, get, etc.) would set any attributes that did not appear in dynamo to their default value. Refresh would set attributes that did not appear in dynamo to `None`.

With this change, any model loaded from dynamo would reflect the stored data. Attribute "defaults" would only apply to user-instantiated models.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -418,13 +418,9 @@ class Model(AttributeContainer):
         if data is None:
             raise ValueError("Received no data to construct object")
 
-        attributes = {}
-        for name, value in data.items():
-            attr_name = cls._dynamo_to_python_attr(name)
-            attr = cls.get_attributes().get(attr_name, None)
-            if attr:
-                attributes[attr_name] = attr.deserialize(attr.get_value(value))
-        return cls(**attributes)
+        item = cls()
+        item._deserialize(data)
+        return item
 
     @classmethod
     def count(cls,
@@ -714,9 +710,7 @@ class Model(AttributeContainer):
             attributes[range_keyname] = {
                 range_keytype: range_key
             }
-        item = cls()
-        item._deserialize(attributes)
-        return item
+        return cls.from_raw_data(attributes)
 
     @classmethod
     def _get_schema(cls):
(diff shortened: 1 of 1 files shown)


jpinner-lyft:
@garrettheel @ikonst we should probably move the deserialize logic to AttributeContainer -- subclassed map attributes still behave differently from models w.r.t. attribute defaults

jpinner-lyft:
@ikonst notes that this relates to https://github.com/pynamodb/PynamoDB/issues/424

ikonst:
This would also address #627

pynamodb/models.py
```diff
@@ -897,18 +891,18 @@ def _get_connection(cls):
...
-            value = attrs.get(attr.attr_name, None)
+            value = data.get(attr.attr_name)
+            if value is not None:
+                value = attr.get_value(value)
             if value is not None:
-                value = value.get(ATTR_TYPE_MAP[attr.attr_type], None)
-                if value is not None:
-                    value = attr.deserialize(value)
+                value = attr.deserialize(value)
             setattr(self, name, value)
```
ikonst: The changes to this function are purely syntactic?

ikonst:
> we should probably move the deserialize logic to AttributeContainer 

Ah, yes, we should. Here or in a follow-up?

pynamodb/models.py
```diff
@@ -317,19 +325,22 @@ def batch_write(cls, auto_commit=True):
...
 
     def delete(self, condition=None):
         """
         Deletes this object from dynamodb
         """
-        args, kwargs = self._get_save_args(attributes=False, null_check=False)
-        kwargs.update(condition=condition)
+        hash_key, range_key = self._serialize_keys(self._hash_key, self._range_key)
+        args = (hash_key,)
+        kwargs = dict(
```
ikonst: why not just
```
return self._get_connection().delete_item(*args, range_key=range_key, condition=condition)
```

pynamodb/models.pyi
```diff
@@ -30,6 +30,10 @@ class Model(metaclass=MetaModel):
     _range_keyname: Optional[str]
     _connection: Optional[TableConnection]
     def __init__(self, hash_key: Optional[KeyType] = ..., range_key: Optional[Any] = ..., **attrs) -> None: ...
+    @property
+    def _hash_key(self) -> KeyType: ...
+    @property
+    def _range_key(self) -> Optional[Any]: ...
```
hallie: why not `Optional[KeyType]`? (I recognize that the `__init__` has `range_key: Optional[Any]`, but I'm curious why that it for both)

PR #648, opened 2019-06-12
Require botocore>=1.12.54

### LINKED ISSUE #647: pynamodb uses botocore 1.12 features
In #590, we’ve merged `BillingMode` support to pynamodb. We currently require `botocore>=1.11.0`:
https://github.com/pynamodb/PynamoDB/blob/71c33e8fce228c02ad0aa327a0dedaabb2bfd15d/setup.py#L27
However, `BillingMode` is only available since botocore ~1.12.166~ 1.12.54:
https://github.com/boto/botocore/commit/224642e01396fb7683757ba03451384e1184ffc9

Moreover, #637 was misguided. Despite what I claimed in the comments, integration tests are, indeed, running, and I was simply using locally an old botocore version.

ikonst:
`BillingMode` was introduced in botocore 1.12.54, so we'll set it as the minimal version and partially undo #637.

Closes #647.


setup.py
```diff
@@ -24,7 +24,7 @@ def find_stubs(package):
 
 install_requires = [
     'six',
-    'botocore>=1.11.0',
+    'botocore>=1.12.167',
```
garrettheel: Can we just set this to the earliest version we need (`1.12.54`)? Requiring the latest version is more likely to introduce conflicts with other packages

ikonst: Done.

PR #653, opened 2019-06-22
Allow None to be used to bootstrap condition chaining.

jpinner-lyft:
This change attempts to allow users to more easily chain condition statements:
```
condition = None
if (foo):
    condition &= foo_condition
if (bar):
    condition &= bar_condition
if (baz):
    condition &= baz_condition
```

### DIFF
--- a/pynamodb/expressions/condition.py
+++ b/pynamodb/expressions/condition.py
@@ -42,13 +42,6 @@ class Condition(object):
                             .format(self.__class__.__name__, other.__class__.__name__))
         return Or(self, other)
 
-    def __ror__(self, other):
-        # special case 'None | condition' to enable better syntax for chaining
-        if other is not None:
-            raise TypeError("unsupported operand type(s) for |: '{}' and '{}'"
-                            .format(other.__class__.__name__, self.__class__.__name__))
-        return self
-
     def __invert__(self):
         return Not(self)
 
--- a/pynamodb/expressions/condition.py
+++ b/pynamodb/expressions/condition.py
@@ -29,11 +29,25 @@ class Condition(object):
                             self.__class__.__name__, other.__class__.__name__)
         return And(self, other)
 
+    def __rand__(self, other):
+        # special case 'None & condition' to enable better syntax for chaining
+        if other is not None:
+            raise TypeError("unsupported operand type(s) for &: '{0}' and '{1}'",
+                            self.__class__.__name__, other.__class__.__name__)
+        return self
+
     def __or__(self, other):
         if not isinstance(other, Condition):
             raise TypeError("unsupported operand type(s) for |: '{0}' and '{1}'",
                             self.__class__.__name__, other.__class__.__name__)
         return Or(self, other)
+
+    def __ror__(self, other):
+        # special case 'None | condition' to enable better syntax for chaining
+        if other is not None:
+            raise TypeError("unsupported operand type(s) for |: '{0}' and '{1}'",
+                            self.__class__.__name__, other.__class__.__name__)
+        return self
 
     def __invert__(self):
         return Not(self)
--- a/pynamodb/expressions/condition.py
+++ b/pynamodb/expressions/condition.py
@@ -25,28 +25,28 @@ class Condition(object):
 
     def __and__(self, other):
         if not isinstance(other, Condition):
-            raise TypeError("unsupported operand type(s) for &: '{0}' and '{1}'",
-                            self.__class__.__name__, other.__class__.__name__)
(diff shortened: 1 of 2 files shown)


pynamodb/expressions/condition.py
```diff
@@ -29,12 +29,26 @@ def __and__(self, other):
                             self.__class__.__name__, other.__class__.__name__)
         return And(self, other)
 
+    def __rand__(self, other):
+        # special case 'None & condition' to enable better syntax for chaining
+        if other is not None:
+            raise TypeError("unsupported operand type(s) for &: '{0}' and '{1}'",
+                            self.__class__.__name__, other.__class__.__name__)
```
ikonst: Did you mean to format?
```python
try:
    raise TypeError('foo: {0}', 'bar')
except Exception as ex:
    print(str(ex))
```
prints
```
('foo: {0}', 'bar')
```

pynamodb/tests/test_expressions.py
```diff
@@ -271,6 +271,15 @@ def test_and(self):
         assert placeholder_names == {'foo': '#0'}
         assert expression_attribute_values == {':0': {'S': 'bar'}, ':1': {'S': 'baz'}}
 
+    def test_rand(self):
```
ikonst: Should we also test something like `42 & self.attribute < 'bar'` to exercise the type error branch?

pynamodb/expressions/condition.py
jpinner-lyft: I copied the format error from above but I think it's wrong their too -- will fix all instances

pynamodb/expressions/condition.py
```diff
@@ -25,28 +25,28 @@ def serialize(self, placeholder_names, expression_attribute_values):
 
     def __and__(self, other):
         if not isinstance(other, Condition):
-            raise TypeError("unsupported operand type(s) for &: '{0}' and '{1}'",
-                            self.__class__.__name__, other.__class__.__name__)
+            raise TypeError("unsupported operand type(s) for &: '{0}' and '{1}'"
```
ikonst: nit - I think `pyupgrade` skipped the `{0}`s during its rewrite due to the omitted `.format`

pynamodb/expressions/condition.py
```diff
@@ -25,16 +25,30 @@ def serialize(self, placeholder_names, expression_attribute_values):
 
     def __and__(self, other):
         if not isinstance(other, Condition):
```
hallie: sorry if this is a silly question, but why introduce `__rand__` and `__ror__`, instead of just allowing `other: Optional[Condition]`? Where it returns `self` if `other is None` and returns the joined condition if `isinstance(other, Condition)`

hallie: I ask because when I first saw this question come up, it reminded me of all of the times I've had to write something along the lines of:
```
existing_condition = additional_condition & existing_condition if existing_condition is not None else additional_condition
```
:P

jpinner-lyft: This should allow you to write that as:
```
existing_condition &= additional_condition
```
The reason to introduce `__rand__` and `__ror__` is because without them `None & additional_condition` would raise a `TypeError` since `&` is not a valid operand on `None`.

hallie: O wait, sorry, I think my misunderstanding was what `__rand__` and `__ror__` stood for :0 Didn't realize they were basically "reverse" statement checks, and assumed they had something to do with the `&=`/`|=` operators themselves--nevermind!

jpinner-lyft:
@ikonst thinking of deleting the `__ror__` part of this PR

From just a logical perspective, returning results that meet "no condition and this condition" is equivalent to "this condition", but the same isn't true about `or`.

Thoughts?

ikonst:
Agreed, it's inconsistent for the `or` case.

With `and`, it aligns nicely, since `None` already denotes 'no filtering' in the underlying DynamoDB API:
```python
def get_users_in_postal_code(postal_code, name, max_age):
    condition = None
    if name:
       condition &= (User.name = name)
    if max_age:
       condition &= (User.age <= max_age)
    return User.postal_code_index.query(postal_code, condition=condition)
```

If there was a "sentinel" value (e.g. `builtins.False)` that'd cause `query` to either raise an error or (locally) return an empty result set, then we'd have something like:
```python
def get_demographic_in_postal_code(postal_code, name, age):
    condition = False
    if name:
       condition |= (User.name = name)
    if age:
       condition |= (User.age = age)
    return User.postal_code_index.query(postal_code, condition=condition)
```

(not sure how likely is such an access pattern for DynamoDB, but...)

PR #658, opened 2019-07-05
adding helper functions to get error code and message from response

hallie:
I find myself pretty often writing (or seeing other people write) helper methods to get the boto error code/message from the pynamo error (especially for things like asserting the `Code` is a conditional check failure). Want to add helper methods to more easily extract that info


pynamodb/exceptions.py
```diff
@@ -14,6 +14,16 @@ def __init__(self, msg=None, cause=None):
         self.cause = cause
         super(PynamoDBException, self).__init__(self.msg)
 
+    def get_cause_response_code(self):
+        if hasattr(self.cause, 'response'):
```
ikonst: nit -
```python
return getattr(self.cause, 'response', {}).get('Error', {}).get('Code')
```

ikonst: Also, seems rather lightweight and no side effects, maybe make it a property?

hallie: sounds good!

hallie: updated

PR #660, opened 2019-07-09
Catch IndexError when trying to fast parse a date time  string value

bmsilva:
When creating models for a already existing database, The ```UTCDateTimeAttribute``` fails to deserialise a date in the format ```2019-07-01T00:00:00``` because it doesn't have the tzinfo, example:

```
File “.../versions/3.7.0/lib/python3.7/site-packages/pynamodb/attributes.py”, line 897, in _fast_parse_utc_datestring
   datestring[13] != ‘:’ or datestring[16] != ‘:’ or datestring[19] != ‘.’ or
IndexError: string index out of range
```

In previous versions of pynamodb it was working, so I'm proposing this fix.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -565,6 +565,8 @@ class UTCDateTimeAttribute(Attribute):
                 return datetime.strptime(value, DATETIME_FORMAT)
             except ValueError:
                 return parse(value)
+        except IndexError:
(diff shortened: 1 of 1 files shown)


ikonst:
cc @rowillia can review?

garrettheel:
Thanks for the fix!

PR #661, opened 2019-07-09
making docs for transactions

hallie:


docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
+Transaction Operations
+======================
+
+Transact operations are supported using context managers. The DynamoDB API has limits for the number of items in
```
ikonst:
```suggestion
Transact operations are supported using context managers. The DynamoDB API has limits on the number of items in
```

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
+Transaction Operations
+======================
+
+Transact operations are supported using context managers. The DynamoDB API has limits on the number of items in
```
ikonst:
```suggestion
Transaction operations are supported using context managers. The DynamoDB API has limits on the number of items in
```

ikonst: I'd probably start with describing what transaction operations are, and only later go into the "we don't paginate yet" caveat.

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
...
+each request, but PynamoDB doesn't currently handle grouping or paginating, so this is something you must handle on your
+own.
+
+.. note::
+
+    An ordered array of up to 25 `TransactWriteItem` objects, each of which contains a `ConditionCheck`, `Put`, `Update`, or `Delete` object. These can operate on items in different tables, but the tables must reside in the same AWS account and Region, and no two of them can operate on the same item.
```
ikonst: Is this documenting the underlying DynamoDB API or does this talk in PynamoDB terms?

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Transact Writes
+^^^^^^^^^^^^^^^
+
+Here's an example of using a context manager for a `TransactWrite` operation:
+
+.. code-block:: python
+
+    with TransactWrite(client_request_toke='super-unique-key') as transaction:
+        """ attempting to transfer funds from user1's account to user2's """
```
ikonst:
```suggestion
        # attempting to transfer funds from user1's account to user2's
```

docs/transaction.rst
hallie: this the underlying dynamodb api

ikonst: I'd keep our docs to be "pynamodb first". We only need to go into DynamoDB API (which is documented by AWS already) if we need to explain something that's unclear from the PynamoDB API.

garrettheel: Perhaps something like "Keep in mind that DynamoDB imposes limits on the number of items that a single transaction can contain." with a link to their docs

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
...
+
+
+Transact Writes
+^^^^^^^^^^^^^^^
+
+Here's an example of using a context manager for a `TransactWrite` operation:
+
+.. code-block:: python
+
+    with TransactWrite(client_request_toke='super-unique-key') as transaction:
```
garrettheel:
```suggestion
    with TransactWrite(client_request_token='super-unique-key') as transaction:
```

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
...
+.. code-block:: python
+
+    from pynamodb.models import Model
+    from pynamodb.attributes import BooleanAttribute, NumberAttribute, UnicodeAttribute
+
+    class BankStatement(Model):
+        class Meta:
+            table_name = 'BankStatement'
+
+    user_id = UnicodeAttribute(hash_key=True)
```
garrettheel: The indentation is off here

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
...
+    is_active = BooleanAttribute()
+
+
+Transact Writes
+^^^^^^^^^^^^^^^
+
+Here's an example of using a context manager for a `TransactWrite` operation:
+
+.. code-block:: python
+
```
garrettheel: It would be helpful to include the import here, i.e

```
from pynamodb.connection.transactions import TransactWrite
```

As I write this, I'm also realizing that this API expects folks to import out of `.connection`, which we would usually reserve for lower-level operations. What do you think about bumping the package up one level for the 4.0 release? i.e `from pynamodb.transactions import ...`?

docs/transaction.rst
garrettheel: We'll also need to pass a `connection=` kwarg here

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
...
+    with TransactWrite(client_request_toke='super-unique-key') as transaction:
+        # attempting to transfer funds from user1's account to user2's
+        transfer_amount = 1000
+        transaction.update(
+            BankStatement(user_id='user1'),
+            actions=[BankStatement.account_balance.add(transfer_amount * -1)],
+            condition=(
+                (BankStatement.account_balance >= transfer_amount) &
+                (BankStatement.is_active)
```
garrettheel: You'll need `is_active == True` for both of these

docs/transaction.rst
```diff
@@ -0,0 +1,57 @@
...
+
+    user_id = UnicodeAttribute(hash_key=True)
+    account_balance = NumberAttribute(default=0)
+    is_active = BooleanAttribute()
+
+
+Transact Writes
+^^^^^^^^^^^^^^^
+
+Here's an example of using a context manager for a `TransactWrite` operation:
```
garrettheel: It could be helpful to provide some setup values that would result in the transaction succeeding, i.e

```
BankStatement('user1', account_balance=2000, is_active=True).save()
BankStatement('user2', account_balance=0, is_active=True).save()
```

docs/transaction.rst
hallie: sure, i can move it :0

hallie: https://github.com/pynamodb/PynamoDB/pull/667

docs/transaction.rst
```diff
@@ -0,0 +1,225 @@
+Transaction Operations
+======================
+
+Transact operations are similar to Batch operations, with the key differences being that the writes support the
+inclusion of condition checks, and they all must fail or succeed together.
+
+
+Transaction operations are supported using context managers. Keep in mind that DynamoDB imposes limits on the number of
+items that a single transaction can contain. PynamoDB doesn't currently handle automatic grouping or paginating, so this
```
garrettheel: The note about grouping/pagination is probably unnecessary, since it would remove transaction semantics anyways

docs/transaction.rst
```diff
@@ -0,0 +1,225 @@
...
+
+To retrieve the resolved model, you say `model_future.get()`. Any attempt to access this model before the transaction is complete
+will result in a `InvalidStateError`:code:.
+
+Error Types
+^^^^^^^^^^^
+
+You can expect some new error types with transactions, such as:
+
+* `TransactWriteError`:code: - thrown when a `TransactWrite`:code: request returns a bad response.
```
garrettheel: note that you can just use double two backticks to denote code blocks. e.g 
```
``TransactWriteError``
```

docs/transaction.rst
```diff
@@ -0,0 +1,225 @@
...
+        user2_statement.refresh()
+        # and both models should be unchanged
+        assert user1_statement.account_balance == 1000
+        assert user2_statement.account_balance == 1000
+
+
+Condition Check
+---------------
+
+The `ConditionCheck`:code: operation is used on a `TransactWrite`:code: to check if the current state of a record you
```
garrettheel: when referring to classes, you can use the following syntax to link directly to api docs for them:
```
:py:class:`UnicodeAttribute <pynamodb.attributes.UnicodeAttribute>`
```

docs/transaction.rst
```diff
@@ -0,0 +1,225 @@
...
+aren't modifying within the overall transaction fits some criteria that, if it fails, would cause the entire
+transaction to fail. The `condition`:code: argument is of type `Condition <https://pynamodb.readthedocs.io/en/latest/conditional.html>`_.
+
+* `model_cls`:code: (required)
+* `hash_key`:code:  (required)
+* `range_key`:code: (optional)
+* `condition`:code: (required) - of type `Condition <https://pynamodb.readthedocs.io/en/latest/conditional.html>`_
```
garrettheel: you can use a `ref` directive here instead of linking to support browsing docs at different versions.

take a look at the following example from `docs/tutorial.md`, which does:

```
:ref:`updates`
```

PR #664, opened 2019-07-18
Support for versioned optimistic locking a la DynamoDBMapper

ezmuller:
Adds a new nullable version attribute type which initializes at 1 and increments on save/update for optimistic locking.

### DIFF
--- a/docs/optimistic_locking.rst
+++ b/docs/optimistic_locking.rst
@@ -1,0 +1,172 @@
+==================
+Optimistic Locking
+==================
+
+Optimistic Locking is a strategy for ensuring that your database writes are not overwritten by the writes of others.
+With optimistic locking, each item has an attribute that acts as a version number. If you retrieve an item from a
+table, the application records the version number of that item. You can update the item, but only if the version number
+on the server side has not changed. If there is a version mismatch, it means that someone else has modified the item
+before you did. The update attempt fails, because you have a stale version of the item. If this happens, you simply
+try again by retrieving the item and then trying to update it. Optimistic locking prevents you from accidentally
+overwriting changes that were made by others. It also prevents others from accidentally overwriting your changes.
+
+.. warning:: - Optimistic locking will not work properly if you use DynamoDB global tables as they use last-write-wins for concurrent updates.
+
+See also:
+`DynamoDBMapper Documentation on Optimistic Locking <https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html>`_.
+
+Version Attribute
+-----------------
+
+To enable optimistic locking for a table simply add a ``VersionAttribute`` to your model definition.
+
+.. code-block:: python
+
+  class OfficeEmployeeMap(MapAttribute):
+      office_employee_id = UnicodeAttribute()
+      person = UnicodeAttribute()
+
+      def __eq__(self, other):
+          return isinstance(other, OfficeEmployeeMap) and self.person == other.person
+
+      def __repr__(self):
+          return str(vars(self))
+
+
+  class Office(Model):
+      class Meta:
+          read_capacity_units = 1
+          write_capacity_units = 1
+          table_name = 'Office'
+          host = "http://localhost:8000"
+      office_id = UnicodeAttribute(hash_key=True)
+      employees = ListAttribute(of=OfficeEmployeeMap)
+      name = UnicodeAttribute()
+      version = VersionAttribute()
+
+The attribute is underpinned by an integer which is initialized with 1 when an item is saved for the first time
+and is incremented by 1 with each subsequent write operation.
+
+.. code-block:: python
+
+  justin = OfficeEmployeeMap(office_employee_id=str(uuid4()), person='justin')
+  garrett = OfficeEmployeeMap(office_employee_id=str(uuid4()), person='garrett')
+  office = Office(office_id=str(uuid4()), name="office", employees=[justin, garrett])
+  office.save()
+  assert office.version == 1
+
+  # Get a second local copy of Office
+  office_out_of_date = Office.get(office.office_id)
+
+  # Add another employee and persist the change.
+  office.employees.append(OfficeEmployeeMap(office_employee_id=str(uuid4()), person='lita'))
+  office.save()
+  # On subsequent save or update operations the version is also incremented locally to match the persisted value so
+  # there's no need to refresh between operations when reusing the local copy.
+  assert office.version == 2
+  assert office_out_of_date.version == 1
+
+The version checking is implemented using DynamoDB conditional write constraints, asserting that no value exists
+for the version attribute on the initial save and that the persisted value matches the local value on subsequent writes.
+
+
+Model.{update, save, delete}
+----------------------------
+These operations will fail if the local object is out-of-date.
+
+.. code-block:: python
+
+  @contextmanager
+  def assert_condition_check_fails():
+      try:
+          yield
+      except (PutError, UpdateError, DeleteError) as e:
+          assert isinstance(e.cause, ClientError)
+          assert e.cause_response_code == "ConditionalCheckFailedException"
+      except TransactWriteError as e:
+          assert isinstance(e.cause, ClientError)
+          assert e.cause_response_code == "TransactionCanceledException"
+          assert "ConditionalCheckFailed" in e.cause_response_message
+      else:
+          raise AssertionError("The version attribute conditional check should have failed.")
+
+
+  with assert_condition_check_fails():
+      office_out_of_date.update(actions=[Office.name.set('new office name')])
+
+  office_out_of_date.employees.remove(garrett)
+  with assert_condition_check_fails():
+      office_out_of_date.save()
+
+  # After refreshing the local copy our write operations succeed.
+  office_out_of_date.refresh()
+  office_out_of_date.employees.remove(garrett)
+  office_out_of_date.save()
+  assert office_out_of_date.version == 3
+
+  with assert_condition_check_fails():
+      office.delete()
+
+Transactions
+------------
+
+Transactions are supported.
+
+Successful
+__________
+
+.. code-block:: python
+
+  connection = Connection(host='http://localhost:8000')
+
+  office2 = Office(office_id=str(uuid4()), name="second office", employees=[justin])
+  office2.save()
+  assert office2.version == 1
+  office3 = Office(office_id=str(uuid4()), name="third office", employees=[garrett])
+  office3.save()
+  assert office3.version == 1
+
+  with TransactWrite(connection=connection) as transaction:
+      transaction.condition_check(Office, office.office_id, condition=(Office.name.exists()))
+      transaction.delete(office2)
+      transaction.save(Office(office_id=str(uuid4()), name="new office", employees=[justin, garrett]))
+      transaction.update(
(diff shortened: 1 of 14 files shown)


ikonst:
@garrettheel @jpinner-lyft This is a pretty well-done contribution!

We have been traditionally rejecting new features that don't map somehow to core DynamoDB features, instead pointing authors at [pynamodb-attributes](https://pypi.org/project/pynamodb-attributes/). However, this feature wouldn't be possible through attributes alone, since it modifies `Model.save` etc. Also, a version/revision attribute is a common pattern, and I can see it being useful.

What do you think?

pynamodb/models.py
```diff
@@ -881,6 +906,40 @@ def _get_save_args(self, attributes=True, null_check=True):
...
+    def _handle_version_attribute(self, serialized_attributes, actions=None):
+        """
+        Handles modifying the request to set or increment the version attribute.
+
+        :param serialized_attributes: A dictionary mapping attribute names to serialized values.
+        :param actions: A non-empty list when performing an update, otherwise None.
+        """
+        version_condition = None
+
+        for name, attr in self.get_attributes().items():
```
garrettheel: Would it be possible to keep track of the `VersionAttribute` once upfront so that we don't need to iterate through attributes each and every time?

pynamodb/attributes.pyi
```diff
@@ -119,6 +119,11 @@ class NumberAttribute(Attribute[float]):
     @overload
     def __get__(self, instance: Any, owner: Any) -> float: ...
 
+class VersionAttribute(Attribute[float]):
+    @overload
+    def __get__(self: _A, instance: None, owner: Any) -> _A: ...
+    @overload
+    def __get__(self, instance: Any, owner: Any) -> float: ...
```
garrettheel: Did you consider coercing and typing this as an `int` instead?

garrettheel:
I'm supportive of putting this into the core given that this is such a common pattern. I took a quick first pass and it mostly looks good! I'll give it more of a look over the next few days.

Would love to see some docs with usage examples (especially exceptions) here too.

pynamodb/models.py
```diff
@@ -881,6 +906,40 @@ def _get_save_args(self, attributes=True, null_check=True):
...
+            value = getattr(self, name)
```
ikonst: In any case, you can move the `getattr` under the `if`.

ikonst:
Ahh, I somehow missed that this maps to a [DynamoDBMapper feature](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html).

ikonst:
(On a side note, I'd love if we could add [blinker](https://pythonhosted.org/blinker/) signals so that such behaviors could be implemented by plugins.)

pynamodb/models.py
ezmuller: Yes that would make more sense. I'll pull it into an attribute where I check for duplicates in init.

pynamodb/attributes.pyi
ezmuller: I didn't since this is really meant to be a _hidden_ attribute, so I just left the same interface that exists for `NumberAttribute`. I'm fine with making these changes if you think it's worth doing though.

pynamodb/models.py
```diff
@@ -881,6 +913,38 @@ def _get_save_args(self, attributes=True, null_check=True):
...
+        :param serialized_attributes: A dictionary mapping attribute names to serialized values.
+        :param actions: A non-empty list when performing an update, otherwise None.
+        """
+        if self._version_attribute_name is None:
+            return
+
+        version_attribute = self.get_attributes()[self._version_attribute_name]
+        version_attribute_value = getattr(self, self._version_attribute_name)
+
+        if not version_attribute_value:
```
ikonst: (nitpick) is it more readable to do `if version_attribute_value:` first?

pynamodb/models.py
```diff
@@ -1002,7 +1066,6 @@ def _serialize(self, attr_map=False, null_check=True):
             if isinstance(value, MapAttribute):
                 if not value.validate():
                     raise ValueError("Attribute '{}' is not correctly typed".format(attr.attr_name))
-
```
```suggestion
```

ezmuller:
The more recent update modifies the behavior of save to update the local object i.e. can be called multiple times without needing to refresh. I implemented this in Model.save since [https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html#DDB-PutItem-request-ReturnValues](PutItem) only supports ReturnValues of NONE or ALL_OLD. 

I'll also update this later tonight for Transactions.

pynamodb/models.py
ezmuller: Yes

ezmuller:
> Would love to see some docs with usage examples (especially exceptions) here too.

I was thinking I would wait to add docs and examples until people are happy with what I've implemented, so that I only have do it once.

For exceptions it doesn't look like I can differentiate whether a condition check failed because of the version attribute or some other conditional expression, so there won't be anything interesting there. I'll be sure to add examples with exceptions when I get there though.

tests/integration/test_transaction_integration.py
```diff
@@ -72,7 +85,7 @@ def connection(ddb_url):
     yield Connection(host=ddb_url)
 
 
-@pytest.fixture(scope='module', autouse=True)
+@pytest.fixture(scope='function', autouse=True)
```
ikonst: Why?

ezmuller: So that the `Foo` table will be created before and deleted after each of the test cases that I added. No special reasoning here, it was just an easy way to avoid collisions.

ikonst: Can you instead use a random (or incremental) ID for your tests, e.g. `str(uuid4())`? This way you don't need to delete tables between tests.

ezmuller: Sure

pynamodb/attributes.pyi
garrettheel: Although the version is managed for you here, I think it'll still be common to read the value for logging or other purposes, so I do think typing as an int would be helpful

docs/optimistic_locking.rst
```diff
@@ -0,0 +1,175 @@
...
+
+Optimistic Locking is a strategy for ensuring that your database writes are not overwritten by the writes of others.
+With optimistic locking, each item has an attribute that acts as a version number. If you retrieve an item from a
+table, the application records the version number of that item. You can update the item, but only if the version number
+on the server side has not changed. If there is a version mismatch, it means that someone else has modified the item
+before you did. The update attempt fails, because you have a stale version of the item. If this happens, you simply
+try again by retrieving the item and then trying to update it. Optimistic locking prevents you from accidentally
+overwriting changes that were made by others. It also prevents others from accidentally overwriting your changes.
+
+Note
```
garrettheel: nit: can you use the `warning` directive here instead? i.e

```
.. warning::
   warning goes here
```

See docs/backup_restore.rst for an example

docs/optimistic_locking.rst
```diff
@@ -0,0 +1,175 @@
...
+
+  with assert_condition_check_fails():
+      office.delete()
+
+Transactions
+------------
+
+Transactions are supported.
+
+**Note:** *local objects that were used in a successful write transaction will not have their version attribute updated
```
garrettheel: Is this something we should fix instead of including a warning?

ezmuller: Fixed, updated tests + example + docs

docs/optimistic_locking.rst
ezmuller: Switched to a warning

ezmuller:
Rebased and squashed to a single commit.

docs/optimistic_locking.rst
```diff
@@ -0,0 +1,167 @@
...
+
+  # Add another employee and persist the change.
+  office.employees.append(OfficeEmployeeMap(office_employee_id=str(uuid4()), person='lita'))
+  office.save()
+  # On subsequent save or update operations the version is also incremented locally to match the persisted value so
+  # there's no need to refresh between operations when reusing the local copy.
+  assert office.version == 2
+  assert office_out_of_date.version == 1
+
+The version checking is implemented using DynamoDB conditional write constraints. Asserting that no value exists
```
ikonst:
```suggestion
The version checking is implemented using DynamoDB conditional write constraints, asserting that no value exists
```

tests/test_model.py
```diff
@@ -2964,6 +2974,154 @@ def fake_dynamodb(*args, **kwargs):
...
+                'TableName': 'VersionedModel',
+                'ConditionExpression': '#0 = :0',
+                'ExpressionAttributeNames': {'#0': 'version'},
+                'ExpressionAttributeValues': {':0': {'N': '1'}}
+            }
+
+            deep_eq(args, params, _assert=True)
+
+    def test_version_attribute_increments_on_update(self):
+
```
ikonst: nit - stray newline

ikonst:
The squash-and-force-push workflow used to make it hard to track changes, but I just realized GitHub allows now to see the diffs between the force-push and the commit it replaced.

ezmuller:
> The squash-and-force-push workflow used to make it hard to track changes, but I just realized GitHub allows now to see the diffs between the force-push and the commit it replaced.

Still error-prone though. :D

ezmuller:
@ikonst @garrettheel Any idea when you'll have time for another pass?

pynamodb/attributes.pyi
```diff
@@ -119,6 +119,11 @@ class NumberAttribute(Attribute[float]):
     @overload
     def __get__(self, instance: Any, owner: Any) -> float: ...
 
+class VersionAttribute(Attribute[float]):
```
garrettheel:
```suggestion
class VersionAttribute(NumberAttribute):
```

docs/optimistic_locking.rst
```diff
@@ -0,0 +1,167 @@
...
+      transaction.save(Office(office_id=str(uuid4()), name="new office", employees=[justin, garrett]))
+      transaction.update(
+          office3,
+          actions=[
+              Office.name.set('birdistheword'),
+          ]
+      )
+
+  try:
+    office2.refresh()
```
garrettheel: should we `pytest.fail` on the next line to ensure this was deleted?

garrettheel:
@ezmuller docs look good, thanks for writing those! I think this is pretty much ready to go, just a couple of final small comments

docs/optimistic_locking.rst
ezmuller: Yes, I should definitely raise an exception if `DoesNotExist` isn't thrown. I think pytest.fail would be a bit out-of-place since this isn't a test, but I'll add an else block which raises an `AssertionError`.

ezmuller:
> @ezmuller docs look good, thanks for writing those! I think this is pretty much ready to go, just a couple of final small comments

Np feature wouldn't be all that useful undocumented. Thanks for taking the time to go through all the revisions @garrettheel @ikonst !

ezmuller:
Squashed

docs/optimistic_locking.rst
garrettheel: Ah, didn't even realize that this wasn't a test when I was reading it 😄 Sounds good!

PR #667, opened 2019-07-22
moving transactions up a level

hallie:
Moving the package path up a level, since `.connection` is reserved for lower-level operations


ikonst:
I'd update PR description with @garrettheel's rationale:
> As I write this, I'm also realizing that this API expects folks to import out of .connection, which we would usually reserve for lower-level operations. What do you think about bumping the package up one level for the 4.0 release? i.e from pynamodb.transactions import ...?

PR #670, opened 2019-07-23
parsing errors from transaction

hallie:
Trying to make it easier to understand which item(s) caused the transaction to be canceled. Current approach is to return list of tuples with format `({item}, 'reason')`, but may make more sense to do
```
{
    'reason1': [{}, {}],
    'reason2': [{}]
}
```
or adding a `failure_reason` key to each item in the list

### DIFF
--- a/pynamodb/exceptions.py
+++ b/pynamodb/exceptions.py
@@ -119,7 +119,7 @@ class TransactError(PynamoDBException):
         if self.cause_response_code != 'TransactionCanceledException':
             return None
         reason_list = self._get_reason_list_from_message(self.cause_response_message)
-        return [item for item in zip(transact_items, reason_list)]
+        return list(zip(transact_items, reason_list))
 
 
 class TransactWriteError(TransactError):
--- a/pynamodb/exceptions.py
+++ b/pynamodb/exceptions.py
@@ -1,8 +1,11 @@
 """
 PynamoDB exceptions
 """
-
+import re
 import botocore.exceptions
+
+
+TRANSACTION_CANCELED_REGEX = re.compile(r'\[([A-Za-z]+,?\s?)+\]')
 
 
 class PynamoDBException(Exception):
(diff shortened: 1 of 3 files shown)


pynamodb/exceptions.py
```diff
@@ -95,14 +98,38 @@ def __init__(self, table_name):
...
+        reasons = TRANSACTION_CANCELED_REGEX.search(message)
+        if not reasons:
+            return None
+        return reasons.group()[1:-1].split(', ')
+
+    def parse_cancel_reasons(self, transact_items):
+        if self.cause_response_code != 'TransactionCanceledException':
+            return None
+        reason_list = self._get_reason_list_from_message(self.cause_response_message)
+        return [item for item in zip(transact_items, reason_list)]
```
ikonst:
```suggestion
        return list(zip(transact_items, reason_list))
```

pynamodb/exceptions.py
```diff
@@ -95,14 +98,38 @@ def __init__(self, table_name):
...
+        super(TransactError, self).__init__(*args, **kwargs)
+        self._cancel_reasons = self.parse_cancel_reasons(transact_items=transact_items)
+
+    @property
+    def cancel_reasons(self):
+        return self._cancel_reasons
+
+    @staticmethod
+    def _get_reason_list_from_message(message):
+        reasons = TRANSACTION_CANCELED_REGEX.search(message)
```
ikonst: Rather confused about it. Is it a regex with a variable number of groups?

hallie: no, should just be one :0

ikonst: ahh, it's indexing on a string

ikonst:
```py
TRANSACTION_CANCELED_REGEX = re.compile(r'^.* \[(.*)\]$')
...
m = TRANSACTION_CANCELED_REGEX.match(message)
if not m:
   return None
return m.group(1).split(', ')
```

tests/test_exceptions.py
```diff
@@ -40,3 +40,44 @@ def test_get_cause_response_message__no_message():
...
+            error_response={
+                'Error': {
+                    'Code': 'TransactionCanceledException',
+                    'Message': 'Transaction cancelled, please refer cancellation reasons for specific reasons [None, ConditionalCheckFailed]'
+                }
+            },
+            operation_name='test'
+        )
+    )
+    assert error.cancel_reasons == [({'TableName': 'toot'}, 'None'), ({'TableName': 'boot'}, 'ConditionalCheckFailed')]
```
garrettheel: should `'None'` be `None` here?

pynamodb/exceptions.py
garrettheel: We could also do something like:

```ipython
In [20]: re.compile(r'\[([A-Za-z]+)(?:,\s([A-Za-z]+))*\]').search(msg).groups()
Out[20]: ('None', 'ConditionalCheckFailed')
```

ikonst: Slightly less readable than a `split` IMO, but fine too -- maybe change `[A-Za-z]+` to `\w+` since we have no reason to assume the contrary?

tests/test_exceptions.py
ikonst: It's `None` from the string generated by DynamoDB. Do you suggest we treat `'None'` specially?

hallie: yeah, probably makes the most sense to go through and replace all the `"None"`s with `None`

pynamodb/exceptions.py
garrettheel: Possibly a matter of preference, but I find a single regex to be more readable than a regex + split combined parse

pynamodb/exceptions.py
```diff
@@ -95,14 +98,38 @@ def __init__(self, table_name):
         super(TableDoesNotExist, self).__init__(msg)
 
 
-class TransactWriteError(PynamoDBException):
+class TransactError(PynamoDBException):
```
garrettheel: Don't forget to add type stubs for these!

garrettheel:
hey @hallie, did you still wanna get this in?

hallie:
made a new pr because i have no idea how to use github https://github.com/pynamodb/PynamoDB/pull/736

PR #691, opened 2019-09-12
adding some more info for ScanIndexForward

ahujamoh:


pynamodb/models.py
```diff
@@ -574,8 +574,9 @@ def query(cls,
         :param consistent_read: If True, a consistent read is performed
         :param index_name: If set, then this index is used
         :param limit: Used to limit the number of results returned
-        :param scan_index_forward: If set, then used to specify the same parameter to the DynamoDB API.
-            Controls descending or ascending results
+        :param scan_index_forward: (boolean) If set, then used to specify the same parameter to the DynamoDB API.
```
```suggestion
        :param scan_index_forward: If set, then used to specify the same parameter to the DynamoDB API.
```
to be consistent with other documentation. We already have typing in the typeshed.

ikonst: Or maybe simply:
```
        :param scan_index_forward: If False, the sort order will be descending on the range key. The default sort order is ascending.
```
I don't think saying it's "the same parameter to the DynamoDB API" is very helpful, since it can be said about most of our features, since we're just wrapping DynamoDB and trying not remain relatively transparent.

PR #711, opened 2019-10-20
Disable botocore parameter validation

garrettheel:
In moving to botocore in 4.0, we're spending unnecessary time validating parameters that we weren't previously doing. Given that pynamodb already does parameter validation, I don't think there's a ton of value in doing this so let's turn it off for performance reasons.


ikonst:
The only hesitation I have is that botocore is more of an authority on what's legal than an 'unofficial' project, but then again why does botocore need to perform client-side validation?

garrettheel:
Agreed that it's more of an authority, though we lived without it until 4.0 so I dare say we can live without it after. I'm open to making this configurable in future if someone does want to enable it

PR #717, opened 2019-10-31
Fix incorrect documentation for Transaction.update

### LINKED ISSUE #716: TransactWrite update() documentation is wrong
The documentation of TransactWrite's update() method is incorrect. It states that it is expecting model classes as arguments:

> The Update operation functions similarly to Model.update.
> 
> model_cls (required)
> hash_key (required) 
[Documentation](https://pynamodb.readthedocs.io/en/latest/transaction.html)

The update() method is actually implemented differently, it expects a model instance: 
`def update(self, model, actions, condition=None, return_values=None):`
[Source File](https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/transactions.py)

hallie:
closes #716

### DIFF
--- a/docs/transaction.rst
+++ b/docs/transaction.rst
@@ -170,7 +170,7 @@ Update
 
 The ``Update`` operation functions similarly to ``Model.update``.
 
-* ``model_cls`` (required)
+* ``model`` (required)
 * ``hash_key``  (required)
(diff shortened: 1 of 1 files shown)


hallie:
@ikonst

ikonst:
Looks good, let's update the PR subject?

Issue #714, opened 2019-10-30
create-table in ap-northeast-3 will fail for 'Unsupported input parameter BillingMode'


JiangKeshu:
I'm currently using 4.0.0 and got `Unsupported input parameter BillingMode` error when creating table in ap-northeast-3.
Here is the error stack:
```
Traceback (most recent call last):
  File "/.../lib/python3.4/site-packages/pynamodb/connection/base.py", line 629, in create_table
    data = self.dispatch(CREATE_TABLE, operation_kwargs)
  File "/.../lib/python3.4/site-packages/pynamodb/connection/base.py", line 336, in dispatch
    data = self._make_api_call(operation_name, operation_kwargs)
  File "/.../lib/python3.4/site-packages/pynamodb/connection/base.py", line 439, in _make_api_call
    raise VerboseClientError(botocore_expected_format, operation_name, verbose_properties)
pynamodb.exceptions.VerboseClientError: An error occurred (ValidationException) on request (XXX) on table (my_table_name) when calling the CreateTable operation: One or more parameter values were invalid: Unsupported input parameter BillingMode
```

My table defination is very simple:
```
class MyModel(Model):
    class Meta:
        table_name = 'my_table'
        read_capacity_units = 5
        write_capacity_units = 5
    region = UnicodeAttribute(hash_key=True, null=False)
    name = UnicodeAttribute(range_key=True, null=False)
    arn = UnicodeAttribute(null=True)
    checksum = UnicodeAttribute(null=True)
```

This was because ap-northeast-3 doesn't support on-demand thus billing_mode is not yet supported in this region. 
You can simply reproduce the error with CLI:

```
>> aws --region ap-northeast-3 dynamodb create-table --cli-input-json file:///tmp/Items.json

An error occurred (ValidationException) when calling the CreateTable operation: One or more parameter values were invalid: Unsupported input parameter BillingMode
>> aws --region ap-northeast-2 dynamodb create-table --cli-input-json file:///tmp/Items.json
{
    "TableDescription": {
        "AttributeDefinitions": [
            {
                "AttributeName": "Id",
                "AttributeType": "S"
            }
        ],
        "TableName": "Items",
        "KeySchema": [
            {
                "AttributeName": "Id",
                "KeyType": "HASH"
            }
        ],
        "TableStatus": "CREATING",
        "CreationDateTime": 1572419463.816,
        "ProvisionedThroughput": {
            "NumberOfDecreasesToday": 0,
[...]

PR #721, opened 2019-11-07
Do not pass BillingMode: ‘PROVISIONED’ to CreateTable

ikonst:
DynamoDB in some AZs (e.g. ap-northeast-3) does not support the BillingMode parameter (and any billing modes other than 'provisioned'). We'll avoid sending BillingMode if it is 'provisioned'.


ikonst:
Closes #714.

PR #723, opened 2019-11-12
use importlib instead of deprecated library imp

reginalin:
This PR uses `importlib` rather than the deprecated module `imp` for loading modules from a path. 

The error that this PR addresses is a pytest DeprecationWarning:
`
/lib/python3.7/site-packages/pynamodb/settings.py:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp
`

### DIFF
--- a/pynamodb/_compat.py
+++ b/pynamodb/_compat.py
@@ -7,4 +7,7 @@ else:
     from inspect import getfullargspec
     from importlib.machinery import SourceFileLoader
 
-__all__ = ('getfullargspec',)
+load_source_compat = lambda name, path: load_source(name, path) if six.PY2 else SourceFileLoader(name, path).load_module()
+
+__all__ = ('getfullargspec', 'load_source')
+
--- a/pynamodb/_compat.py
+++ b/pynamodb/_compat.py
@@ -11,9 +11,9 @@ def load_module(name, path):
     if sys.version_info >= (3, 3):
         from imp import load_source
         from importlib.machinery import SourceFileLoader
+        return SourceFileLoader(name, path).load_module()
     else: 
         from imp import load_source
         return load_source(name, path)
-    return SourceFileLoader(name, path).load_module()
 
 __all__ = ('getfullargspec', 'load_module')
--- a/pynamodb/_compat.py
+++ b/pynamodb/_compat.py
@@ -7,11 +7,10 @@ else:
     from inspect import getfullargspec
     from importlib.machinery import SourceFileLoader
 
-def load_source_compat(name, path):
-    """Load module using the Python version compatible function. """
+def load_module(name, path):
+    """Load module using the Python version compatible function."""
     if six.PY2:
         return load_source(name, path)
     return SourceFileLoader(name, path).load_module()
 
-__all__ = ('getfullargspec', 'load_source_compat')
-
+__all__ = ('getfullargspec', 'load_module')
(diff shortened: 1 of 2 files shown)


pynamodb/settings.py
```diff
@@ -1,8 +1,8 @@
-import imp
 import logging
 import os
 import warnings
 from os import getenv
+from importlib.machinery import SourceFileLoader
```
ikonst: We still maintain Python 2.7 compatibility in PynamoDB 4.x. Maybe add something to pynamodb.compat?

reginalin: Thanks! I've updated the _compat module.

pynamodb/_compat.py
```diff
@@ -2,7 +2,12 @@
 
 if six.PY2:
     from inspect import getargspec as getfullargspec
+    from imp import load_source
 else:
     from inspect import getfullargspec
+    from importlib.machinery import SourceFileLoader
+
+load_source_compat = lambda name, path: load_source(name, path) if six.PY2 else SourceFileLoader(name, path).load_module()
```
ikonst: nit - maybe just a function? :)

reginalin: haha gotcha

pynamodb/_compat.py
```diff
@@ -2,7 +2,16 @@
 
 if six.PY2:
     from inspect import getargspec as getfullargspec
+    from imp import load_source
 else:
     from inspect import getfullargspec
+    from importlib.machinery import SourceFileLoader
+
+def load_source_compat(name, path):
+    """Load module using the Python version compatible function. """
```
ikonst:
```suggestion
    """Load module using the Python version compatible function."""
```

Also, maybe name it `load_module`, similar to how preexisting `getfullargspec` isn't exported as `getfullargspec_compat`?

ikonst:
cc @garrettheel `SourceFileLoader ` is "New in version 3.3". Do we have a cut-off for Python 3.x versions?

garrettheel:
@ikonst I haven't thought about that too much yet, but I'd probably look at what the major distros have packaged and what specific features we need. I'd guess we could limit to >=3.4 or so for the next major pynamodb version, but we could also go higher if we need to for asyncio

pynamodb/_compat.py
```diff
@@ -2,7 +2,15 @@
...
     from inspect import getargspec as getfullargspec
+    from imp import load_source
 else:
     from inspect import getfullargspec
+    from importlib.machinery import SourceFileLoader
 
-__all__ = ('getfullargspec',)
+def load_module(name, path):
+    """Load module using the Python version compatible function."""
+    if six.PY2:
```
ikonst: Consider
```python
def load_module(name, path):
    """Load module using the Python version compatible function."""
    if sys.version_info >= (3, 3):
       from imp import load_source
       from importlib.machinery import SourceFileLoader
    else:
       from imp import load_source
       return load_source(name, path)
```

ikonst:
oops would probably put the `return` after the import :)

pynamodb/_compat.py
```diff
@@ -1,8 +1,19 @@
...
 if six.PY2:
     from inspect import getargspec as getfullargspec
 else:
     from inspect import getfullargspec
 
-__all__ = ('getfullargspec',)
+def load_module(name, path):
+    """Load module using the Python version compatible function."""
+    if sys.version_info >= (3, 3):
+        from imp import load_source
```
asottile: this still triggers the warning and is unused

Issue #218, opened 2017-01-11
ProvisionedThroughputExceeded does not trigger exponential backoff


JohnEmhoff:
I'm getting lots of `ProvisionThroughputExceeded` errors and they're surfacing immediately instead of after retries with exponential backoff. It looks like it's because of [this check](https://github.com/jlafon/PynamoDB/blob/4a808b7872f390b9dcce4724bdae33837391047c/pynamodb/connection/base.py#L335): response codes < 500 aren't retried. According to the [AWS docs](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html), this error is in the 400 range and should be retried.

I'm happy to supply a PR if you agree this is in fact a bug. Otherwise, it seems client code has to catch this exception after doing a `batch.save()` and retry with `batch.commit()`?

JohnEmhoff:
Actually it looks like there's no decent way to recover from this because `batch.commit()` clears out its list of pending requests.

PR #728, opened 2019-11-20
Implement exponential backoff for batch writes

edholland:
There is currently a problem with exponential backoff for batch write requests.

This is because a batch that exceeds provisioned capacity does not return an error code that can be caught in https://github.com/pynamodb/PynamoDB/blob/28b3a4589fdc00868e902864d1ab6552f782f545/pynamodb/connection/base.py#L441-L466 code block, but instead returns a success code with an object describing the failing records.

This PR introduces sleep time + a max retry limit to the existing batch write loop.
Additionally it saves the state of failed records when failing so the client can retry these records if needed. This resolves some comments in #218

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -61,6 +61,7 @@ from pynamodb.signals import pre_dynamodb_send, post_dynamodb_send
 from pynamodb.types import HASH, RANGE
 
 BOTOCORE_EXCEPTIONS = (BotoCoreError, ClientError)
+RATE_LIMITING_EXCEPTIONS = ['ProvisionedThroughputExceededException', 'ThrottlingException']
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
@@ -442,7 +443,7 @@ class Connection(object):
                     if is_last_attempt_for_exceptions:
                         log.debug('Reached the maximum number of retry attempts: %s', attempt_number)
                         raise
-                    elif status_code < 500 and code != 'ProvisionedThroughputExceededException':
+                    elif status_code < 500 and code not in RATE_LIMITING_EXCEPTIONS:
                         # We don't retry on a ConditionalCheckFailedException or other 4xx (except for
                         # throughput related errors) because we assume they will fail in perpetuity.
                         # Retrying when there is already contention could cause other problems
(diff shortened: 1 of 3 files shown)


edholland:
I think this also resolves #598 and #594, as their use-cases seem to be handled transparently with this change

pynamodb/connection/base.py
```diff
@@ -61,6 +61,7 @@
 from pynamodb.types import HASH, RANGE
 
 BOTOCORE_EXCEPTIONS = (BotoCoreError, ClientError)
+RATE_LIMITING_EXCEPTIONS = ['ProvisionedThroughputExceededException', 'ThrottlingException']
```
```suggestion
RATE_LIMITING_ERROR_CODES = ['ProvisionedThroughputExceededException', 'ThrottlingException']
```

edholland:
@ikonst: I have implemented the requested changed & added a test case as the coverage dipped when i applied your changes

pynamodb/models.py
```diff
@@ -130,16 +132,24 @@ def commit(self):
...
+                raise PutError("Failed to batch write items: max_retry_attempts exceeded")
             put_items = []
             delete_items = []
             for item in unprocessed_items:
                 if PUT_REQUEST in item:
                     put_items.append(item.get(PUT_REQUEST).get(ITEM))
                 elif DELETE_REQUEST in item:
                     delete_items.append(item.get(DELETE_REQUEST).get(KEY))
-            log.info("Resending %s unprocessed keys for batch operation", len(unprocessed_items))
+            log.info("Resending %s unprocessed keys for batch operation after %d seconds sleep",
```
```suggestion
            log.info("Resending %d unprocessed keys for batch operation after %d seconds sleep",
```

ikonst: not like it matters :)

edholland:
@ikonst: I'm happy for you to merge + release as part of your normal cycle

ikonst:
What I meant is to describe the change in the "in progress" release notes, for when we release.

ikonst:
To clarify, the previous behavior was a potential endless loop and you made it retry a limited number of times, right?

edholland:
@ikonst: Yes that is correct. The old code failed to respect configured backoff policy, this PR changes it so it will no longer retry in infinite tight loop but instead apply the same backoff/retry policy as the rest of the api

ikonst:
Thanks, it'll help me compose a commit message and a change log entry.

PR #736, opened 2019-12-25
Transact error

hallie:
Trying to make it easier to understand which item(s) caused the transaction to be canceled. Current approach is to return list of tuples with format ({item}, 'reason'), but may make more sense to do
```python
{
    'reason1': [{}, {}],
    'reason2': [{}]
}
```
or adding a failure_reason key to each item in the list

formerly [this](https://github.com/pynamodb/PynamoDB/pull/670) but i have no idea how to use github


ikonst:
Me and @jpr71 just stumbled on another reason for this.

PR #754, opened 2020-02-13
Remove elements from a ListAttribute using a list of indexes

dotpmrcunha:
Based on the AWS documents https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.REMOVE

It is possible to delete items from a list inside a ListAttribute using:

`--update-expression "REMOVE RelatedItems[1], RelatedItems[2]"`

This way I extended the remove in the ListAttribute and added the indexes argument, if there is a valid integer List it will remove the selected elements.

Usage:

`list_attribute.remove(indexes=[0, 10])`


garrettheel:
This looks good overall! Would you mind adding some docs to `docs/updates.rst` with a usage example? We should be good to go after that.

dotpmrcunha:
Hi, added the document and the example. Thanks

garrettheel:
Thanks @dotpmrcunha. I fixed the travis build reporting and it now shows a type error. I'd suggest changing the type signature of `Attribute.remove` to be compatible

pynamodb/attributes.py
```diff
@@ -961,6 +962,13 @@ def __init__(self, hash_key=False, range_key=False, null=None, default=None, att
                 raise ValueError("'of' must be subclass of MapAttribute")
             self.element_type = of
 
+    def remove(self, indexes=None):
```
garrettheel: Apologies for not suggesting earlier, but having thought about this a bit it's probably better to have a separate `remove_indexes` function on `ListAttribute` rather than overloading the existing `remove`. What do you think?

dotpmrcunha: Well... while adding the indexes attribute, I was also thinking in same, separate the methods. I will do that then, it makes more sense.

dotpmrcunha: And also, instead of using:
```
class ListAttribute:

    def remove_indexes(self, *indexes):
        ....

Thread.notes.remove_indexes([1,2,...])
```

I add like this:
```
class ListAttribute:

    def remove_indexes(self, *indexes):
        ....

Thread.notes.remove_indexes(1, 2, ...)
```
What do you think?

garrettheel: Makes sense to use varargs 👍

docs/updates.rst
```diff
@@ -44,4 +44,5 @@ Any value provided will be serialized using the serializer defined for that attr
     `attr_or_value_1` \- `attr_or_value_2`, `attr_or_value_1` \- `attr_or_value_2`, 5 - Thread.views
     "list_append( `attr` , `value` )", append( `value` ), Thread.notes.append(['my last note'])
     "list_append( `value` , `attr` )", prepend( `value` ), Thread.notes.prepend(['my first note'])
+    "REMOVE list[index1], list[index1]", "remove_indexes(`index1`, `index2`)", "Thread.notes.remove_indexes(0, 1)"
```
garrettheel:
```suggestion
    "REMOVE list[index1], list[index2]", "remove_indexes(`index1`, `index2`)", "Thread.notes.remove_indexes(0, 1)"
```

pynamodb/attributes.py
```diff
@@ -173,7 +173,8 @@ def prepend(self, other):
     def set(self, value):
         return Path(self).set(value)
 
-    def remove(self):
+    def remove(self, indexes=None):
+        # Indexes argument used in the ListAttribute class
```
garrettheel:
```suggestion
    def remove(self):
```

docs/updates.rst
dotpmrcunha: Fixed

pynamodb/attributes.py
dotpmrcunha: Fixed

PR #765, opened 2020-04-08
Add exception descriptions to the Docs

meawoppl:
I had to do quite a bit of guess and check to figure these all out. It would be helpful if they were in the docs.

Thanks!


meawoppl:
Ty!

meawoppl:
When do docs get pushed to the site?

ikonst:
Once we cut a new release, "readthedocs.io" will be updated with the docs.

PR #767, opened 2020-04-15
Add Model.record_exists helper method

dan-kez:
I recently added this helper method in my own project and thought it may be useful upstream. In short, this adds a short form utility to check if a record exists by looking up the hash_key and optional range_key.

I'm less familiar with how the tests are set up here and attempted to add both an integration and unit test. If there is something I could do better here please let me know!

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -676,6 +676,26 @@ class Model(AttributeContainer):
         )
 
     @classmethod
+    def record_exists(cls, hash_key, range_key=None):
+        """
+        Checks if the given hash_key and optional range_key pair exists in the table.
+
+        This will return true if at least one record matching hash_key and range_key exists
+
+        :param hash_key: The hash key to look up
+        :param range_key: If set, include in lookup
+        """
+        try:
+            if range_key is None:
+                cls.query(hash_key).next()
+            else:
+                cls.query(hash_key, cls._range_key_attribute() == range_key).next()
+        except StopIteration:
+            return False
+
+        return True
+
+    @classmethod
     def exists(cls):
         """
         Returns True if this table exists, False otherwise
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -687,9 +687,9 @@ class Model(AttributeContainer):
         """
         try:
             if range_key is None:
-                cls.query(hash_key).next()
+                next(cls.query(hash_key, page_size=1))
             else:
-                cls.query(hash_key, cls._range_key_attribute() == range_key).next()
+                next(cls.query(hash_key, cls._range_key_attribute() == range_key, page_size=1))
         except StopIteration:
             return False
 
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -676,20 +676,27 @@ class Model(AttributeContainer):
         )
 
     @classmethod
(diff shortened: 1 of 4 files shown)


pynamodb/models.py
```diff
@@ -675,6 +675,26 @@ def scan(cls,
             rate_limit=rate_limit,
         )
 
+    @classmethod
+    def record_exists(cls, hash_key, range_key=None):
```
ikonst, line 679: If we add this method, we should probably update models.pyi with its type information.

pynamodb/models.py
```diff
@@ -675,6 +675,26 @@ def scan(cls,
...
+        This will return true if at least one record matching hash_key and range_key exists
+
+        :param hash_key: The hash key to look up
+        :param range_key: If set, include in lookup
+        """
+        try:
+            if range_key is None:
+                cls.query(hash_key).next()
+            else:
+                cls.query(hash_key, cls._range_key_attribute() == range_key).next()
```
ikonst, line 692:
```suggestion
            if range_key is None:
                next(cls.query(hash_key, page_size=1))
            else:
                next(cls.query(hash_key, cls._range_key_attribute() == range_key, page_size=1))
```

1. I recall restricting `page_size` leads to spending fewer RCUs. @garrettheel?
2. Total nit, but maybe using `next` built-in is more idiomatic?

pynamodb/models.py
dan-kez: Sounds good!

pynamodb/models.py
dan-kez: Sounds good to me, I applied your suggestion.

dan-kez:
@ikonst I just addressed your comments and added an additional kwarg, `range_key` for, what I think is the most common use case of direct equality comparisons. 

In the doc string I indicated that `range_key` would have priority over `range_key_condition` if they are both present. If you would rather me throw an error instead of allowing this let me know. I'm happy to throw a `ValueError` or something

pynamodb/models.py
```diff
@@ -675,6 +675,33 @@ def scan(cls,
             rate_limit=rate_limit,
         )
 
+    @classmethod
+    def record_exists(cls, hash_key, range_key_condition=None, range_key=None):
```
ikonst, line 679: Maybe a way to avoid the "what if both arguments are passed" problem would be to have just one argument and branch on `isinstance(range_key, Condition)`?

dan-kez: So this does match your current argument structure and naming where the second key is the condition expression (see `query`)

I am happy to do that branching mechanism but was concerned that it would be a divergent API compared to query and others (e.g. `MyModel.query('hash_key', 'range_key'` will not work today).

If you'd like me still reorder, rename, or remove something let me know!

ikonst: Yeah, it's `query(hash_key, range_key_condition=None)`. I guess we don't have a `query(hash_key, range_key)` shortcut form, but we want to have a shortcut form here?

dan-kez: I'm happy to remove it if you'd rather not have it. In all my use cases I always want to check if a single record exists by looking up both the hash_key and range_key directly. The short form is nice since since i don't need to write out the whole range condition.

All that said, I'm happy to just drop the short form

PR #769, opened 2020-04-21
Make (GS,LS)Index a Generic in runtime

ikonst:
Following 5426ed9081e91316b214d6c10b447cfa1da8b0af, users found themselves having to write some rather unergonomic code to reconcile Python's and a type-checker's perspectives:
```python
if TYPE_CHECKING:
    _MyModelGSI = GlobalSecondaryIndex["MyModel"]
else:
    _MyModelGSI = GlobalSecondaryIndex

class MyModelUserIdGSI(_MyModelGSI):
    """A GSI for looking up by user ID."""
   ...

class MyModelEmailGSI(_MyModelGSI):
    """A GSI for looking up by email."""
   ...
```

This addresses it by allowing users to rewrite this as:
```python
class MyModelUserIdGSI(GlobalSecondaryIndex["MyModel"]):
    """A GSI for looking up by user ID."""
   ...

class MyModelEmailGSI(GlobalSecondaryIndex["MyModel"]):
    """A GSI for looking up by email."""
   ...
```

Since we maintain Python 2 compatibility, we cannot assume the typing module is available so we're making a mock Generic class instead.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+v4.3.2
+----------
+
+* Fix discrepancy between runtime and type-checker's perspective of Index and derived types
+
 v4.3.1
 ----------
 
--- a/pynamodb/__init__.py
+++ b/pynamodb/__init__.py
@@ -7,4 +7,4 @@ A simple abstraction over DynamoDB
 """
 __author__ = 'Jharrod LaFon'
 __license__ = 'MIT'
-__version__ = '4.3.1'
+__version__ = '4.3.2'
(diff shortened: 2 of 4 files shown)


ikonst:
ptal @mxr @garrettheel

docs/release_notes.rst
```diff
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+v4.3.2
+----------
+
+* Fix discrepancy between runtime and type-checker's perspective of Index and derived types
```
mxr, line 7: Will this be edited later to include all changes between 4.3.1 and 4.3.2?

ikonst: Missed that one!

PR #775, opened 2020-05-01
feat(Base): Add region when scan exception occures

acolombier:
Hi PynamoDB team!

First, thanks for your great work!

This PR comes after a painful debug session to catch a "table not found" issue. In the end, it turned out that the region had been change in our config file.

A few day ago, the issue occurred again, and even if this time, it took me only a few minutes to remember this issue, I thought it would be great to display the region which is currently used for scanning, and then help people catching a potential region misconfiguration, especially since the region is defaulted in `settings.py`. 

Hope you like the idea,
Thanks

### DIFF
--- a/pynamodb/__init__.py
+++ b/pynamodb/__init__.py
@@ -7,4 +7,4 @@ A simple abstraction over DynamoDB
 """
 __author__ = 'Jharrod LaFon'
 __license__ = 'MIT'
-__version__ = '4.3.2'
+__version__ = '4.3.3'
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -1182,7 +1182,7 @@ class Connection(object):
         try:
             return self.dispatch(SCAN, operation_kwargs)
         except BOTOCORE_EXCEPTIONS as e:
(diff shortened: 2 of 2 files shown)


ikonst:
I'm not sure what this exception normally says (does it have the table's name?) but I think this change would be overly specific to the problem you faced -- "fighting the last war", so to speak.

acolombier:
Hi @ikonst ,

Sorry, I should have stated that first. 

Basically the error message from `boto3` looks like the following:

```
An error occurred (ResourceNotFoundException) when calling the <OperationName> operation: Requested resource not found: Table: <TableName> not found
```

`boto3` wouldn't display the region because it is a mandatory setting (if you don't set it up in the code, it is expecting it to find it in your profile definition). Here is what happen if `boto3` cannot resolve an explicitly set region:

```
botocore.exceptions.NoRegionError: You must specify a region.
```

However, `PynamoDB` is "breaking" this safeguard by explicitly defaulting the region to `us-east-1` and thus, brings to confusion when you are used to AWS services(/`boto3`) requirements.

If you wish, the aim of this PR is to bring the safeguard in a non-destructive way. :)

I hope that makes it more clear.

Thanks.

PR #776, opened 2020-05-01
Clarification in conditional docs

szerfas1:
Ran around looking for clarification this morning and figured I'd save others the time.

### DIFF
--- a/docs/conditional.rst
+++ b/docs/conditional.rst
@@ -100,7 +100,7 @@ This example will update a `Thread` item, if the `views` attribute is less than 
 .. code-block:: python
 
     thread_item.update(condition=(Thread.views < 5) | (Thread.views > 10))
-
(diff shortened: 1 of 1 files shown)


docs/conditional.rst
```diff
@@ -100,7 +100,7 @@ This example will update a `Thread` item, if the `views` attribute is less than
 .. code-block:: python
 
     thread_item.update(condition=(Thread.views < 5) | (Thread.views > 10))
-
+Note that here capital-T `Thread` refers to the record currently saved in the DB with the same unique ID as the Thread model instance `thread_item`, rather than the Thread model class.
```
ikonst, line 103: Capital-T `Thread` here, as in `Thread.views < 5` is the Thread class, from whose attribute you derive a query condition. Or am I misunderstanding something?

PR #786, opened 2020-05-17
Remove py2 support

garrettheel:
This is the point of no return for the 5.0 release

TODO:
 - [x] Still more # type: ignores than I would like. Look into these further


pynamodb/_compat.py
```diff
@@ -1,10 +1,6 @@
...
 import sys
 
-if six.PY2:
-    from inspect import getargspec as getfullargspec
-else:
-    from inspect import getfullargspec
+from inspect import getfullargspec
 
 
 def load_module(name, path):
```
ikonst, line 6: This is not needed for `>= 3.5`

pynamodb/connection/table.py
```diff
@@ -41,25 +48,27 @@ def __init__(self,
                                                     aws_secret_access_key,
                                                     aws_session_token)
 
-    def get_meta_table(self, refresh=False):
+    def get_meta_table(self, refresh: bool = False):
```
ikonst, line 51: `->`?

pynamodb/exceptions.py
```diff
@@ -2,24 +2,26 @@
...
 class PynamoDBException(Exception):
     """
     A common exception class
     """
-    def __init__(self, msg=None, cause=None):
-        self.msg = msg or self.msg
+    def __init__(self, msg: Optional[str] = None, cause: Optional[Exception] = None):
+        self.msg = msg
         self.cause = cause
         super(PynamoDBException, self).__init__(self.msg)
```
ikonst, line 17: Let's modernize super syntax when we burn the bridges

pynamodb/exceptions.py
```diff
@@ -2,24 +2,26 @@
...
 import botocore.exceptions
 
 
 class PynamoDBException(Exception):
     """
     A common exception class
     """
-    def __init__(self, msg=None, cause=None):
-        self.msg = msg or self.msg
+    def __init__(self, msg: Optional[str] = None, cause: Optional[Exception] = None):
```
ikonst, line 14: May as well `-> None` for clarity

pynamodb/indexes.py
```diff
@@ -32,26 +41,28 @@ def __init__(cls, name, bases, attrs):
                         attr_obj.attr_name = attr_name
 
 
-class Index(with_metaclass(IndexMeta)):
+class Index(Generic[_M], metaclass=IndexMeta):
     """
     Base class for secondary indexes
     """
-    Meta = None
+    Meta = None  # type: Any
```
ikonst, line 48: Has to be a comment?

pynamodb/exceptions.py
garrettheel: Yep, I haven't run the tools over this yet

pynamodb/_compat.py
```diff
@@ -13,21 +9,4 @@ def __getitem__(self, item):
...
-        # https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
-        loader = SourceFileLoader(name, path)
-        return loader.load_module()  # type: ignore
-    else:
-        from imp import load_source
-        return load_source(name, path)
-
-
-__all__ = ('getfullargspec', 'FakeGenericMeta', 'load_module')
+__all__ = ('getfullargspec', 'FakeGenericMeta')
```
ikonst, line 12: probably don't need to import/export 'getfullargspec' anymore?

pynamodb/_compat.py
```diff
@@ -1,10 +1,6 @@
...
 import sys
 
-if six.PY2:
-    from inspect import getargspec as getfullargspec
-else:
-    from inspect import getfullargspec
+from inspect import getfullargspec
 
 
 class FakeGenericMeta(type):
```
ikonst, line 6: From Python 3.7 we can use real `typing.Generic`. Classes derived from `Generic` don't require the type params, so existing code will be unaffected.
```
>>> from typing import *
>>> T = TypeVar('T')
>>> class MyBase(Generic[T]): pass
...
>>> class ThisIsOk(MyBase[int]): pass
...
>>> class ThisIsAlsoOk(MyBase): pass
...
```
However, IIRC Python 3.6's `Generic` doesn't like coexisting with other meta classes :( In fact, if following this modernization we don't have .pyi files anymore, we might need to branch out with an `if TYPE_CHECKING` while defining class Index.

pynamodb/models.py
```diff
@@ -193,8 +192,6 @@ class MetaModel(AttributeContainerMeta):
     billing_mode: Optional[str]
     stream_view_type: Optional[str]
 
-    _version_attribute_name: str
```
ikonst: This was plain wrong. It's a property of the Model type (and it's indeed defined there too), not the MetaModel type.

pynamodb/indexes.py
```diff
@@ -22,15 +23,15 @@
...
+class IndexMeta(GenericMeta):
     """
     Index meta class
 
     This class is here to allow for an index `Meta` class
     that contains the index settings
     """
     def __init__(cls, name, bases, attrs, *args, **kwargs):
-        super().__init__(name, bases, attrs, *args, **kwargs)
+        super().__init__(name, bases, attrs, *args, **kwargs)  # type: ignore
```
ikonst, line 34: No way to make mypy happier here: its definition of GenericMeta is incorrect as it doesn't accept `tvars` etc.

ikonst:
There's a functional change, in that one of the attributes in a model now must be have `hash_key=True` (except for the `Model` type itself which is excluded from the metaclass' doings).

ikonst:
Actually, would it break the ability to have "abstract" models?

garrettheel:
@ikonst I think I'm happy to merge this and iterate from there, what do you think?

Issue #787, opened 2020-05-21
Stubs for indexes range_key_condition: Argument 2 to query of Index has incompatible type bool expected Condition  [arg-type]


Luis-Palacios:
Hello,

I'm using vscode and mypy and when I try to use query from a Index or Model I get an error from mypy

```python
UserModel.email_index.query('foo@example.com', UserModel.tenant_id == 'tenantId')
```
The message is
```
Argument 2 to "query" of "Index" has incompatible type "bool"; expected "Condition"  [arg-type]
```

And it makes sense since this would return `UserModel.tenant_id == 'tenantid'` a bool value not a Condition instance which is what the stubs define

```python
def query(
        cls,
        hash_key,
        range_key_condition: Optional[Condition] = ...,
        filter_condition: Optional[Condition] = ...,  <- Expect condition not a bool?
        scan_index_forward: Optional[Any] = ...,
        consistent_read: Optional[bool] = ...,
        limit: Optional[int] = ...,
        last_evaluated_key: Optional[Dict[Text, Dict[Text, Any]]] = ...,
        attributes_to_get: Optional[Any] = ...,
        page_size: Optional[int] = ...,
        rate_limit: Optional[float] = ...,
    ) -> ResultIterator[_M]: ...
```

If it helps this is my Index and Model Definition

```python
class EmailIndex(GlobalSecondaryIndex):
    class Meta:
        name = 'index_email'
        projection = AllProjection()

    email = UnicodeAttribute(hash_key=True)
    tenant_id = UnicodeAttribute(range_key=True, attr_name='sort_key_id')


class UserModel(Model):
    user_id = UnicodeAttribute(hash_key=True, attr_name='part_key_id')
    tenant_id = UnicodeAttribute(range_key=True, attr_name='sort_key_id')
    email = UnicodeAttribute()

    email_index = EmailIndex()
```

Am I passing the range condition in the wrong way? or the stubs have issues?

Luis-Palacios:
I can clear the error if I update the stubs

FROM:
```python
range_key_condition: Optional[Condition]
```

```python
range_key_condition: Union[Optional[Condition], Optional[bool]]
```

So that it would take A Condition or a Bool but I'm not sure if this is a good fix since I'm still a noob with typings on python

Luis-Palacios:
Anyways I ended up creating #788 hopefully is the right approach or at least might serve as starting point

ikonst:
[...]

PR #788, opened 2020-05-21
Update condition type to handle Optional bools

Luis-Palacios:
Deals with #787 

Currently when calling any method that uses Condition as the type for one of it's argument eg: range_key_condition in query method mypy will complain if you pass something like `Model.field == 'somevalue'` or any other comparison that directly returns bool such as <, >, etc.. 

It is only happy if you pass something like `Model.field.startswith('somevalue')`  since that actually return an Condition instance. 

Given that the actual function actually handles ==, <, > This PR augment the type for conditions to handle both condition and bools
Update condition type to allow bool


jpinner-lyft:
This isn't quite right -- these APIs do not take booleans.

The model comparisons (<, >, etc.) return Condition expressions by operator overloading:
https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/expressions/operand.py#L51-L79

My suspicion is that mypy does understand the overloading.

Luis-Palacios:
Haa I see, I'll check on mypy side if there a way to let it know about operator overloading

Luis-Palacios:
So browsing trough mypy issues I realized they expect for `__eq__` to always return bool see: https://github.com/python/mypy/issues/6710#issuecomment-485580032

So I can't think of a good solution to keep `Model.prop == 'value'` as a possible argument while keeping mypy happy.

Only workarounds I can see are:
1. Use Any like so: `condition: Any = Model.prop == request.user.tenant_id` and then pass condition as argument, I have checked this work
2. Use cast if you want to keep type safety `condition = cast(Condition, Model.prop == 'somevalue')`
3. Accept my PR, but this would actually be tricky since is not really a boolean what the API uses

ikonst:
@Luis-Palacios We happen to have a test for that:
https://github.com/pynamodb/PynamoDB/blob/c414d108e4c853d6ce0dec19780406a90bdefb87/tests/test_mypy.py#L32-L33

Does this test happen to miss something?

Luis-Palacios:
It's on my side see #787 for more details.

We can forget about this

PR #799, opened 2020-06-16
Add the ability to abort transactions

th0114nd:
Currently when we want to abort a transaction, we raise an exception that we catch immediately outside of the transaction context:
```
try:
  with TransactWrite(...) as txn:
    if should_skip_txn():
      raise TransactionAbort()
except TransactionAbort:
  pass
```
instead, adding an `abort` method makes this clearer:
```
with TransactWrite(...) as txn:
  if should_skip_txn():
    txn.abort()
```

### DIFF
--- a/pynamodb/transactions.py
+++ b/pynamodb/transactions.py
@@ -18,15 +18,22 @@ class Transaction:
     def __init__(self, connection: Connection, return_consumed_capacity: Optional[str] = None) -> None:
         self._connection = connection
         self._return_consumed_capacity = return_consumed_capacity
+        self._aborted = False
 
     def _commit(self):
         raise NotImplementedError()
+
+    def abort(self):
+        """
+        Cancels the current transaction.
+        """
+        self._aborted = True
 
     def __enter__(self) -> 'Transaction':
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        if exc_type is None and exc_val is None and exc_tb is None:
+        if (not self._aborted) and exc_type is None and exc_val is None and exc_tb is None:
             self._commit()
 
 
--- a/pynamodb/transactions.py
+++ b/pynamodb/transactions.py
@@ -25,7 +25,7 @@ class Transaction:
 
     def abort(self):
         """
-        Cancels the current transaction.
+        Aborts the current transaction.
         """
         self._aborted = True
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
-pytest
+pytest>=4.6
 pytest-env
 pytest-mock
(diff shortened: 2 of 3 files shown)


th0114nd:
```
0.51s$ pytest --cov-report term-missing --cov=pynamodb tests
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/pluggy/manager.py", line 267, in load_setuptools_entrypoints
    plugin = ep.load()
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2410, in load
    self.require(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2433, in require
    items = working_set.resolve(reqs, env, installer, extras=self.extras)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.VersionConflict: (pytest 4.3.1 (/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages), Requirement.parse('pytest>=4.6'))
```
`pytest>=4.6` is required to work with the latest `pytest-cov` that was released on 6-12:
https://github.com/pytest-dev/pytest-cov/blob/master/setup.py#L125

th0114nd:
ptal @garrettheel

pynamodb/transactions.py
```diff
@@ -18,15 +18,22 @@ class Transaction:
...
         self._connection = connection
         self._return_consumed_capacity = return_consumed_capacity
+        self._aborted = False
 
     def _commit(self):
         raise NotImplementedError()
 
+    def abort(self):
+        """
+        Cancels the current transaction.
```
ikonst, line 28:
```suggestion
        Aborts the current transaction.
```

ikonst:
@hallie what do you think?

th0114nd:
It turned out that on our side we could work around this more cleanly than catching the exception, so this is not a blocker for us. I do think that it's an improvement to the API, as I had to read the source in order to figure out how to cancel a commit

hallie:
@th0114nd What's an example of a use case here? Why wouldn't you do the `if should_skip_txn()` check outside of the transaction context?
```
if should_skip_txn():
    return
with TransactWrite(...) as txn:
    # stuff
```

th0114nd:
>  Why wouldn't you do the if should_skip_txn() check outside of the transaction context?

That's actually what we ended up doing, and I'm guessing that it can always be done if no information is returned from the TX before the commit

PR #803, opened 2020-06-28
Remove Model.load/dump

ikonst:
I think I called 'my_model_instance.dumps()` one time too many.


garrettheel:
Should we provide guidance for an alternative? `._get_json()` is private, so folks would not be able to maintain this functionality particularly easily right now

ikonst:
I'll punt on this until we can offer an alternative to people who might've been using this functionality for testing or small-scale backups (hope no one is using this "toy backup" for anything serious).

PR #806, opened 2020-06-30
Add usage example for ListAttribute

### LINKED ISSUE #434: how do you append to a list of map attributes?
http://pynamodb.readthedocs.io/en/latest/attributes.html#list-attributes

My guess would be like this, but it does not work?
```
emp5 = OfficeEmployeeMap(
    office_employee_id=555,
    person='derek'
)

Office(
    office_id=3
).update(
    actions=[
        Office.employees.append([emp5])
    ]
)
```

ValueError: unsupported action type: '_ListAppend'"

yelizariev:
This closes #434

### DIFF
--- a/examples/model.py
+++ b/examples/model.py
@@ -29,6 +29,8 @@ class Thread(Model):
     answered = NumberAttribute(default=0)
     tags = UnicodeSetAttribute()
     last_post_datetime = UTCDateTimeAttribute(null=True)
+    notes = ListAttribute(default=list)
+
 
 # Delete the table
 # print(Thread.delete_table())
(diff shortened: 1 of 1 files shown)


examples/model.py
```diff
@@ -193,6 +195,13 @@ class Meta:
     Thread.tags.remove()
 ]))
 
+# Update list attribute
+print(thread_item.update(actions=[
+    Thread.notes.set(
```
ikonst, line 200: Is this really how it works? You set an append action?

yelizariev: Yes. You cannot use append alone. 

Here is an example in raw format:

    "UpdateExpression" : "SET #attrName = list_append(#attrName, :attrValue)",

From https://stackoverflow.com/questions/33278475/dynamodb-set-list-append-not-working-using-aws-sdk

Also, check #434

ikonst: Thanks!

PR #815, opened 2020-07-29
More context in exception on attr deserialization failure

jonathantan:
When deserialization error occurs in `from_raw_data`, it's unclear which attribute it is failing on. Add attribute name in exception

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -553,8 +553,8 @@ class Model(AttributeContainer, metaclass=MetaModel):
             if attr:
                 try:
                     attributes[attr_name] = attr.deserialize(attr.get_value(value))  # type: ignore
-                except TypeError:
-                    raise TypeError(f'attribute: {attr_name}')
+                except TypeError as e:
+                    raise ValueError(f'attribute: {attr_name}') from e
         return cls(_user_instantiated=False, **attributes)
 
     @classmethod
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -11,7 +11,7 @@ from typing import Any, Dict, Generic, Iterable, Iterator, List, Optional, Seque
     Tuple, Union, cast
 
 from pynamodb.expressions.update import Action
-from pynamodb.exceptions import DoesNotExist, TableDoesNotExist, TableError, InvalidStateError, PutError
+from pynamodb.exceptions import DoesNotExist, TableDoesNotExist, TableError, InvalidStateError, PutError, AttributeDeserializationError
 from pynamodb.attributes import (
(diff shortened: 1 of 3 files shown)


pynamodb/models.py
```diff
@@ -551,7 +551,10 @@ def from_raw_data(cls: Type[_T], data: Dict[str, Any]) -> _T:
             attr_name = cls._dynamo_to_python_attr(name)
             attr = cls.get_attributes().get(attr_name, None)  # type: ignore
             if attr:
-                attributes[attr_name] = attr.deserialize(attr.get_value(value))  # type: ignore
+                try:
+                    attributes[attr_name] = attr.deserialize(attr.get_value(value))  # type: ignore
+                except TypeError:
+                    raise TypeError(f'attribute: {attr_name}')
```
garrettheel, line 557: I would consider introducing a new pynamo exception here that we can attach the attribute name context to, and do `raise SomePynamoExc() from e` to maintain the original `TypeError` context.

@ikonst thoughts?

ikonst: Agree, let's add it :)

garrettheel: @jonathantan let me know if this is clear, otherwise I can give more details!

jonathantan: Interesting I'm learning exception chaining for the first time but it sounds pretty straight forward according to https://www.python.org/dev/peps/pep-3134/#explicit-exception-chaining

jonathantan: done :)

pynamodb/models.py
```diff
@@ -551,7 +551,10 @@ def from_raw_data(cls: Type[_T], data: Dict[str, Any]) -> _T:
             attr_name = cls._dynamo_to_python_attr(name)
             attr = cls.get_attributes().get(attr_name, None)  # type: ignore
             if attr:
-                attributes[attr_name] = attr.deserialize(attr.get_value(value))  # type: ignore
+                try:
+                    attributes[attr_name] = attr.deserialize(attr.get_value(value))  # type: ignore
+                except TypeError as e:
+                    raise ValueError(f'attribute: {attr_name}') from e
```
garrettheel, line 557: I'm suggesting that we add a new exception type to `pynamodb/exceptions.py`, such as `AttributeDeserializationError` and then attach the attribute name as context there.

e.g

`raise AttributeDeserializationError(attr_name=attr_name) from e`

The `msg` can be constructed to include the attr name in the init of `AttributeDeserializationError`. I'd also consider subclassing `TypeError` for the new exception.

It would also be nice to add a test to cover this path

PR #818, opened 2020-08-21
Fix nullable attribute deserialization after update

WassimKallel:
Fix nullable attribute deserialization after an update for both NumberAttribute and JSONAttribute.

Linked issue: #627


garrettheel:
I don't think it makes sense to special-case this for a subset of attributes, I'd rather fix this at the root. I left a comment on the linked issue to help understand what is causing this issue

PR #819, opened 2020-08-29
add a test for version attribute

hallie:
I've been seeing an issue where if a model inherits from a base class that already has `VersionAttribute` defined, it will raise an error (even if it doesn't attempt to redefine the attribute). So updating the logic to check if the "duplicate" version attribute has the same name (which means its not actually being redefined)


ikonst:
eek, dynamodb-local crashed

ikonst:
@hallie can you merge master and retry? it crashes for no fault of yours, travis-ci just upgraded the jdk to a version that's too new for Amazon

Yaronn44:
@hallie Would be cool if this could be merge. I'm really needing this right now

lukedeo:
+1 🙂

jmphilli (approved):

jpinner-lyft (approved):

PR #826, opened 2020-09-03
Only add custom attribute names to the mapping.

jpinner-lyft:
This fixes a "bug" where attributes without custom names are added to the mapping in MapAttribute sub-subclasses.
```
MyMap(MapAttribute):
    foo = StringAttribute()

MySubMap(MyMap):
    pass
```
With this change `foo` is no longer added to `MySubMap._dynamo_to_python_attrs`

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -237,10 +237,10 @@ class AttributeContainerMeta(GenericMeta):
                 initialized = attribute._make_attribute()
 
             cls._attributes[name] = attribute
-            if attribute.attr_name is not None:
-                cls._dynamo_to_python_attrs[attribute.attr_name] = name
-            else:
+            if attribute.attr_name is None:
                 attribute.attr_name = name
+            if attribute.attr_name != name:
+                cls._dynamo_to_python_attrs[attribute.attr_name] = name
 
             if initialized and isinstance(attribute, MapAttribute):
                 # To support creating expressions from nested attributes, MapAttribute instances


jpinner-lyft:
rebased onto master

jpinner-lyft:
Normally the dynamo attribute map stores the attr_name overrides only when they differ:
```
>>> from pynamodb.attributes import MapAttribute, UnicodeAttribute
>>> class MyMap(MapAttribute):
...     foo = UnicodeAttribute()
...     bar = UnicodeAttribute(attr_name='baz')
... 
>>> MyMap._dynamo_to_python_attrs
{'baz': 'bar'}
```
(Note `foo` is absent because the pynamo and dynamo attribute names are the same.)

But when you subclass, all attributes on the root class have `attr_name` set from class creation so all get added:
```
>>> class MySubMap(MyMap):
...     pass
... 
>>> MySubMap._dynamo_to_python_attrs
{'baz': 'bar', 'foo': 'foo'}
```

A similar issue is that if you declare an `attr_name` that matches the python name it is added to the map:
```
>>> class MyMap(MapAttribute):
...     foo = UnicodeAttribute(attr_name='foo')
...     bar = UnicodeAttribute(attr_name='baz')
... 
>>> MyMap._dynamo_to_python_attrs
{'baz': 'bar', 'foo': 'foo'}
```

With this change we only add elements where the attribute name in python and dynamo differ
```
>>> from pynamodb.attributes import MapAttribute, UnicodeAttribute
>>> class MyMap(MapAttribute):
...     foo = UnicodeAttribute(attr_name='foo')
...     bar = UnicodeAttribute(attr_name='baz')
... 
>>> MyMap._dynamo_to_python_attrs
{'baz': 'bar'}
>>> class MySubMap(MyMap):
...     pass
... 
>>> MySubMap._dynamo_to_python_attrs
{'baz': 'bar'}
```

ikonst (approved):

PR #827, opened 2020-09-03
Remove internal attribute type names.

jpinner-lyft:
Replace the "long" version of the attribute types with the "short" version used in DynamoDB.
This change removes the "_SHORT" attribute type constants entirely.
ATTR_TYPE_MAP  is also removed as serialization no longer requires the map lookup for each attribute.


ikonst (approved):
Should we drop a note in the change log just so we don't forget?

PR #828, opened 2020-09-03
Always use attribute.attr_type to serialize List and Map elements.

jpinner-lyft:


ikonst (approved):

ikonst:
I'm not really sure what this even allowed.

When I try
```py
from pynamodb.models import *
from pynamodb.attributes import *


class FooModel(Model):
    bar = Attribute()


foo = FooModel()
foo.bar = 42
print(foo._serialize())
```
I get
```
  File "test.py", line 11, in <module>
    print(foo._serialize())
  File ".../site-packages/pynamodb/models.py", line 1089, in _serialize
    serialized = self._serialize_value(attr, value, null_check)
  File ".../site-packages/pynamodb/models.py", line 1124, in _serialize_value
    return {ATTR_TYPE_MAP[attr.attr_type]: serialized}
KeyError: None
```

garrettheel (approved):

garrettheel:
@ikonst yeah, I also poked around and couldn't find a reason why any of this should have been necessary. Given that we don't serialize `NULL` but just omit fields too, it all appears to be redundant

Issue #813, opened 2020-07-14
How to update an attribute in a ListAttribute of MapAttribute?


marzigolbaz:
I am learning how to use PynamoDB to work with AWS DynamoDB in my Python code. Here is what I have defined so far:

```
class OfficeEmployeeMap(MapAttribute):
    office_employee_id = NumberAttribute(hash_key=True)
    office_employee_direct_ids = UnicodeSetAttribute()

class Office(Model):
    class Meta:
        table_name = 'OfficeModel'
    office_id = NumberAttribute(hash_key=True)
    employees = ListAttribute(of=OfficeEmployeeMap)

Office.create_table(read_capacity_units=1, write_capacity_units=1) 
```

Here is how I appended to the list of map attributes (`employees`):

```
emp1 = OfficeEmployeeMap(
    office_employee_id=111,
    office_employee_direct_ids={"222"}
)

Office(
    office_id=1
).update(
    actions=[
        Office.employees.set((Office.employees|[]).append([emp1]))
    ]
)
```

And here is what I am seeing in the newly created DynamoDB table OfficeModel:

![image](https://user-images.githubusercontent.com/10997397/87453538-7886ab00-c5b7-11ea-993b-4aebc8f003e3.png)

Now I need to add another value, like `333`, to the set `office_employee_direct_ids` where `office_employee_id` == `111`, how can I do it with PynamoDB?

marzigolbaz:
@ikonst can you please help? this is blocking me.

ikonst:
@yelizariev You've been looking at this recently. Would have some time to advice?

marzigolbaz:
Tried this code:

```
TestPyDB(
    dict_name='host_vuln_plugin_dict'
).update(
    actions=[
        TestPyDB.host_vuln_plugin_dict[0].host_vuln_v_id.add({"hv2"})
    ],
    condition=(
        (TestPyDB.host_vuln_plugin_dict[0].plugin_id == '123')
    )
)
```
It did raise this error:

```
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
AttributeError: 'Path' object has no attribute 'host_vuln_v_id'
```

marzigolbaz:
Tried this code:

```
Office(
    office_id=1
).update(
    actions=[
        Office.employees[0].office_employee_direct_ids.add({"333"})
    ],
    condition=(
        (Office.employees[0].office_employee_id == '111')
    )
) 
```
But, I got this error:

```
[...]

PR #829, opened 2020-09-05
Protect against deserialize exceptions when attributes are unexpectedly NULL

bobbyciraldo:
Sometimes when adding db items through some other client than PynamoDB, there is a possibility of setting an attribute value to NULL. This can cause a json.dumps() exception when doing for example:
```
(Assume a model called FooModel has been defined with NumberAttribute `bar`. 
Now assume AWS Console has been used to create an item where `bar` is set to NULL.)
```
```
items = FooModel.query('pkey_val')
for foo_item in items:
    pass
```
```
Exception has occurred: TypeError
the JSON object must be str, bytes or bytearray, not NoneType
```

This PR special-cases `None` values and simply returns `None` in that case.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -364,6 +364,8 @@ class BinaryAttribute(Attribute[bytes]):
         """
         Returns a decoded string from base64
         """
+        if value is None:
+            return None
         try:
             return b64decode(value.decode(DEFAULT_ENCODING))
         except AttributeError:
@@ -470,7 +472,9 @@ class JSONAttribute(Attribute[Any]):
         """
         Deserializes JSON
         """
-        return json.loads(value, strict=False)
+        if value is None:
+            return None
+        return json.loads(value)
 
 
 class BooleanAttribute(Attribute[bool]):
@@ -488,6 +492,8 @@ class BooleanAttribute(Attribute[bool]):
             return False
 
     def deserialize(self, value):
+        if value is None:
+            return None
         return bool(value)
 
 
@@ -515,6 +521,8 @@ class NumberAttribute(Attribute[float]):
         """
         Decode numbers from JSON
         """
+        if value is None:
+            return None
         return json.loads(value)
 
 
@@ -547,6 +555,8 @@ class VersionAttribute(NumberAttribute):
         """
         Decode numbers from JSON and cast to int.
         """
+        if value is None:
+            return None
         return int(super().deserialize(value))
 
 
@@ -616,6 +626,8 @@ class UTCDateTimeAttribute(Attribute[datetime]):
         """
         Takes a UTC datetime string and returns a datetime object
         """
+        if value is None:
+            return None
         try:
             return _fast_parse_utc_datestring(value)
         except (ValueError, IndexError):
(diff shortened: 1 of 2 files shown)


jpinner-lyft:
This isn't quite correct. If an attribute is stored as `{'NULL': True}` then when the model deserializes the value it will be ignored since the attribute type will not match.

bobbyciraldo:
@jpinner-lyft Hi! Thanks a ton for taking a look at this. Here is a more concrete example.
```
# demo.py

class UserModel(Model):
    class Meta:
        table_name = 'temp_table'
        region = 'us-west-2'
    age = NumberAttribute(null=True)

if __name__ == "__main__":
    UserModel.create_table(read_capacity_units=1, write_capacity_units=1)
    user = UserModel('test@example.com', age=21)
    user.save()
    user.update([UserModel.age.set(None)])  # Simulate setting NumberAttribute to NullAttribute
    users = UserModel.query('test@example.com')

    # The iteration below is what raises an exception
    for user in users:
        pass
```
```
$ python demo.py
  File "demo.py", line 20, in <module>
    for user in users:
  File "/Users/rc250072/code/demo/venv/lib/python3.7/site-packages/pynamodb/pagination.py", line 183, in __next__
    item = self._map_fn(item)
  File "/Users/rc250072/code/demo/venv/lib/python3.7/site-packages/pynamodb/models.py", line 513, in from_raw_data
    attributes[attr_name] = attr.deserialize(attr.get_value(value))
  File "/Users/rc250072/code/demo/venv/lib/python3.7/site-packages/pynamodb/attributes.py", line 499, in deserialize
    return json.loads(value)
  File "/Users/rc250072/.pyenv/versions/3.7.7/lib/python3.7/json/__init__.py", line 341, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not NoneType
```

ikonst:
The first surprising thing here is the way `UserMode.age.set(None)` works. In pynamodb.attributes._get_class_for_serialize, we pick an attribute type based on the pythonic value's type rather than use `age`'s attribute. In general the `Value` class (in `pynamodb.expressions.operand`) does some strange things, in how it'd accept arbitrary pythonic values as well as dictionaries of the form `{"S": "value"}` (where S is any DynamoDB type).

This same weirdness was exhibited in this issue:
https://github.com/pynamodb/PynamoDB/issues/813

This is some tech debt we should clean up.

The crash is due to:
```
class NumberAttribute(Attribute):
    def deserialize(self, value):
        return json.loads(value)
```

This is being called as  `NumberAttribute.deserialize({"NULL": "true"}.get("N"))` which ends up invoking `json.loads(None)` which is an error.

So, yeah, PynamoDB isn't gracefully handling items whose attributes' DynamoDB types are not what the model says they should be.

jpinner-lyft:
I'm surprised this is happening because `Model._deserialize` should be skipping `None` values. What version of pynamo are you using?

jpinner-lyft:
Ok, I see what's going on here. Model._deserialize and Model.from_raw_data handle None values differently.

@ikonst I think we should modify from_raw_data to skip deserializing attributes where the type doesn't match.

jpinner-lyft:
@bobbyciraldo thank you for the example. instead of modifying the attribute serialization behavior I would take a different approach and modify `Model.from_raw_data` so that it skips deserializing `None` values (in the same way `Model._deserialize` does.

jpinner-lyft:
As part of some other work refactoring deserialization, this has been fixed.
If you continue to see issues here on v5.0.0b1 please open a new issue and tag me on it so I can investigate.
Thanks!

PR #830, opened 2020-09-05
DynamoDB now supports empty values for Binary and String attributes. Fixes #804

### LINKED ISSUE #804: Empty String/Binary values support?
DynamoDB recently supported empty values for _non-key_ String/Binary attributes.

Though PynamoDB currently supports empty value as null (None), 
does PynamoDB have any plans to support this new feature of DynamoDB?

jpinner-lyft:


pynamodb/attributes.py
```diff
@@ -362,12 +362,9 @@ def serialize(self, value):
...
     def deserialize(self, value):
         """
-        Returns a decoded string from base64
+        Returns a decoded byte string from a base64 encoded value
         """
-        try:
-            return b64decode(value.decode(DEFAULT_ENCODING))
-        except AttributeError:
-            return b64decode(value)
+        return b64decode(value)
```
ikonst, line 367: This is a result of being py3 only now?

ikonst (approved):

pynamodb/attributes.py
jpinner-lyft: yes, in py3 b64decode takes either a string or bytes so we don't have to try to decode first:
https://docs.python.org/3/library/base64.html#base64.b64decode

PR #831, opened 2020-09-06
Simplify set attribute serialization/deserialization.

jpinner-lyft:
This change removes python 2 support, the `SetMixin` class, the unnecessary sorting during serialization.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -350,7 +350,7 @@ class BinarySetAttribute(Attribute[Set[bytes]]):
         """
         Returns a list of base64 encoded binary strings. Encodes empty sets as "None".
         """
-        return [b64encode(v).decode(DEFAULT_ENCODING) for v in value] if value else None
+        return [b64encode(v).decode(DEFAULT_ENCODING) for v in value] or None
 
     def deserialize(self, value):
         """
@@ -377,7 +377,7 @@ class UnicodeSetAttribute(Attribute[Set[str]]):
         """
         Returns a list of strings. Encodes empty sets as "None".
         """
-        return list(value) if value else None
+        return list(value) or None
 
     def deserialize(self, value):
         """
@@ -458,7 +458,7 @@ class NumberSetAttribute(Attribute[Set[float]]):
         """
         Encodes a set of numbers as a JSON list. Encodes empty sets as "None".
         """
-        return [json.dumps(v) for v in value] if value else None
+        return [json.dumps(v) for v in value] or None
 
     def deserialize(self, value):
         """
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -320,34 +320,6 @@ class AttributeContainer(metaclass=AttributeContainerMeta):
         return self is not other
 
 
-class SetMixin(object):
-    """
-    Adds (de)serialization methods for sets
-    """
-    def serialize(self, value):
-        """
-        Serializes a set
-
-        Because dynamodb doesn't store empty attributes,
-        empty sets return None
-        """
-        if value is not None:
-            try:
-                iter(value)
-            except TypeError:
-                value = [value]
-            if len(value):
-                return [json.dumps(val) for val in sorted(value)]
-        return None
-
-    def deserialize(self, value):
-        """
-        Deserializes a set
-        """
-        if value and len(value):
-            return {json.loads(val) for val in value}
-
-
 class BinaryAttribute(Attribute[bytes]):
     """
     A binary attribute
@@ -367,7 +339,7 @@ class BinaryAttribute(Attribute[bytes]):
         return b64decode(value)
 
 
-class BinarySetAttribute(SetMixin, Attribute[Set[bytes]]):
+class BinarySetAttribute(Attribute[Set[bytes]]):
     """
     A binary set
     """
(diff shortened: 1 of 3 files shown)


jpinner-lyft:
(Temporary notes to remind myself what I'm trying to do.)

DynamoDB treats empty sets as equivalent to the element not existing.
This distinction is made clear when looking at how update expressions (specifically ADD/REMOVE) operate on sets.

Adding an element to a set attribute that does not exist creates it. Removing the final element in a set attribute removes it.

This may seem like empty sets and `None` could be treated interchangeably but there is a subtle difference when looking at lists. Consider the following list: `{"string_set_list":{"L":[{"SS":["foo"]},{"SS":["bar"]}]}}`
The update expressions `REMOVE string_set_list[0] {"SS": ["foo"]}, ADD string_set_list[0] {"SS": ["foo"]}` result in a list with only one element: `{"string_set_list":{"L":[{"SS":["bar", "foo"]}]}}`

This has implications for how we would serialize the following python dictionary: `{"string_set_list:[{"foo"}, {}, {"bar"}]}`

jpinner-lyft:
@ikonst @garrettheel this is a precursor to an upcoming change to ListAttribute.

I will be expanded ListAttribute so that `of` can take any attribute type (including set attributes).
```
list_of_string_sets = ListAttribute(of=UnicodeSetAttribute)
list_of_string_sets = [{'foo'}, {'bar'}]
```

Before I make the ListAttribute change I want to be more explicit about raising an exception during deserialization of None by set attributes to prevent introducing any bugs since there are a few edge cases around sets and update expressions.

Because DynamoDB does not support empty sets in the above example, if you do an `REMOVE list_of_string_sets {'SS':['foo']}` followed by an ADD, you get a single element when the remove operation empties the set:
```
list_of_string_sets = [{'bar', 'foo'}]
```

pynamodb/attributes.py
```diff
@@ -376,65 +348,42 @@ class BinarySetAttribute(SetMixin, Attribute[Set[bytes]]):
...
     def serialize(self, value):
         """
-        Returns a base64 encoded binary string
+        Returns a list of base64 encoded binary strings. Encodes empty sets as "None".
         """
-        if value and len(value):
-            return [b64encode(val).decode(DEFAULT_ENCODING) for val in sorted(value)]
-        else:
-            return None
+        return [b64encode(v).decode(DEFAULT_ENCODING) for v in value] if value else None
```
jpinner-lyft, line 353: removes the sorting and merges the `if value and len(value)` check

pynamodb/attributes.py
```diff
@@ -376,65 +348,42 @@ class BinarySetAttribute(SetMixin, Attribute[Set[bytes]]):
...
         """
-        Returns a decoded string from base64
+        Returns a set of decoded byte strings from base64 encoded values.
         """
-        try:
-            if value and len(value):
-                return {b64decode(val.decode(DEFAULT_ENCODING)) for val in value}
-        except AttributeError:
-            return {b64decode(val) for val in value}
+        return {b64decode(v) for v in value}
```
jpinner-lyft, line 359: In python 3 b64decode takes bytes and string objects so this removes the `decode`.
Removes the check if `value` is `None` to enforce decoding only DynamoDB supported values for the `BS` type.

pynamodb/attributes.py
```diff
@@ -376,65 +348,42 @@ class BinarySetAttribute(SetMixin, Attribute[Set[bytes]]):
...
-    def serialize(self, value):
-        if value is not None:
-            try:
-                iter(value)
-            except TypeError:
-                value = [value]
-            if len(value):
-                return [self.element_serialize(val) for val in sorted(value)]
-        return None
+        return list(value) if value else None
```
jpinner-lyft, line 380: Removes the sorting and the check for a non-iterable value. (The iteration check was already a bit broken since singleton-strings are utterable and would return a set of their characters.

pynamodb/attributes.py
```diff
@@ -376,65 +348,42 @@ class BinarySetAttribute(SetMixin, Attribute[Set[bytes]]):
...
-
-class UnicodeAttribute(Attribute[str]):
-    """
-    A unicode attribute
-    """
-    attr_type = STRING
+        """
+        Returns a set from a list of strings.
+        """
+        return set(value)
```
jpinner-lyft, line 386: Removes the check if value is None to enforce decoding only DynamoDB supported values for the SS type.

pynamodb/attributes.py
```diff
@@ -506,6 +447,26 @@ def deserialize(self, value):
...
+        """
+        Encodes a set of numbers as a JSON list. Encodes empty sets as "None".
+        """
+        return [json.dumps(v) for v in value] if value else None
+
+    def deserialize(self, value):
+        """
+        Returns a set from a JSON list of numbers.
+        """
+        return {json.loads(v) for v in value}
```
jpinner-lyft, line 467: Removes the check if value is None to enforce decoding only DynamoDB supported values for the NS type.

pynamodb/attributes.py
ikonst:
```suggestion
        return [b64encode(v).decode(DEFAULT_ENCODING) for v in value] or None
```

pynamodb/attributes.py
ikonst:
```suggestion
        return list(value) or None
```

pynamodb/attributes.py
jpinner-lyft: that throws a type error when value is None: `TypeError: 'NoneType' object is not iterable`

tests/test_model.py
```diff
@@ -2119,7 +2119,7 @@ def test_index_queries(self):
                         'S': u'bar'
                     },
                     ':2': {
-                        'S': '1'
+                        'S': u'baz'
```
ikonst, line 2122: bah, we need to run modernize over this (in a separate PR)

jpinner-lyft:
@ikonst checked the code and we should not be attempting to serialize None

ikonst (approved):

PR #839, opened 2020-09-10
Simplify map attribute deserialization.

jpinner-lyft:


pynamodb/attributes.py
```diff
@@ -860,31 +857,17 @@ def _get_serialize_class(cls, key, value):
...
-    return value[key]
-
-
-def _get_class_for_deserialize(value):
-    value_type = next(iter(value.keys()))
-    if value_type not in DESERIALIZE_CLASS_MAP:
-        raise ValueError('Unknown value: ' + str(value))
-    return DESERIALIZE_CLASS_MAP[value_type]
+    def _get_deserialize_class(cls, attr_name, attr_type):
+        if not cls.is_raw() and attr_type != NULL:
```
jpinner-lyft, line 861: Instead of special casing the value handling, support NULL attributes by bypassing the attribute lookup and returning NullAttribute().

pynamodb/attributes.py
```diff
@@ -820,18 +820,15 @@ def deserialize(self, values):
...
-                deserialized_value = attr_class.deserialize(attr_value)
-
-            deserialized_dict[key] = deserialized_value
+        for k, v in values.items():
+            attr_name = self._dynamo_to_python_attr(k)
+            attr_type, attr_value = next(iter(v.items()))
+            attr_class = self._get_deserialize_class(attr_name, attr_type)
+            if attr_class and attr_class.attr_type != attr_type:
+                raise ValueError("Cannot deserialize '{}' attribute from type: {}".format(
+                    attr_name, attr_type))
```
jpinner-lyft, line 829: Adds a check to ensure we are deserializing the correct type (previously we would send any value to `deserialize` and wait for a type error).

pynamodb/attributes.py
```diff
@@ -860,31 +857,17 @@ def _get_serialize_class(cls, key, value):
...
+            return cls.get_attributes().get(attr_name)
+        return DESERIALIZE_CLASS_MAP[attr_type]
 
 
 def _get_class_for_serialize(value):
     if value is None:
         return NullAttribute()
     if isinstance(value, MapAttribute):
-        return type(value)()
+        return value
```
jpinner-lyft, line 870: Serialize subclasses using the instance stored in the attribute_values dictionary instead of instantiating a new one.

pynamodb/attributes.py
ikonst: 👏  finally, saw it bit folks more than once

pynamodb/attributes.py
```diff
@@ -820,18 +820,15 @@ def deserialize(self, values):
...
-
-            deserialized_dict[key] = deserialized_value
+        for k, v in values.items():
+            attr_name = self._dynamo_to_python_attr(k)
+            attr_type, attr_value = next(iter(v.items()))
+            attr_class = self._get_deserialize_class(attr_name, attr_type)
+            if attr_class and attr_class.attr_type != attr_type:
+                raise ValueError("Cannot deserialize '{}' attribute from type: {}".format(
+                    attr_name, attr_type))
+            if attr_class:
```
ikonst, line 830: nit - nest the ifs?

pynamodb/attributes.py
garrettheel: I've seen people write custom attributes that handle multiple types to support migrations. IIRC that was not within a MapAttribute, but is it something we should still support?

pynamodb/attributes.py
```diff
@@ -860,31 +857,17 @@ def _get_serialize_class(cls, key, value):
...
-        return None
-    return value[key]
-
-
-def _get_class_for_deserialize(value):
-    value_type = next(iter(value.keys()))
-    if value_type not in DESERIALIZE_CLASS_MAP:
-        raise ValueError('Unknown value: ' + str(value))
-    return DESERIALIZE_CLASS_MAP[value_type]
+    def _get_deserialize_class(cls, attr_name, attr_type):
```
ikonst, line 860: add type hints for clarity?

pynamodb/attributes.py
ikonst: (I wonder why it was done this way in the first place)

pynamodb/attributes.py
ikonst: @garrettheel

Was it something like this?
```
foo_old = NumberAttribute(attr_name='foo')
foo_new =UnicodeAttribute(attr_name='foo') 
```

garrettheel: Nope, more like:

```python
class NumberUnicodeAttribute(UnicodeAttribute):
    def deserialize(self, value: str) -> int:
        try:
            deserialized_value = json.loads(value)
        except json.decoder.JSONDecodeError:
            deserialized_value = value
        return int(deserialized_value)
```

garrettheel: I suppose the issue there is that the attr_type being used to deserialize did not match the one stored in DDB, but it happened to work anyway. This will uncover those issues immediately by failing to deserialize, but might cause a little bit of pain.

ikonst: So supposedly there's just one `Attribute.attr_type` so an attribute has to choose whether it's an `S` or an `N`. However, you can override `Attribute.get_value` to modify this behavior **for when your attribute is directly on the the Model**. This doesn't apply nested MapAttribute deserializations, since it's not implemented symmetrically (but maybe it should be).

jpinner-lyft: I like the idea of standardizing on using `get_value` to special case this behavior and raising an exception if this returns `None`.

pynamodb/attributes.py
jpinner-lyft: removed entirely

pynamodb/attributes.py
jpinner-lyft: replaced by splitting deserialization into raw and non-raw variants

pynamodb/attributes.py
```diff
@@ -311,6 +314,15 @@ def _set_attributes(self, **attributes: Attribute) -> None:
                 raise ValueError("Attribute {} specified does not exist".format(attr_name))
             setattr(self, attr_name, attr_value)
 
+    def _deserialize(self, attribute_values: Dict[str, Dict[str, Any]]) -> None:
+        """
+        Sets attributes sent back from DynamoDB on this object
+        """
+        for name, attr in self.get_attributes().items():
+            value = attribute_values.get(attr.attr_name, {NULL: True})
+            value = None if NULL in value else attr.deserialize(attr.get_value(value))
```
jpinner-lyft, line 323: this is now where we skip any attributes that serialized to NULL (or had NULL set on them outside of pynamo)
still not my favorite work-around but i guess no worse than what we currently do

garrettheel (approved):
Beautiful

pynamodb/attributes.py
```diff
@@ -311,6 +313,18 @@ def _set_attributes(self, **attributes: Attribute) -> None:
                 raise ValueError("Attribute {} specified does not exist".format(attr_name))
             setattr(self, attr_name, attr_value)
 
+    def _deserialize(self, attribute_values: Dict[str, Dict[str, Any]]) -> None:
+        """
+        Sets attributes sent back from DynamoDB on this object
+        """
+        self.attribute_values = {}
+        self._set_defaults(_user_instantiated=False)
```
jpinner-lyft, line 321: This fixes a bug where default values weren't getting set when refreshing or updating models.

PR #841, opened 2020-09-11
Improve typed list attribute expression creation.

jpinner-lyft:
Return properly configured attributes when accessing typed list attribute elements to enable accessing nested elements.

Consider the following list of maps:
```
from pynamodb.attributes import  ListAttribute, MapAttribute,UnicodeAttribute

class StringMap(MapAttribute):
    bar = UnicodeAttribute()

class MyModel(Model):
    foo = UnicodeAttribute(hash_key=True)
    maps = ListAttribute(of=StringMap)
```
With this change it is now possible to write expressions that reference the embedded `bar` attribute:
```
>>> MyModel.maps[0].bar.exists()
attribute_exists (maps[0].bar)
```
Previously this would raise an exception:
```
>>> MyModel.maps[0].bar.exists()
  File "<stdin>", line 1, in <module>
AttributeError: 'Path' object has no attribute 'bar'
```

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -7,7 +7,7 @@ import json
 import time
 import warnings
 from base64 import b64encode, b64decode
-from copy import copy, deepcopy
+from copy import deepcopy
 from datetime import datetime, timedelta
 from dateutil.parser import parse
 from dateutil.tz import tzutc
@@ -898,7 +898,7 @@ def _fast_parse_utc_datestring(datestring):
 
 class ListAttribute(Generic[_T], Attribute[List[_T]]):
     attr_type = LIST
-    element_type: Optional[Type[Attribute[_T]]] = None
+    element_type: Optional[Type[Attribute]] = None
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


pynamodb/attributes.py
```diff
@@ -945,18 +943,41 @@ def deserialize(self, values):
...
+            element_attr = self.element_type()
+            if isinstance(element_attr, MapAttribute):
+                element_attr._make_attribute()  # ensure attr_name exists
+            deserialized_lst = []
+            for idx, attribute_value in enumerate(values):
+                value = None
+                if NULL not in attribute_value:
+                    # set attr_name in case `get_value` raises an exception
+                    element_attr.attr_name = '{}[{}]'.format(self.attr_name, idx)
+                    value = element_attr.deserialize(element_attr.get_value(attribute_value))
```
jpinner-lyft, line 956: Additionally this change modifies `deserialize` to call `get_value` so that attributes which special case attribute lookup (e.g. for migrations) are properly supported in lists.

garrettheel (approved):

PR #842, opened 2020-09-11
Enable polymorphism by serializing all attributes in subclasses.

jpinner-lyft:
Currently when serializing MapAttribute subclasses, only attributes defined in the type being serialized are written.
```
>>> from pynamodb.attributes import ListAttribute, MapAttribute, UnicodeAttribute
>>> from pynamodb.models import Model
>>> class SuperclassMap(MapAttribute):
...     bar = UnicodeAttribute()
... 
>>> class SubclassMap(SuperclassMap):
...     baz = UnicodeAttribute()
... 
>>> class MyModel(Model):
...     foo = UnicodeAttribute(hash_key=True)
...     map = SuperclassMap()
...     list = ListAttribute(of=SuperclassMap)
... 
>>> subclass_map = SubclassMap(bar='bar', baz='baz')
>>> MyModel(foo='foo', map=subclass_map, list=[subclass_map])._serialize()
{'attributes': {'list': {'L': [{'M': {'bar': {'S': 'bar'}, 'baz': {'S': 'baz'}}}]}, 'map': {'M': {'bar': {'S': 'bar'}}}}, 'HASH': 'foo'}
```
(note `mm.list[0]` serialize `baz` but `mm.map` does not)

With this change (needed for discriminator support) all attributes in the subclass are written to dynamo.
```
>>> subclass_map = SubclassMap(bar='bar', baz='baz')
>>> MyModel(foo='foo', map=subclass_map, list=[subclass_map])._serialize()
{'attributes': {'list': {'L': [{'M': {'bar': {'S': 'bar'}, 'baz': {'S': 'baz'}}}]}, 'map': {'M': {'bar': {'S': 'bar'}, 'baz': {'S': 'baz'}}}}, 'HASH': 'foo'}
```

In addition, this commit fixes a bug where required elements in inner maps were not correctly validated (see added test).

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -878,12 +878,6 @@ class MapAttribute(Attribute[Mapping[_KT, _VT]], AttributeContainer):
             result[key] = value.as_dict() if isinstance(value, MapAttribute) else value
         return result
 
-    @classmethod
-    def _get_serialize_class(cls, key, value):
-        if not cls.is_raw():
-            return cls.get_attributes().get(key)
-        return _get_class_for_serialize(value)
-
 
 def _get_class_for_serialize(value):
     if value is None:
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -825,18 +825,20 @@ class MapAttribute(Attribute[Mapping[_KT, _VT]], AttributeContainer):
         return all(self.is_correctly_typed(k, v) for k, v in self.get_attributes().items())
 
     def serialize(self, values):
-        if isinstance(values, type(self)) and not values.is_raw():
-            return values._serialize()
-
         if not self.is_raw():
-            # Need to serialized based on the attributes in the class; however,
-            # the value passed in was not an instance of the class.
-            instance = type(self)()
-            instance.attribute_values = {}  # clear any defaults
-            for name in values:
-                if name in self.get_attributes():
-                    setattr(instance, name, values[name])
-            return instance._serialize()
+            # This is a subclassed MapAttribute that acts as an AttributeContainer.
+            # Serialize the values based on the attributes in the class.
+
+            if not isinstance(values, type(self)):
+                # Copy the values onto an instance of the class for serialization.
+                instance = type(self)()
+                instance.attribute_values = {}  # clear any defaults
+                for name in values:
+                    if name in self.get_attributes():
+                        setattr(instance, name, values[name])
+                values = instance
+
+            return values._serialize()
 
         # Continue to serialize NULL values in "raw" map attributes for backwards compatibility.
         # This special case behavior for "raw" attributes should be removed in the future.
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -828,25 +828,29 @@ class MapAttribute(Attribute[Mapping[_KT, _VT]], AttributeContainer):
         if isinstance(values, type(self)) and not values.is_raw():
             return values._serialize()
 
+        if not self.is_raw():
+            # Need to serialized based on the attributes in the class; however,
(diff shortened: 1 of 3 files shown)


pynamodb/attributes.py
```diff
@@ -807,6 +825,9 @@ def validate(self):
         return all(self.is_correctly_typed(k, v) for k, v in self.get_attributes().items())
 
     def serialize(self, values):
+        if isinstance(values, type(self)) and not values.is_raw():
```
ikonst, line 828: Is this because ListAttribute creates an anonymous `element_type` attribute instance to serialize?

jpinner-lyft: ListAttributes serializes based on the value itself if `isinstance` passes.

ikonst: Well now it does, just validating why this logic was needed.

jpinner-lyft: Right, so before that change ListAttribute also had the same behavior where it would only serialize elements based on the attributes in the declared type.

Supporting polymorphism requires that we serialize all elements so we don't drop any data. Now until discriminator map support lands we won't be able to deserialize all of the data on disk since we won't know which subtype to instantiate.

pynamodb/models.py
```diff
@@ -1115,27 +1115,14 @@ def _serialize(self, attr_map=False, null_check=True) -> Dict[str, Any]:
...
-                    attrs[attributes][attr.attr_name] = serialized
-
+        attrs: Dict[str, Dict] = {attributes: super()._serialize(null_check)}
+        if not attr_map:
+            if self._hash_key_attribute().attr_name in attrs[attributes]:
+                hash_key_attribute_value = attrs[attributes].pop(self._hash_key_attribute().attr_name)
+                attrs[HASH] = hash_key_attribute_value[self._hash_key_attribute().attr_type]
+            if self._range_keyname and self._range_key_attribute().attr_name in attrs[attributes]:
+                range_key_attribute_value = attrs[attributes].pop(self._range_key_attribute().attr_name)
+                attrs[RANGE] = range_key_attribute_value[self._range_key_attribute().attr_type]
```
ikonst, line 1125: nit - maybe use `pop(..., None)`
```suggestion
            hash_key_attribute_value = attrs[attributes].pop(self._hash_key_attribute().attr_name, None)
            if hash_key_attribute_value is not None:
                attrs[HASH] = hash_key_attribute_value[self._hash_key_attribute().attr_type]
            if self._range_keyname:
                range_key_attribute_value = attrs[attributes].pop(self._range_key_attribute().attr_name, None)
                if range_key_attribute_value is not None:
                    attrs[RANGE] = range_key_attribute_value[self._range_key_attribute().attr_type]
```

ikonst (approved):

pynamodb/models.py
jpinner-lyft: :+1:

pynamodb/attributes.py
```diff
@@ -807,25 +825,32 @@ def validate(self):
...
-            rval[attr_name] = {attr_class.attr_type: serialized}
+        if not self.is_raw():
+            # Need to serialized based on the attributes in the class; however,
+            # the value passed in was not an instance of the class.
+            instance = type(self)()
+            instance.attribute_values = {}  # clear any defaults
+            for name in values:
+                if name in self.get_attributes():
+                    setattr(instance, name, values[name])
+            return instance._serialize()
```
jpinner-lyft, line 839: This fixes a bug where we did not check that required attributes were set since we iterated through the values to be serialized and not the attributes in the class. Now we "cast" whatever attribute container is passed in into the class we are serializing which performs the appropriate validation / null checks.

PR #850, opened 2020-09-16
UTCDateTimeAttribute now requires the date string format '%Y-%m-%dT%H:%M:%S.%f%z'.

jpinner-lyft:
This removes the dependency on `python-dateutil`.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -11,7 +11,6 @@ from copy import deepcopy
 from datetime import datetime
 from datetime import timedelta
 from datetime import timezone
-from dateutil.parser import parse
 from inspect import getfullargspec
 from inspect import getmembers
 from typing import Any, Callable, Dict, Generic, List, Mapping, Optional, TypeVar, Type, Union, Set, overload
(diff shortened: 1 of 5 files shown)


pynamodb/attributes.py
```diff
@@ -668,16 +667,25 @@ def deserialize(self, value):
         """
         Takes a UTC datetime string and returns a datetime object
         """
+        return self._fast_parse_utc_date_string(value)
+
+    @staticmethod
+    def _fast_parse_utc_date_string(date_string):
```
jpinner-lyft, line 673: This has been moved inside the class and slightly modified to:
- check if strings have the proper length (avoids IndexError and a varying like microsecond field)
- rely on the microsecond field being left padded with 0s to remove the multiplication and rounding

rowillia (approved):

pynamodb/attributes.py
ikonst:
```suggestion
    def _fast_parse_utc_date_string(date_string: str) -> datetime:
```

ikonst (approved):

PR #853, opened 2020-09-28
[Experimental] add asyncio support

garrettheel:
I will likely break this up into separate PRs if we decide to move ahead, but opening this up as a place to discuss what asyncio support might look like.

The approach is inspired by [this comment](https://github.com/python-trio/hip/issues/1#issuecomment-322028457) in python-trio, which details an interesting way to run async functions in a sync context, so long as they're actually synchronous under the hood. This allows us to write a single async implementation, and then dynamically generate the sync interface.

As far as the public interface goes, this currently leaves the sync interface unchanged and exposes `_async` variants for methods. I've only implemented the base connection so far, so this is far from complete.

I'm interested in any feedback on the approach and the interface in particular.

### DIFF
--- a/pynamodb/async_util.py
+++ b/pynamodb/async_util.py
@@ -1,0 +1,19 @@
+import functools
+
+
+def run_secretly_sync_async_fn(async_fn, *args, **kwargs):
+    # From https://github.com/python-trio/hip/issues/1#issuecomment-322028457
+    coro = async_fn(*args, **kwargs)
+    try:
+        coro.send(None)
+    except StopIteration as exc:
+        return exc.value
+    else:
+        raise RuntimeError("you lied, this async function is not secretly synchronous")
+
+
+def wrap_secretly_sync_async_fn(async_fn):
+    @functools.wraps(async_fn)
+    def wrap(*args, **kwargs):
+        return run_secretly_sync_async_fn(async_fn, *args, **kwargs)
+    return wrap
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -2,6 +2,7 @@
 Lowest level connection
 """
 import asyncio
+import functools
 import json
 import logging
 import random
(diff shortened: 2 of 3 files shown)


svix-yair:
Hey @garrettheel,
Can you please shed some light on why you abandoned this PR?
We are experimenting with Pynamodb and we need async support.

It would be great to give us some pointers and share any information you already have so we can help adding the Async support for Pynamodb.

garrettheel:
hey @svix-yair, apologies for forgetting to leave an update here. I'll need to page this back in, but I think the main issue that I hit was `aiobotocore` being the only library that supports this and:
 1) it brings along a large number of dependencies, where we've typically tried to keep deps very minimal for PynamoDB
 2) it's not sanctioned by AWS or officially supported. there's little to suggest that this will stick around for the long term

I was waiting/hoping that botocore would implement asyncio natively, which would be a much better path to supporting it, but that still hasn't happened sadly. The other alternative is to forgo botocore for everything but a few utility tasks (e.g request signing, auth) and construct/send the requests ourselves (either using lower-level native python APIs or with something like httpx).

I'm definitely still interested in making progress on this and would welcome any help from others.

tasn:
Hey @garrettheel, thanks for all of the information and all of your work on pynamo! I work with @svix-yair, and I've been trying to get an async pynamo working for us.

I understand your concerns about using a 3rd party not sanctioned by AWS, though realistically AWS isn't going to have an async version any time soon, so a community effort is the best we've got. I think the aiobotocore people aren't going anyway, it's the same ones behind apg and aiohttp (which aiobotocore uses). Additionally, many of the deps are just other deps of their own projects. In summary, I don't think it's a massive concern.

The httpx idea sounds very promising, though it also sounds like a lot of work and essentially re-implementing aiobotocore. I think aiobotocore just uses aiohttp so maybe you'd have better luck replacing aiohttp there with something that doesn't pull as many deps?

I wish I could have helped you with all of the above, though tbh, it sounds like a very difficult task for an AWS noob like myself, especially since this task requires deep AWS expertise, or at least deep understanding of how to interact with the AWS services.

I'll try to get this patch working with `master` and test it with our code. Happy to contribute it back if of interest, let me know.

PR #855, opened 2020-09-28
use f-strings

ikamensh:
I've used [flynt](https://github.com/ikamensh/flynt) to convert the source to use f-strings:
```
Execution time:                            1.278s
Files modified:                            19
Character count reduction:                 1400 (0.22%)

Per expression type:
No old style (`%`) expressions attempted.
`.format(...)` calls attempted:            169/170 (99.4%)
No concatenations attempted.
F-string expressions created:              158
```
Since unit tests only run for python 3.6+, I gathered this should be acceptable for the dependencies.

I've also done a manual change in a place where f-string wouldn't be created automatically:

`"{0} is not a valid comparison operator: {0}".format(operator)`
-> `"{operator} is not a valid comparison operator."`

### DIFF
--- a/examples/attributes.py
+++ b/examples/attributes.py
@@ -14,7 +14,7 @@ class Color(object):
         self.name = name
 
     def __str__(self):
-        return "<Color: {}>".format(self.name)
+        return f"<Color: {self.name}>"
 
 
 class PickleAttribute(BinaryAttribute):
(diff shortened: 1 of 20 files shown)


ikamensh:
Hello hello - any opinions on this?

ikonst:
Maybe we should pre-commit along with [pyupgrade](https://github.com/asottile/pyupgrade) to modernize this and other aspects of our codebase?

Issue #152, opened 2016-08-18
Convert model instance to JSON


karloscodes:
Is there a way to convert a model instance to JSON??, I'm getting `<Obj> is not JSON serializable` when trying to serialize with JSON module

csimmons0:
Hi, Carlos. I just submitted a pull request (#156) that may provide what you're looking for. It lets you convert a model instance to a Python dictionary, and you can easily convert a dictionary to JSON.

yedpodtrzitko:
@ccverak you can use `Model()._serialize()`

jdno:
As @csimmons0 said in his pull request: `_serialize()` is too close to the data representation in DynamoDB and thus not suitable for (my) APIs. Besides, `_serialize()` looks like a  "private" method and should not be used by third-party code.

I agree that we need a method that produces a dict or JSON representation of the model object that abstracts from the underlying data representation. For example, here is what I get when I serialize my object with `_serialize()`:

```
{
  'attributes': {
    'owner': {'S': u'test|12345678'},
    'updated': {'S': u'2016-12-19T14:52:32.140080+00:00'},
    'name': {'S': u'Test Organization'},
    'members': {'SS': [u'test|12345678']},
    'created': {'S': u'2016-12-19T14:52:32.140106+00:00'}
  },
  'HASH': u'59fc2bcd-230d-427c-ae9f-9834bf29fc8a'
}
```

However, the following is what I need to pass it along to my user-facing API:

```python
{
  'id': '59fc2bcd-230d-427c-ae9f-9834bf29fc8a',
  'name': 'Test Organization',
  'owner': 'test|12345678',
  'members': [
    'test|12345678'
  ],
  'created': '2016-12-19T14:52:32.140106+00:00',
  'updated': '2016-12-19T14:52:32.140106+00:00'
}
```

Would be great to get some feedback on this from @jlafon so that we can merge (or work on) #156.

mathom:
@jdno and @ccverak:

I'm using this in my project - I hope it helps if you're still stuck!

```
class ModelEncoder(json.JSONEncoder):
    def default(self, obj):
        if hasattr(obj, 'attribute_values'):
            return obj.attribute_values
        elif isinstance(obj, datetime.datetime):
            return obj.isoformat()
        return json.JSONEncoder.default(self, obj)


def json_dumps(obj):
    return json.dumps(obj, cls=ModelEncoder)
```

kadrach:
[...]

PR #857, opened 2020-10-01
Add support for JSON serialization.

jpinner-lyft:
Model instances now support two types of serialization:
- `serialize/deserialize` can be used to convert models to DynamoDB AttributeValue dictionaries
- `to_json/from_json` can be used to convert models to JSON strings

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -4,6 +4,7 @@ Release Notes
 Unreleased
 ----------
 * The ``IndexMeta`` class has been removed. Now ``type(Index) == type``.
+* JSON serialization support (``Model.to_json`` and ``Model.from_json``) has been added.
 
 
 v5.1.0
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -17,11 +17,20 @@ from typing import Any, Callable, Dict, Generic, List, Mapping, Optional, TypeVa
 from typing import TYPE_CHECKING
 
 from pynamodb._compat import GenericMeta
-from pynamodb.constants import (
-    BINARY, BINARY_SET, BOOLEAN, DATETIME_FORMAT, DEFAULT_ENCODING,
-    LIST, MAP, NULL, NUMBER, NUMBER_SET, STRING, STRING_SET
-)
-from pynamodb.exceptions import AttributeDeserializationError, AttributeNullError
+from pynamodb.constants import BINARY
+from pynamodb.constants import BINARY_SET
+from pynamodb.constants import BOOLEAN
+from pynamodb.constants import DATETIME_FORMAT
+from pynamodb.constants import DEFAULT_ENCODING
+from pynamodb.constants import LIST
+from pynamodb.constants import MAP
+from pynamodb.constants import NULL
+from pynamodb.constants import NUMBER
+from pynamodb.constants import NUMBER_SET
+from pynamodb.constants import STRING
+from pynamodb.constants import STRING_SET
+from pynamodb.exceptions import AttributeDeserializationError
+from pynamodb.exceptions import AttributeNullError
 from pynamodb.expressions.operand import Path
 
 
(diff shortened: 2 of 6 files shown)


pynamodb/attributes.py
```diff
@@ -372,6 +382,43 @@ def _deserialize(self, attribute_values: Dict[str, Dict[str, Any]]) -> None:
                 value = attr.deserialize(attr.get_value(attribute_value))
                 setattr(self, name, value)
 
+    def to_json(self) -> str:
```
ikonst, line 385: Does it produce a JSON in the DynamoDB format of `{S: "foo"}`? From the tests it doesn't seem this way. Maybe it'll be beneficial, and also help with code reuse?

jpinner-lyft: no, this renders primitive values -- i'm not sure how useful serializing with the dynamo types would be -- and of course if the user wants that they can already do that with:

```
json.dumps(AttributeContainer._serialize(model))
```

My thinking here is that we already support serializing to dynamo types.
If you want a python dictionary with the python values themselves you can get that via `item.attribute_values`.
If you want a python dictionary with the serialized values (so you get JSON primitives) you can now do that via
`json.loads(item.to_json())`

jpinner-lyft:
splitting backup/restore removal into its own PR: https://github.com/pynamodb/PynamoDB/pull/858

docs/release_notes.rst
```diff
@@ -21,11 +21,14 @@ Items written using other formats must be rewritten before upgrading.
 
 ** Model Serialization **
 
-The ``Model._serialize`` method has changed and now only returns a dictionary of the DynamoDB attribute values.
+THe ``Model`` class now includes public methods for serializing and deserializing its attributes.
```
jmphilli, line 24:
```suggestion
The ``Model`` class now includes public methods for serializing and deserializing its attributes.
```

jmphilli (approved):

jpinner-lyft:
moved public serialization change to https://github.com/pynamodb/PynamoDB/pull/860

jpinner-lyft:
(Note to self to cross-reference issue #152)

jpinner-lyft:
Open question: should we make `to_json`/`from_json` methods on `AttributeContainer` or on `Model`?

garrettheel:
> Open question: should we make `to_json`/`from_json` methods on `AttributeContainer` or on `Model`?

The only reason I could think to put it on `AttributeContainer` is to be consistent for `MapAttribute`, but I can't think of a good reason you'd need to use these for that. Even if you did, we still have `MapAttribute.as_dict` + `json.dumps` as another way

garrettheel (approved):

pynamodb/models.py
```diff
@@ -1110,6 +1112,14 @@ def deserialize(self, attribute_values: Dict[str, Dict[str, Any]]) -> None:
         """
         return self._container_deserialize(attribute_values=attribute_values)
 
+    def to_json(self) -> str:
+        return json.dumps({k: attribute_value_to_json(v) for k, v in self.serialize().items()})
```
oak-shield, line 1116: UTF8 fails to this dumps

oak-shield (commented):
Encoding error

jovana:
How can I use to_json is version 6.0.0?

```
customer_response = CustomersDatabaseModel.customer_id_index.query(
    customer_id,
    filter_condition=CustomersDatabaseModel.vendor_id == current_customer["customer_id"],
)

for item in customer_response:
    customer_details = item.to_json
```

AttributeError: 'CustomersDatabaseModel' object has no attribute 'to_json'
also tried to_json()

dleung-bao:
I am also seeing the same problem as jovana in 6.0.0 that was not present before.  I just rolled back to 5.5.0, and `to_json()` is working for me.

jovana:
I found the solution for this. This function has been renamed to to_simple_json(). 

https://github.com/pynamodb/PynamoDB/commit/671cb5176fad2b66ae017c7df50160d9f7801dcd

ikonst:
Thanks, @jovana. I'll update the docs.

PR #858, opened 2020-10-01
Remove table backup/restore functionality.

jpinner-lyft:
The backup and restore methods are not scalable and store data in a custom format:
`(<serialized-hash-key>, {"range_key": <serialized-range-key>, "attributes": <dynamodb-attribute-values>})`


tests/test_discriminator.py
```diff
@@ -45,14 +45,12 @@ def test_serialize(self):
         dtm.value = StringValue(name='foo', value='Hello')
         dtm.values = [NumberValue(name='bar', value=5), RenamedValue(name='baz', value='World')]
         assert dtm._serialize() == {
```
ikonst, line 47: Should we call out that the format is changing? I know at least one piece of code which it'll disrupt (luckily I could increase the cache version there along with the PynamoDB update).

jpinner-lyft: I'll put a release note in -- I was also planning in the JSON pr to make the serialize/deserialize methods public.

pynamodb/models.py
```diff
@@ -13,7 +12,7 @@
 from pynamodb.expressions.update import Action
 from pynamodb.exceptions import DoesNotExist, TableDoesNotExist, TableError, InvalidStateError, PutError
 from pynamodb.attributes import (
-    Attribute, AttributeContainer, AttributeContainerMeta, MapAttribute, TTLAttribute, VersionAttribute
```
jmphilli, line 16: interesting to see this was unused - i wonder if we should include a pre-commit linter or something going forward

jmphilli:
+1

jmphilli (approved):

PR #861, opened 2020-10-08
Remove ModelContextManager class.

jpinner-lyft:


garrettheel (approved):
we should add something to the release notes for this too

jpinner-lyft:
added

PR #862, opened 2020-10-08
Support model class inheritance. Fixes #164

### LINKED ISSUE #164: Inheriting from Model - should this work?

jpinner-lyft:
`DefaultMeta` is removed so that a model's Meta class is inherited from the parent class.

Note: this is technically a backwards-incompatible change for those users who relied on dynamically setting attributes of their Meta class without creating it first.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, Generic, Iterable, Iterator, List, Optional, Seque
 from pynamodb.expressions.update import Action
 from pynamodb.exceptions import DoesNotExist, TableDoesNotExist, TableError, InvalidStateError, PutError
 from pynamodb.attributes import (
-    Attribute, AttributeContainer, AttributeContainerMeta, TTLAttribute, VersionAttribute
+    AttributeContainer, AttributeContainerMeta, TTLAttribute, VersionAttribute
 )
 from pynamodb.connection.table import TableConnection
 from pynamodb.expressions.condition import Condition
@@ -151,10 +151,6 @@ class BatchWrite(Generic[_T]):
             unprocessed_items = data.get(UNPROCESSED_ITEMS, {}).get(self.model.Meta.table_name)
 
 
-class DefaultMeta(object):
-    pass
-
-
 class MetaModel(AttributeContainerMeta):
     table_name: str
     read_capacity_units: Optional[int]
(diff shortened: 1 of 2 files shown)


jpinner-lyft:
reattempt of https://github.com/pynamodb/PynamoDB/pull/412

pynamodb/models.py
```diff
@@ -151,10 +151,6 @@ def commit(self) -> None:
             unprocessed_items = data.get(UNPROCESSED_ITEMS, {}).get(self.model.Meta.table_name)
 
 
-class DefaultMeta(object):
```
jpinner-lyft, line 154: this is leftover from when default values were stored in this object

jpinner-lyft:
Question from the previous attempt: say I have

```
ModelA(Model):
    class Meta:
        table_name = 'foo'

ModelB(ModelA):
    pass
```
with this change, ModelB.Meta.table_name == 'foo'

Now if i do ModelA.Meta.table_name = 'bar' what should ModelB.Meta.table_name be?
Or vice-versa, if i do ModelB.Meta.table_name = 'bar' should that change ModelA?

Do I need to make Meta some sort of "inherited, copy-on-write-per-property" class?

My current thoughts on this - subclassing will be initially used to support polymorphism where the parent model and its children classes all point to the same table so modifying the Meta class on any of the classes should modify it on all of them (they all reference the same configuration).

jpinner-lyft:
Note: this PR does not fix #361 since connections are still cached on the Model object. A follow-up change to cache connections on the Meta class will address that issue.

garrettheel (approved):

PR #863, opened 2020-10-08
Partial support for model inheritence backed by different tables. Fixes #361

### LINKED ISSUE #361: Support for inheritance between models

jpinner-lyft:
This is very much an incomplete hack. This will support the use case in #361 for models with different tables;
however, inheritance of the same table but with other attributes differing (e.g. region) are still not supported.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -1009,6 +1009,8 @@ class Model(AttributeContainer, metaclass=MetaModel):
                     cls.__module__, cls.__name__,
                 ),
             )
+        # For now we just check that the connection exists and (in the case of model inheritance)
+        # points to the same table. In the future we should update the connection if any of the attributes differ.
         if cls._connection is None or cls._connection.table_name != cls.Meta.table_name:
             cls._connection = TableConnection(cls.Meta.table_name,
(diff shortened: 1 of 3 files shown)


pynamodb/models.py
```diff
@@ -1009,7 +1009,7 @@ def _get_connection(cls) -> TableConnection:
                     cls.__module__, cls.__name__,
                 ),
             )
-        if cls._connection is None:
+        if cls._connection is None or cls._connection.table_name != cls.Meta.table_name:
```
garrettheel, line 1012: Worth leaving a comment that the connection should be recreated (or updated) if any of these values change? It's also fine to do this in a future PR

garrettheel (approved):

Issue #328, opened 2017-08-14
No support for descrimantor maps


timbrownsf:
There's no way currently to have a MapAttribute on Model be castable into one of multiple subclass types. This functionality would be great!

PR #864, opened 2020-10-09
Support model polymorphism using discriminators. Fixes #247, #328

### LINKED ISSUE #247: Polymorphism?
Can dynamodb and pynamodb do polymorphism?
I have searched can't find any example

class BaseModel(Model):
  class meta:
    table_name = "BaseTable"
    ...
  type = UnicodeAttribute(range_key=True)

class Child1Model(BaseModel):
  type = UnicodeAttribute(range_key=True)
  child1data.....

class Child2Model(BaseModel):
  type = UnicodeAttribute(range_key=True)
  child2data....

I tried it wouldn't let me do this. even if I added class meta, i am not sure how it would work.
any idea? thanks.

jpinner-lyft:
Reads are modified to ensure that reads from DynamoDB return instances of the class being queried.


pynamodb/attributes.py
```diff
@@ -372,6 +370,22 @@ def deserialize(self, attribute_values: Dict[str, Dict[str, Any]]) -> None:
                 value = attr.deserialize(attr.get_value(attribute_value))
                 setattr(self, name, value)
 
+    @classmethod
+    def _instantiate(cls: Type[_ACT], attribute_values: Dict[str, Dict[str, Any]]) -> _ACT:
+        discriminator_attr = cls._get_discriminator_attribute()
+        if discriminator_attr:
+            discriminator_attribute_value = attribute_values.pop(discriminator_attr.attr_name, None)
+            if discriminator_attribute_value:
```
hallie, line 378: Why would we allow someone to have a discriminator attribute with no value? (I see this is just being moved from down below, so sorry if this has been answered before)

docs/polymorphism.rst
```diff
@@ -49,3 +48,30 @@ A class can also be registered manually:
...
+        foo = UnicodeAttribute()
+
+    class BarModel(ParentModel, discriminator='Bar'):
+        bar = UnicodeAttribute()
+
+.. note::
+
+    Read operations that are performed on a class that has a discriminator value are slightly modified to ensure that only instances of the class are returned.
+    Query and scan operations transparently add a filter condition to ensure that only items with a matching discriminator value are returned.
+    Get and batch get operations will raise a ``ValueError`` if the returned item(s) are not a subclass of the model being read.
```
jmphilli, line 77: can you add some actual serialized dynamo data here - i was confused about how the discriminator attr was being used until i read your tests below
specifically - the discriminator attribute has the value of `foo` or `bar` which tells us which one to deserialize

this makes a ton of sense based on .... everything
but i feel like the example you've got here could be more clear - maybe change the class names or just add sample serialized documents (like the tests) to make it dumb obvious

jmphilli: if the only thing i read was this doc (not the code) i would do it wrong

pynamodb/attributes.py
jpinner-lyft: My common use case is for an abstract base class. See the TypedValue class in test_discriminator.py

docs/polymorphism.rst
jmphilli: awesome thank you

jmphilli (approved):

pynamodb/models.py
```diff
@@ -5,9 +5,28 @@
...
+from typing import Tuple
+from typing import Type
+from typing import TypeVar
+from typing import Union
+from typing import cast
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
```
garrettheel, line 29: I get `ModuleNotFoundError: No module named 'typing_extensions'` on python3.6 without explicitly installing `typing-extensions`. If we want to do this, I think we'd have to add it to `setup.py` as required

jpinner-lyft: i added it to install_requires below

garrettheel: Somehow missed this, my bad!

garrettheel (approved):

PR #869, opened 2020-10-14
Allow VersionAttribute to operate on items with a version of zero

bobwanglyft:
This PR will allow PynamoDB to read & update historical data (persisted using another python client library) whose initial version value is less than 1. 

Currently, `version = 0` is [treated the same](https://github.com/pynamodb/PynamoDB/blob/b42c326529f9043bafccf334eaa0c16ac610d795/pynamodb/models.py#L939) as `version = None`: PynamoDB will add [does_not_exist condition check](https://github.com/pynamodb/PynamoDB/blob/b42c326529f9043bafccf334eaa0c16ac610d795/pynamodb/models.py#L948) when persisting to DynamoDB and this leads to `ConditionalCheckFailedException` as the item already exists.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -936,7 +936,7 @@ class Model(AttributeContainer, metaclass=MetaModel):
         version_attribute = self.get_attributes()[self._version_attribute_name]
         version_attribute_value = getattr(self, self._version_attribute_name)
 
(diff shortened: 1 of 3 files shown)


garrettheel (approved):

HenryYYang:
👍

PR #880, opened 2020-11-04
Fix item refresh when using model discriminators. Fixes #879

### LINKED ISSUE #879: Issues with refresh() and update() when using Model Discriminators
Small contained test case. I am using dynamodb-local so I hope that's not the problem...

```
    DiscriminatorAttribute,
    UnicodeAttribute,
)

class ParentModel(Model):
    class Meta:
        table_name = 'polymorphic_table'
        write_capacity_units = 1
        read_capacity_units = 1

    cls = DiscriminatorAttribute()

class FooModel(ParentModel, discriminator='Foo'):
    foo = UnicodeAttribute()

class BarModel(ParentModel, discriminator='Bar'):
    bar = UnicodeAttribute()

# dynamodb-local
ParentModel.Meta.host = 'http://172.17.0.4:8000'

if ParentModel.exists():
    ParentModel.delete_table()
ParentModel.create_table(wait=True)

bar = BarModel(id='Hello', bar='World!')

bar.save()
bar.refresh() # boom
```

jpinner-lyft:
Allow setting discriminator attributes to the instance type to simplify `refresh` and `update` implementations.


jmphilli (approved):
should we add a test?

PR #887, opened 2020-12-07
Add operation settings to most operations

ikonst:
Most operations (except for table operations) have "operation settings" now.


garrettheel (approved):
LGTM. Would be nice to have a test that pops in an `extra_headers` operation setting

PR #904, opened 2021-02-10
Fix type errors when deriving from a MapAttribute and another type

ikonst:
Doing this:
```python
from pynamodb.attributes import MapAttribute

class MyMap(MapAttribute, object):  # <-- location of type error
    pass
```
would result in the following errors from mypy:
```
error: Definition of "__eq__" in base class "Attribute" is incompatible with definition in base class "AttributeContainer"
error: Definition of "serialize" in base class "Attribute" is incompatible with definition in base class "AttributeContainer"
```

The error is not shown in the simpler form:
```python
class MyMap(MapAttribute):
    pass
```

https://github.com/python/mypy/issues/10066

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -339,7 +339,7 @@ class AttributeContainer(metaclass=AttributeContainerMeta):
                 raise ValueError("Attribute {} specified does not exist".format(attr_name))
             setattr(self, attr_name, attr_value)
 
-    def serialize(self, null_check=True) -> Dict[str, Dict[str, Any]]:
+    def _container_serialize(self, null_check: bool = True) -> Dict[str, Dict[str, Any]]:
         """
         Serialize attribute values for DynamoDB
         """
@@ -357,7 +357,7 @@ class AttributeContainer(metaclass=AttributeContainerMeta):
                 attribute_values[attr.attr_name] = {attr.attr_type: attr_value}
         return attribute_values
 
-    def deserialize(self, attribute_values: Dict[str, Dict[str, Any]]) -> None:
+    def _container_deserialize(self, attribute_values: Dict[str, Dict[str, Any]]) -> None:
         """
         Sets attributes sent back from DynamoDB on this object
         """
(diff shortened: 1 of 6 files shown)


pynamodb/attributes.py
```diff
@@ -339,7 +339,7 @@ def _set_attributes(self, **attributes: Attribute) -> None:
                 raise ValueError("Attribute {} specified does not exist".format(attr_name))
             setattr(self, attr_name, attr_value)
 
-    def serialize(self, null_check=True) -> Dict[str, Dict[str, Any]]:
+    def _container_serialize(self, null_check: bool = True) -> Dict[str, Dict[str, Any]]:
```
ikonst, line 342: just ended up renaming it, since otherwise something like MapAttribute that derives from both Attribute and AttributeContainer ends up getting two incompatible `serialize` signatures

pynamodb/models.py
```diff
@@ -1120,6 +1120,18 @@ def _serialize_keys(cls, hash_key, range_key=None) -> Tuple[_KeyType, _KeyType]:
             range_key = cls._range_key_attribute().serialize(range_key)
         return hash_key, range_key
 
+    def serialize(self, null_check=True) -> Dict[str, Dict[str, Any]]:
```
garrettheel, line 1123:
```suggestion
    def serialize(self, null_check: bool = True) -> Dict[str, Dict[str, Any]]:
```

pynamodb/attributes.py
```diff
@@ -835,14 +827,14 @@ def _update_attribute_paths(self, path_segment):
...
 
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> 'Comparison':  # type: ignore[override]
         if self._is_attribute_container():
-            return AttributeContainer.__eq__(self, other)
+            return self is other  # type: ignore
         return Attribute.__eq__(self, other)
 
-    def __ne__(self, other):
+    def __ne__(self, other: Any) -> 'Comparison':  # type: ignore[override]
```
garrettheel, line 835: Do you think it's worth adding mypy tests for this?

garrettheel (approved):

pynamodb/attributes.py
ikonst: Yup! Added.

PR #905, opened 2021-02-11
Do not serialize all attributes for updates and deletes

ikonst:
Not only it is more efficient not to serialize attributes you don't need, it also manifested in an issue where an `update` operation would fail null-checks on attributes that are not even affected by the update -- see added `test_update_doesnt_do_validation_on_null_attributes` test.

p.s. this uncovered another bug where we fail to propagate `null_checks=True` through MapAttributes, but we'll address it separately

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -397,7 +397,7 @@ class Model(AttributeContainer, metaclass=MetaModel):
         :raises pynamodb.exceptions.DeleteError: If the record can not be deleted
         """
         hk_value, rk_value = self._get_hash_range_key_serialized_values()
-        version_condition = self._get_version_attribute_condition()
+        version_condition = self._handle_version_attribute()
         if version_condition is not None:
             condition &= version_condition
 
@@ -417,7 +417,7 @@ class Model(AttributeContainer, metaclass=MetaModel):
             raise TypeError("the value of `actions` is expected to be a non-empty list")
 
         hk_value, rk_value = self._get_hash_range_key_serialized_values()
-        version_condition = self._get_version_attribute_condition_for_update(actions)
+        version_condition = self._handle_version_attribute(actions=actions)
         if version_condition is not None:
             condition &= version_condition
 
@@ -433,13 +433,11 @@ class Model(AttributeContainer, metaclass=MetaModel):
         """
         Save this object to dynamodb
         """
-        args, kwargs = self._get_save_args()
-        version_condition = self._get_version_attribute_condition_for_save(kwargs)
-        if version_condition is not None:
-            condition &= version_condition
-        kwargs['condition'] = condition
-        kwargs['settings'] = settings
-        data = self._get_connection().put_item(*args, **kwargs)
+        args, kwargs = self._get_save_args(condition=condition)
+        print('x' * 80)
+        print(args, kwargs)
+        print('x' * 80)
+        data = self._get_connection().put_item(*args, settings=settings, **kwargs)
         self.update_local_version_attribute()
         return data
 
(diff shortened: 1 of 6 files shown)


pynamodb/models.py
```diff
@@ -946,18 +948,49 @@ def _get_save_args(self, attributes=True, null_check=True):
...
         args = (hash_key, )
+        kwargs = {}
         if range_key is not None:
             kwargs['range_key'] = range_key
-        if attributes:
-            kwargs['attributes'] = attribute_values
+        kwargs['attributes'] = attribute_values
         return args, kwargs
 
-    def _handle_version_attribute(self, serialized_attributes, actions=None):
```
ikonst, line 955: break into 3 variants even at the cost of slight repetition

pynamodb/models.py
```diff
@@ -474,34 +461,51 @@ def refresh(self, consistent_read: bool = False, settings: OperationSettings = O
             raise ValueError("Cannot refresh this item from the returned class: {}".format(stored_cls.__name__))
         self.deserialize(item_data)
 
-    def get_operation_kwargs_from_instance(
```
ikonst, line 477: break into 3 variants even at the cost of slight repetition

pynamodb/models.py
```diff
@@ -928,16 +932,14 @@ def _get_indexes(cls):
                     cls._indexes['local_secondary_indexes'].append(idx)
         return cls._indexes
 
-    def _get_save_args(self, attributes=True, null_check=True):
```
ikonst, line 931: The `attributes` flag is no longer needed (always True).

pynamodb/models.py
```diff
@@ -946,43 +935,57 @@ def _get_save_args(self, attributes=True, null_check=True):
...
-            kwargs['attributes'] = attribute_values
+        version_condition = self._handle_version_attribute(attributes=attribute_values)
+        if version_condition is not None:
+            condition &= version_condition
+        kwargs['attributes'] = attribute_values
+        kwargs['condition'] = condition
         return args, kwargs
 
-    def _handle_version_attribute(self, serialized_attributes, actions=None):
+    def _get_hash_range_key_serialized_values(self) -> Tuple[Any, Optional[Any]]:
```
ikonst, line 948: This method replaces _get_save_args for non-save scenarios by serializing only the HK and RK.

tests/integration/model_integration_test.py
```diff
@@ -55,9 +55,9 @@ class Meta:
         scores = NumberSetAttribute()
         version = VersionAttribute()
 
-    if not TestModel.exists():
-        print("Creating table")
-        TestModel.create_table(read_capacity_units=1, write_capacity_units=1, wait=True)
+    if TestModel.exists():
```
ikonst, line 58: Clean table between tests

pynamodb/models.py
```diff
@@ -474,35 +457,42 @@ def refresh(self, consistent_read: bool = False, settings: OperationSettings = O
...
-            actions=actions,
-            condition=condition,
-            return_values_on_condition_failure=return_values_on_condition_failure
-        )
-        if not is_update:
-            kwargs.update(save_kwargs)
-        elif 'range_key' in save_kwargs:
-            kwargs['range_key'] = save_kwargs['range_key']
-        return self._get_connection().get_operation_kwargs(*args, **kwargs)
+        return self._get_connection().get_operation_kwargs(hk_value, range_key=rk_value, key=KEY, actions=actions, condition=condition, return_values_on_condition_failure=return_values_on_condition_failure)
```
garrettheel, line 472: I've really been meaning to setup black to lint 😅

pynamodb/models.py
```diff
@@ -946,43 +935,57 @@ def _get_save_args(self, attributes=True, null_check=True):
...
+        if version_condition is not None:
+            condition &= version_condition
+        kwargs['attributes'] = attribute_values
+        kwargs['condition'] = condition
         return args, kwargs
 
-    def _handle_version_attribute(self, serialized_attributes, actions=None):
+    def _get_hash_range_key_serialized_values(self) -> Tuple[Any, Optional[Any]]:
+        if self._hash_keyname is None:
+            raise Exception("The model has no hash key")
```
garrettheel, line 950: Seems like precedent is a `ValueError` - https://github.com/pynamodb/PynamoDB/blob/6d6a8be7a25abf6f0ab5df7d8697c9fe871f7c4c/pynamodb/models.py#L304

garrettheel (approved):

pynamodb/models.py
ikonst: Considered that, but that actually relates to an argument that was passed in.

also lol `{range_key}`

pynamodb/models.py
ikonst: So I'll withhold formatting cause there's no point.

ikonst:
🚀

pynamodb/models.py
garrettheel: exactly

PR #917, opened 2021-03-10
Add some niceties from the dict API to MapAttribute

jlucier:
## Intro
This PR aims to add some of the functionality of `dict` objects to `MapAttribute` for convenience when manipulating model instances. Admittedly, I have not been a PynamoDB user for long, but I thought this would be a good addition.

I came across #156, which seems to be obsolete now, dead, and does not implement what I'm after here.

## What is added to `MapAttribute`
When the `MapAttribute` is acting as an `AttributeContainer`, you may call:
- `get`
- `items`
- `keys`
- `values`

These behave exactly like their equivalents on `dict` objects

## Follow ons?
I chose to steer clear of the parts of `dict` API that mutate the underlying dict, as I'm sure implementing those well would require a bit more familiarity with the existing implementation. I would welcome opinions or input on how these might be added, however I would be fine leaving them out:
- `pop`
- `update`
- `setdefault`
- `__delitem__`

Feedback welcome, also excuse any code style / type hinting no-nos.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -13,7 +13,10 @@ from datetime import timedelta
 from datetime import timezone
 from inspect import getfullargspec
 from inspect import getmembers
-from typing import Any, Callable, Dict, Generic, List, Mapping, Optional, TypeVar, Type, Union, Set, overload
+from typing import (
+    Any, Callable, Dict, Generic, ItemsView, KeysView, List, Mapping,
+    Optional, Set, TypeVar, Type, Union, ValuesView, overload
+)
 from typing import TYPE_CHECKING
 
 from pynamodb._compat import GenericMeta
@@ -983,22 +986,22 @@ class MapAttribute(Attribute[Mapping[_KT, _VT]], AttributeContainer):
 
     def get(self, item: _KT, default: Any = None) -> Union[_VT, None]:
         if not self._is_attribute_container():
-            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.get")
+            raise TypeError("MapAttribute must be acting as attribute container to use MapAttribute.get")
         return self.attribute_values.get(item, default)
 
-    def items(self):
+    def items(self) -> ItemsView[_KT, _VT]:
         if not self._is_attribute_container():
-            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.items")
+            raise TypeError("MapAttribute must be acting as attribute container to use MapAttribute.items")
         return self.attribute_values.items()
 
-    def keys(self):
+    def keys(self) -> KeysView[_KT]:
         if not self._is_attribute_container():
-            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.keys")
+            raise TypeError("MapAttribute must be acting as attribute container to use MapAttribute.keys")
         return self.attribute_values.keys()
 
-    def values(self):
+    def values(self) -> ValuesView[_VT]:
         if not self._is_attribute_container():
-            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.values")
+            raise TypeError("MapAttribute must be acting as attribute container to use MapAttribute.values")
         return self.attribute_values.values()
 
 
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -983,22 +983,22 @@ class MapAttribute(Attribute[Mapping[_KT, _VT]], AttributeContainer):
 
     def get(self, item: _KT, default: Any = None) -> Union[_VT, None]:
         if not self._is_attribute_container():
-            raise AssertionError("MapAttribute must be acting as attribute container to MapAttribute.get")
+            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.get")
         return self.attribute_values.get(item, default)
 
     def items(self):
         if not self._is_attribute_container():
-            raise AssertionError("MapAttribute must be acting as attribute container to MapAttribute.items")
+            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.items")
         return self.attribute_values.items()
 
     def keys(self):
         if not self._is_attribute_container():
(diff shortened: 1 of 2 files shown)


pynamodb/attributes.py
```diff
@@ -979,6 +979,28 @@ def as_dict(self):
             result[key] = value.as_dict() if isinstance(value, MapAttribute) else value
         return result
 
+    # some of the dict API
+
+    def get(self, item: _KT, default: Any = None) -> Union[_VT, None]:
+        if not self._is_attribute_container():
+            raise AssertionError("MapAttribute must be acting as attribute container to MapAttribute.get")
```
ikonst, line 986: Perhaps use TypeError like in similar cases in this class?

ikonst:
1. Given that it's a thin wrapper around `attribute_values` which is already public, I wonder if there's enough motivation to add those.
2. Would the newly added `get` method do what you'd expect for non-raw maps? We already take a different path for non-raw maps in the `__getitem__` implementation.

pynamodb/attributes.py
jlucier: Will do

jlucier:
1. My rationale was that since `MapAttribute` is already _pretty close_ to a `dict` in a few ways, it would be nice to take it slightly further so the user doesn't have to remember "oh, I can access this using `[...]` but I can't call `get` so I need to use `.attribute_values.get(...)` for this one"
2. I'm not sure I fully understand when one would hit the raw pathway in `__getitem__`, so I figured I would avoid the case by raising an error if the `MapAttribute` isn't acting as an `AttributeContainer`. If desired, `get` could be made to behave more similarly to `__getitem__` with the following implementation instead:
```python
def get(self, item: _KT, default: Any = None) -> Union[_VT, None]:
    try:
        return self[item]
    except AttributeError:
        return default
```

ikonst:
I wish we document model vs. raw MapAttributes better. Basically:
```
class MyMap(MapAttribute):
   foo = UnicodeAttribute()

class MyModel(Model):
   my_map = MyMap()
   my_raw_map = MapAttribute()
```

jlucier:
Using your example classes, this is really all I'm hoping to accomplish:
```python
m = MyModel(my_map={"foo": "str"}, my_raw_map={"test": 123})

# unchanged functionality
m.my_raw_map.something => AttributeError
m.my_raw_map.test => 123
m.my_raw_map["test"] => 123

# nice to have
m.my_raw_map.get("something") => None # as opposed to error

# works, but less nice
m.my_raw_map.attribute_values.get("something") => None # as opposed to error
```

The same would hold true of the "non-raw" `MapAttribute`.

ikonst:
In the non-raw map, are the values you're returning before or after serialization?

jlucier:
Everything is before serialization (if I understand your question properly). It's the same as what you would get by using `m.my_map.<some attr>`.

For instance (from my tests):
```python
class NestedThing(MapAttribute):
    double_nested = MapAttribute()
    double_nested_renamed = MapAttribute(attr_name='something_else')

class ThingModel(Model):
    key = NumberAttribute(hash_key=True)
    nested = NestedThing()

t = ThingModel(nested=NestedThing(
    double_nested={'hello': 'world'},
    double_nested_renamed={'foo': 'bar'})
)

assert t.nested.get("double_nested") == t.nested.double_nested
```
Both `t.nested.double_nested` and `t.nested.get("double_nested")` are instances of `MapAttribute`

pynamodb/attributes.py
```diff
@@ -979,6 +979,28 @@ def as_dict(self):
...
+        if not self._is_attribute_container():
+            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.get")
+        return self.attribute_values.get(item, default)
+
+    def items(self):
+        if not self._is_attribute_container():
+            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.items")
+        return self.attribute_values.items()
+
+    def keys(self):
```
ikonst, line 994:
```suggestion
    def keys(self) -> KeysView[_KT]:
```

pynamodb/attributes.py
```diff
@@ -979,6 +979,28 @@ def as_dict(self):
...
+        if not self._is_attribute_container():
+            raise TypeError("MapAttribute must be acting as attribute container to MapAttribute.keys")
+        return self.attribute_values.keys()
+
+    def values(self):
```
ikonst, line 999:
```suggestion
    def values(self) -> ValuesView[_VT]:
```

ikonst:
It does seem to complement the map behavior nicely, though the existing way is also not too cumbersome (through attribute_values).

I suppose that if an existing MapAttribute subclass defines attributes `keys`, `values` or `get`, they would shadow those methods, which is a good thing (otherwise it'd be a breaking change).

@garrettheel WDYT?

pynamodb/attributes.py
jlucier: I added hinting to those methods, but it seems mypy was unhappy with `_KT` matching `str`: https://github.com/pynamodb/PynamoDB/pull/917/checks?check_run_id=2086230923

Any ideas? I'm not that familiar with mypy.

ikonst: I was also surprised by there being a `_KT` parameter, since I thought it's always `str`. Have to look into this.

jlucier: Should I use `str` instead of `_KT`? That would fix the mypy tests but did not seem right to me.

garrettheel:
> It does seem to complement the map behavior nicely, though the existing way is also not too cumbersome (through attribute_values).
> 
> I suppose that if an existing MapAttribute subclass defines attributes `keys`, `values` or `get`, they would shadow those methods, which is a good thing (otherwise it'd be a breaking change).
> 
> @garrettheel WDYT?

My main hesitation would be adding more complexity/functionality here that would make it harder to do a larger overhaul to MapAttribute as we've discussed recently, though this is fairly limited in scope so doesn't worry me too much. 

The other larger question might be whether we want models/attributes to implement the dict interface or require an explicit `to_dict`

PR #934, opened 2021-05-02
improve doc string for create_table

yelizariev:
"on_demad" looks as a value that should be set, but it's not


ikonst (approved):
AWS refer to it as [On-Demand Mode](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html#HowItWorks.OnDemand) but I agree we made it sound like an API constant.

Issue #891, opened 2020-12-18
CancellationReasons are not included in the VerboseClientError


AuxinJeron:
DynamoDB API and Botocore now provides the modeled `CancellationReasons` in the transaction write error exception, but `VerboseClientError` hasn't included `CancellationReasons` yet based on https://github.com/pynamodb/PynamoDB/blob/master/pynamodb/connection/base.py#L424-L444.

jordal:
Is this now resolved by #1144?

ikonst:
Indeed. Meant to backport it to 5.x too.

PR #943, opened 2021-05-20
Add CancellationReasons in the VerboseClientError (#891)

AuxinJeron:
Note
----
This change is to add the CancellationReasons into the VerboseClientError,
which is included in the botocore TransactionCanceledException.
https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.exceptions.TransactionCanceledException

Adding this would help developers to figure out the detail cancellation reasons directly rather than parsing the
message.


ikonst:
Sorry for missing this PR for so long. This was eventually done in #1144.

PR #944, opened 2021-05-30
added new sts role configuration parameter to assume aws role dynamic…

KiraPC:
This PR is intended to add a new feature in order to allow dynamic assume role by just configuring the role-arn in the META class.

```python
class Tip(Model):
    class Meta:
        aws_sts_role_arn='arn:aws:iam::1234567:role/my-aws-role'  
        table_name = "my_table"
        region = 'eu-central-1'

```

This is a draft, to get your POV about this feature: if it is acceptable, if my code can be improved, or so on.

I will add tests and update the doc in next days.

### DIFF
--- a/docs/awsaccess.rst
+++ b/docs/awsaccess.rst
@@ -24,5 +24,19 @@ If for some reason you can't use conventional AWS configuration methods, you can
             aws_secret_access_key = 'my_secret_access_key'
             aws_session_token = 'my_session_token' # Optional, only for temporary credentials like those received when assuming a role
 
+If you need to access DynamoDB passing for a specific AWS Role and so you need to perform an assume-role operation on your Role ARN you can configure it in the Model Meta class:
+
+.. code-block:: python
+
+    from pynamodb.models import Model
+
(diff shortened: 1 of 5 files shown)


garrettheel:
Hey @KiraPC, I think this is straying a little too far into re-implementing botocore concepts. Is there a reason that https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html wouldn't work for you?

KiraPC:
Hi @garrettheel thank you.

To be honest, I didn't tested this approach. I'll try in next days.

Anyway, if you think that implementation is far to standard, close it.

andrewlytle:
> Hey @KiraPC, I think this is straying a little too far into re-implementing botocore concepts. Is there a reason that https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html wouldn't work for you?

How is this re-implementing botocore concepts? This would allow you to assume a role, and then tell Pynamo to use that role. It's not always possible to use that role from the command line.

PR #949, opened 2021-06-17
UTCDateTimeAttribute: support for pre-1000 dates

### LINKED ISSUE #899: _fast_parse_utc_date_string fails on pre year 1000
Python `datetime` supports dates back to year 0, but `_fast_parse_utc_date_string` forces datetimes to have four digit years. This resulted in the following error:

> ValueError: Datetime string '539-02-20T08:36:49.000000+0000' does not match format '%Y-%m-%dT%H:%M:%S.%f%z'

`_fast_parse_utc_date_string` should fall back to a library call if it fails.

Edit:

My bad, looks like `datetime.datetime.strptime('539-02-20T08:36:49.000000+0000', '%Y-%m-%dT%H:%M:%S.%f%z')` also fails, meaning `isoformat()` doesn't produce parseable datetimes. :/

ikonst:
Fixes #899.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -702,7 +702,7 @@ class UTCDateTimeAttribute(Attribute[datetime]):
         # This is ~5.8x faster than using strptime and 38x faster than dateutil.parser.parse.
         _int = int  # Hack to prevent global lookups of int, speeds up the function ~10%
         try:
-            # Fix pre-1000 dates serialized on systems where sfrftime doesn't pad w/older PynamoDB versions.
+            # Fix pre-1000 dates serialized on systems where strftime doesn't pad w/older PynamoDB versions.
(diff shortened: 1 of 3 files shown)


pynamodb/attributes.py
```diff
@@ -700,6 +702,8 @@ def _fast_parse_utc_date_string(date_string: str) -> datetime:
         # This is ~5.8x faster than using strptime and 38x faster than dateutil.parser.parse.
         _int = int  # Hack to prevent global lookups of int, speeds up the function ~10%
         try:
+            # Fix pre-1000 dates serialized on systems where sfrftime doesn't pad w/older PynamoDB versions.
```
jpinner-lyft, line 705: typo: s/sfrftime/strftime/

jpinner-lyft (approved):

PR #966, opened 2021-08-06
Fix type definitions

krkd:
Pull request fixes type definition for Transaction context manager and update method type definition

1. Before type definition for Transaction.__enter__ always returned Transaction type which was not working properly for TransactoinWrite class
2. update method action argument stated that it accepts List[Action]. List[] itself is invariant which was not working properly with subclasses of Action. 

PR itself does not add any new feature

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -404,7 +404,7 @@ class Model(AttributeContainer, metaclass=MetaModel):
 
         return self._get_connection().delete_item(hk_value, range_key=rk_value, condition=condition, settings=settings)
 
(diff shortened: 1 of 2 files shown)


garrettheel:
Thanks for the PR! Would you mind adding some tests in `tests/test_mypy.py` to illustrate your previous issues?

garrettheel:
Gonna close this since most of it has already been implemented in https://github.com/pynamodb/PynamoDB/pull/991. Feel free to put up a separate PR for the `Sequence` change with a test

PR #968, opened 2021-08-10
Asyncio for pynamodb

tasn:
Adding asyncio support to pynamodb.
This makes https://github.com/pynamodb/PynamoDB/pull/853 redundant.

I haven't properly tested everything, but my code seems to work. Notes:
1. For whatever reason my code's performance hasn't improved (requests per second). I wonder if there are any sync/locky things lurking anywhere.
2. The model and index functions only have async variants at the moment, I haven't given them the same `_async` suffix I gave others. This is easy to fix.
3. I made iterators async only (should be easy to make them also support the classic iterator interface).

Would love to get your thoughts on why things may not be as performant as I thought they would/should be.

CC @svix-yair, @ikonst, @garrettheel

### DIFF
--- a/pynamodb/async_util.py
+++ b/pynamodb/async_util.py
@@ -1,0 +1,19 @@
+import functools
+
+
+def run_secretly_sync_async_fn(async_fn, *args, **kwargs):
+    # From https://github.com/python-trio/hip/issues/1#issuecomment-322028457
+    coro = async_fn(*args, **kwargs)
+    try:
+        coro.send(None)
+    except StopIteration as exc:
+        return exc.value
+    else:
+        raise RuntimeError("you lied, this async function is not secretly synchronous")
+
+
+def wrap_secretly_sync_async_fn(async_fn):
+    @functools.wraps(async_fn)
+    def wrap(*args, **kwargs):
+        return run_secretly_sync_async_fn(async_fn, *args, **kwargs)
+    return wrap
--- a/pynamodb/async_util.py
+++ b/pynamodb/async_util.py
@@ -1,19 +1,9 @@
+import asyncio
 import functools
-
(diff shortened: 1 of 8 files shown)


garrettheel:
Thanks for picking this up!

I think the only way this can be mergeable is to separate out the async functionality as an optional [extra](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies). This would require some refactoring of connection-related parts to separate the current botocore sync implementation from the new  aiobotocore async one.

> Would love to get your thoughts on why things may not be as performant as I thought they would/should be.

I wouldn't expect asyncio to be inherently faster than another form of cooperative multitasking. That said, can you share the code you're using to benchmark? There are many variables such as the python version, event loop implementation, etc. to account for. I might guess that `aiobotocore`'s implementation is heavier than the urllib3-based one in botocore too, so it wouldn't surprise me if a naive version of this performed worse than sync.

tasn:
Update: I tried with a remote dynamodb and it doesn't look like it's a matter of the the local dynamodb being overloaded.
I also tried updating the `max_pool_connections` to something that's larger than 10, thinking that may be it, but it's not the case.

tasn:
I don't have a good benchmark unfortunately, I'm just trying to hammer my FastAPI server with requests. One version is using this async version, and another is using a sync implementation (and normal pynamo).

I'd have expected the number of concurrent requests it can handle to increase dramatically due to the networking code happening in parallel. Though I'm getting roughly the same results.

tasn:
OK, one last update: it looks like it's CPU bound indeed. Most of the time is spent between `_convert_to_request_dict` and `_create_prepared_request` (namely `_sign_request`).

PR #970, opened 2021-08-11
Enable retries in botocore.

tasn:
Botocore has built-in retries, we should probably use them rather than reimplement in pynamodb, no?

Any thoughts?

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -360,9 +360,10 @@ class Connection(object):
             operation_model,
         )
 
-        for i in range(0, self._max_retry_attempts_exception + 1):
+        # FIXME: actually remove the loop if we decide to keep this patch
+        for i in range(0, 1):
             attempt_number = i + 1
-            is_last_attempt_for_exceptions = i == self._max_retry_attempts_exception
+            is_last_attempt_for_exceptions = True
 
             http_response = None
             prepared_request = None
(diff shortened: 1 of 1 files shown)


ikonst:
Unless we have firm intentions of going off botocore (see #968), I'd see this as a positive change. As you can see in the docs for [Standard retry mode](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#standard-retry-mode), botocore has logic around which exceptions are transient and good candidates for retries, as well as exponential backoff.

WDYT @garrettheel?

tasn:
I'm happy to improve the patch a bit to get it "ready" (cleaning up the `for` loop), though I'd need help to make sure it *actually* retries, as I'm not sure what's the right way to be doing such testing.

ikonst:
> though I'd need help to make sure it actually retries, as I'm not sure what's the right way to be doing such testing

A unit test wouldn't need to test it, since we'd rely on botocore's own unit tests to make use its retries work well.

For an integration test... maybe write a simple proxy server with [http.server](https://docs.python.org/3/library/http.server.html#http.server.HTTPServer) that'd proxy to the underlying dynamodb-local, and make it fail the 1st request in a predictable way?

tasn:
I'm even unsure about manual testing as I don't know how a retry-worthy response looks like from dynamo.

garrettheel:
> One problem that we've noticed is that (and I'm not sure about this) - this change maybe doesn't retry in all of the same scenarios. Is there a way to sanely add tests for these things to make sure we are good?

I don't know that we need tests exactly, but I think it would be helpful to do some analysis of the botocore `standard` behavior vs. what is already happening here and document it in the PR so we can determine whether it would constitute a breaking change to anyone and handle appropriately

PR #993, opened 2021-11-09
contributing: clarify scope

ikonst:


docs/contributing.rst
```diff
@@ -61,6 +61,34 @@ Outside of data compatibility, follow the usual semver rules for API changes and
 to a major release.
 
 
+The Scope of the Library
+------------------------
+
+The purpose of this library is to provide a Pythonic ORM layer on top of DynamoDB to be used
```
jpinner-lyft, line 67: ODM not ORM

ikonst: agree, we just say "ORM-like" in the README so I thought...

jpinner-lyft: can we just change the readme as well -- we don't support any relationship management in this library

jpinner-lyft: or just drop the bullet point from the features list entirely

garrettheel (approved):

jpinner-lyft (approved):

Issue #992, opened 2021-11-09
Query on GSI not respecting polymorphism filter condition


mstykow:
Hello everyone! Great library.

The [polymorphism documentation](https://pynamodb.readthedocs.io/en/latest/polymorphism.html) states: "Query and scan operations transparently add a filter condition to ensure that only items with a matching discriminator value are returned."

However, when running a query on a GSI using a child class the filter condition does not seem to be correctly added as I still receive results matching another child class.

Setup:
```
class MyGSI(GlobalSecondaryIndex):
    class Meta:
        index_name = 'foo-index'
        read_capacity_units = 2
        write_capacity_units = 1
        projection = AllProjection()

    modelId = UnicodeAttribute(hash_key=True)

class ParentModel(Model):
    class Meta:
        table_name = 'polymorphic_table'
    id = UnicodeAttribute(hash_key=True)
    modelId = UnicodeAttribute()
    my_gsi = MyGSI()
    cls = DiscriminatorAttribute()

class FooModel(ParentModel, discriminator='Foo'):
    foo = UnicodeAttribute()

class BarModel(ParentModel, discriminator='Bar'):
    bar = UnicodeAttribute()
```

Then

```
FooModel.my_gsi.query(...)
```

will also return results from `BarModel`.

Is there a mistake in my setup or is this a bug?

ikonst:
@jpinner-lyft

jpinner-lyft:
This is a bug.

When `ParentModel` is created, the metaclass sets a property on `my_gsi.Meta`: `my_gsi.Meta.model = ParentModel`.
The GSI query actually calls `FooModel.my_gsi.Meta.model.query(...)`, so the filter condition isn't correctly applied.

We need to change the behavior of the index class to either dynamically determine the model class or copy the attribute to the child class and override the model property.

@ikonst This brings up an interesting edge case btw, what should we do if the discriminator attribute isn't projected into the index?

ikonst:
> This brings up an interesting edge case btw, what should we do if the discriminator attribute isn't projected into the index?

Maybe it's something an "audit_against_table_description" function (to be built :) should check.

PR #994, opened 2021-11-10
Initialize indexes in the model metaclass.

jpinner-lyft:
This is a preliminary change to simplify index management before fixing #992.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -203,6 +203,7 @@ class MetaModel(AttributeContainerMeta):
 
     def __init__(self, name, bases, namespace, discriminator=None) -> None:
         super().__init__(name, bases, namespace, discriminator)
+        MetaModel._initialize_indexes(self)
         cls = cast(Type['Model'], self)
         for attr_name, attribute in cls.get_attributes().items():
             if attribute.is_hash_key:
@@ -253,10 +254,6 @@ class MetaModel(AttributeContainerMeta):
                         setattr(attr_obj, 'aws_secret_access_key', None)
                     if not hasattr(attr_obj, 'aws_session_token'):
                         setattr(attr_obj, 'aws_session_token', None)
-                elif isinstance(attr_obj, Index):
-                    attr_obj.Meta.model = cls
-                    if not hasattr(attr_obj.Meta, "index_name"):
-                        attr_obj.Meta.index_name = attr_name
 
             # create a custom Model.DoesNotExist derived from pynamodb.exceptions.DoesNotExist,
             # so that "except Model.DoesNotExist:" would not catch other models' exceptions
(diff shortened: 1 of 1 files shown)


pynamodb/models.py
```diff
@@ -267,6 +264,18 @@ def __init__(self, name, bases, namespace, discriminator=None) -> None:
...
                 cls.DoesNotExist = type('DoesNotExist', (DoesNotExist, ), exception_attrs)
 
+    @staticmethod
+    def _initialize_indexes(cls):
+        """
+        Initialize indexes on the class.
+        """
+        cls._indexes = {}
+        for name, index in getmembers(cls, lambda o: isinstance(o, Index)):
+            index.Meta.model = cls
```
jpinner-lyft, line 274: This changes the current behavior where the class is set once by the parent model and not overwritten by child classes. The new (still broken) behavior is last-write-wins during child model creation.

pynamodb/models.py
```diff
@@ -267,6 +264,18 @@ def __init__(self, name, bases, namespace, discriminator=None) -> None:
                 }
                 cls.DoesNotExist = type('DoesNotExist', (DoesNotExist, ), exception_attrs)
 
+    @staticmethod
+    def _initialize_indexes(cls):
+        """
+        Initialize indexes on the class.
+        """
+        cls._indexes = {}
```
jpinner-lyft, line 272: This changes current behavior where child classes now each have their own index dictionary instead of defaulting to the dictionary on the parent class.

pynamodb/models.py
```diff
@@ -267,6 +264,18 @@ def __init__(self, name, bases, namespace, discriminator=None) -> None:
...
+        for name, index in getmembers(cls, lambda o: isinstance(o, Index)):
+            index.Meta.model = cls
+            if not hasattr(index.Meta, "index_name"):
+                index.Meta.index_name = name
+            cls._indexes[index.Meta.index_name] = index
```
jpinner-lyft, line 277: This keeps the current broken behavior where you can have index name collisions.

ikonst (approved):

pynamodb/models.py
```diff
@@ -886,38 +892,34 @@ def _get_indexes(cls):
         """
```
garrettheel, line 892: Any reason you didn't move the call to `_get_indexes()` in `create_table` over to `cls._indexes`? Not really worried about performance of that, but just for consistency

garrettheel (approved):

pynamodb/models.py
jpinner-lyft: cls._indexes no longer returns the index_data (it's not cached at all since it is only used in create table)

pynamodb/models.py
```diff
@@ -267,6 +264,19 @@ def __init__(self, name, bases, namespace, discriminator=None) -> None:
...
 
+    @staticmethod
+    def _initialize_indexes(cls):
+        """
+        Initialize indexes on the class.
+        """
+        cls._indexes = {}
+        for name, index in getmembers(cls, lambda o: isinstance(o, Index)):
+            if not hasattr(index.Meta, "model"):
+                index.Meta.model = cls
```
jpinner-lyft, line 275: This changes the current behavior where the class is set by any model that has the index in the class namespace. Now first write wins.

pynamodb/models.py
garrettheel: Whoops, missed that. Thanks

PR #995, opened 2021-11-11
Store Model class reference on an instance of index class. Fixes #992

### LINKED ISSUE #992: Query on GSI not respecting polymorphism filter condition

jpinner-lyft:
This is a breaking change to the index API. The count, query, and scan methods are now instance methods.

### DIFF
--- a/pynamodb/indexes.py
+++ b/pynamodb/indexes.py
@@ -27,6 +27,7 @@ class Index(Generic[_M]):
     Base class for secondary indexes
     """
     Meta: Any = None
+    _model: _M
 
     @classmethod
     def __init_subclass__(cls, **kwargs):
@@ -43,14 +44,13 @@ class Index(Generic[_M]):
             raise ValueError("No projection defined, define a projection for this class")
 
     def __set_name__(self, owner: Type[_M], name: str):
-        if not hasattr(self.Meta, "model"):
-            self.Meta.model = owner
+        if not hasattr(type(self), "_model"):
+            type(self)._model = owner
         if not hasattr(self.Meta, "index_name"):
             self.Meta.index_name = name
 
-    @classmethod
     def count(
-        cls,
+        self,
         hash_key: _KeyType,
         range_key_condition: Optional[Condition] = None,
         filter_condition: Optional[Condition] = None,
@@ -61,19 +61,18 @@ class Index(Generic[_M]):
         """
         Count on an index
         """
-        return cls.Meta.model.count(
+        return self._model.count(
             hash_key,
             range_key_condition=range_key_condition,
             filter_condition=filter_condition,
-            index_name=cls.Meta.index_name,
+            index_name=self.Meta.index_name,
             consistent_read=consistent_read,
             limit=limit,
             rate_limit=rate_limit,
         )
 
-    @classmethod
     def query(
-        cls,
+        self,
         hash_key: _KeyType,
         range_key_condition: Optional[Condition] = None,
         filter_condition: Optional[Condition] = None,
@@ -88,12 +87,12 @@ class Index(Generic[_M]):
         """
         Queries an index
         """
-        return cls.Meta.model.query(
+        return self._model.query(
             hash_key,
             range_key_condition=range_key_condition,
             filter_condition=filter_condition,
             consistent_read=consistent_read,
-            index_name=cls.Meta.index_name,
+            index_name=self.Meta.index_name,
             scan_index_forward=scan_index_forward,
             limit=limit,
             last_evaluated_key=last_evaluated_key,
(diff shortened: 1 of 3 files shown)


jpinner-lyft:
@ikonst @garrettheel so this "works" but breaks backwards compatibility a little bit and is certainly not elegant

jpinner-lyft:
A lot of the complexity here comes from storing the Model class reference on the index Meta class itself.
If we instead stored the reference on the instance of the index in the Model class that simplifies things slightly.
That would be a slightly larger breaking change though.

ikonst:
> That would be a slightly larger breaking change though.

In what sense? Do you think the fact that we use the `Index.Meta.model` is part of the contract somehow?

jpinner-lyft:
@ikonst right, removing the classmethod annotation and then changing how an end user could set the model reference

ikonst:
For maximum compatibility, maybe we can do some atrocity such as a method that's both class and instance:
```
class Foo:
    def bar(self=None):
        self = self or Foo
        ...
```

In the "classmethod" path, it'll keep using the old Meta.model. In the "instance" path, it'll use a newly-introduced Index instance field.

Maybe that's too crazy, but just throwing it out there.

jpinner-lyft:
@ikonst I implemented your suggestion, let me know what you think.

I had to replace `Index.Meta.model` with `Index._model` so that I could override the lookup on an instance or class basis.

Then I wrote a descriptor `classorinstancemethod` which passes in either the class or the instance as the first argument.

Finally during Model class instantiation, I store a reference to the Model on the index class if it hasn't been set to make sure `Index.query` and the like keep working while creating a local class to store a reference to the current model class so that subclassing/polymorphism still works.

pynamodb/indexes.py
```diff
@@ -48,16 +48,25 @@ class Index(Generic[_M], metaclass=IndexMeta):
...
+    _model: _M
 
     def __init__(self) -> None:
         if self.Meta is None:
             raise ValueError("Indexes require a Meta class for settings")
         if not hasattr(self.Meta, "projection"):
             raise ValueError("No projection defined, define a projection for this class")
 
-    @classmethod
+    class classorinstancemethod:
```
ikonst, line 59: nit - if we're keeping it in this file, I'd move it to module scope and prefix with underscore. But maybe move it to pynamodb/_compat.py?

pynamodb/indexes.py
```diff
@@ -48,16 +48,25 @@ class Index(Generic[_M], metaclass=IndexMeta):
...
+        def __init__(self, func):
+            self.func = func
+        def __get__(self, instance, owner):
+            def _method(*args, **kwargs):
+                return self.func(instance or owner, *args, **kwargs)
+            return _method
```
ikonst, line 65:
```suggestion
            return functools.partial(self.func, instance or owner)
```

pynamodb/indexes.py
```diff
@@ -48,16 +48,25 @@ class Index(Generic[_M], metaclass=IndexMeta):
...
+
+    @classorinstancemethod
     def count(
-        cls,
+        obj,
```
ikonst, line 69:
```suggestion
        obj: Union[Index, Type[Index]],
```
?

garrettheel:
@jpinner-lyft we're looking at bumping major versions for #1003, do you want to get this in before that? Otherwise I'm fine with going for the breaking change here and making it available from v6 onwards

cc @ikonst

pynamodb/indexes.py
```diff
@@ -43,14 +44,13 @@ def __init__(self) -> None:
             raise ValueError("No projection defined, define a projection for this class")
 
     def __set_name__(self, owner: Type[_M], name: str):
-        if not hasattr(self.Meta, "model"):
-            self.Meta.model = owner
+        if not hasattr(type(self), "_model"):
```
garrettheel, line 47: Do we still need this? Why wouldn't `_initialize_indexes()` set it?

jpinner-lyft: good catch -- i don't think we do now since we never need to access the variable on a class

garrettheel (approved):

PR #996, opened 2021-11-12
Always use Model attribute definitions in create table schema.

jpinner-lyft:
Currently we require indexes to define the hash and range keys using attributes that hopefully match the model definition.
This change begins to unwind this dependency by using the Model attribute definitions to generate the create table schema.

### DIFF
--- a/pynamodb/indexes.py
+++ b/pynamodb/indexes.py
@@ -8,7 +8,8 @@ from typing import TYPE_CHECKING
 from pynamodb._compat import GenericMeta
 from pynamodb.constants import (
     INCLUDE, ALL, KEYS_ONLY, ATTR_NAME, ATTR_TYPE, KEY_TYPE, KEY_SCHEMA,
-    ATTR_DEFINITIONS, META_CLASS_NAME
+    ATTR_DEFINITIONS, META_CLASS_NAME, PROJECTION_TYPE, NON_KEY_ATTRIBUTES,
+    PAY_PER_REQUEST_BILLING_MODE, READ_CAPACITY_UNITS, WRITE_CAPACITY_UNITS,
 )
 from pynamodb.attributes import Attribute
 from pynamodb.expressions.condition import Condition
@@ -151,27 +152,27 @@ class Index(Generic[_M], metaclass=IndexMeta):
         """
         Returns the schema for this index
         """
-        attr_definitions = []
-        schema = []
+        schema = {
+            'index_name': cls.Meta.index_name,
+            'key_schema': [],
+            'projection': {
+                PROJECTION_TYPE: cls.Meta.projection.projection_type,
+            },
+        }
         for attr_name, attr_cls in cls._get_attributes().items():
-            attr_definitions.append({
-                'attribute_name': attr_cls.attr_name,
-                'attribute_type': attr_cls.attr_type
-            })
             if attr_cls.is_hash_key:
-                schema.append({
+                schema['key_schema'].append({
                     ATTR_NAME: attr_cls.attr_name,
                     KEY_TYPE: HASH
                 })
             elif attr_cls.is_range_key:
-                schema.append({
+                schema['key_schema'].append({
                     ATTR_NAME: attr_cls.attr_name,
                     KEY_TYPE: RANGE
                 })
-        return {
-            'key_schema': schema,
-            'attribute_definitions': attr_definitions
-        }
+        if cls.Meta.projection.non_key_attributes:
+            schema['projection'][NON_KEY_ATTRIBUTES] = cls.Meta.projection.non_key_attributes
+        return schema
 
     @classmethod
     def _get_attributes(cls):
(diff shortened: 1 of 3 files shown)


pynamodb/indexes.py
```diff
@@ -151,27 +152,27 @@ def _get_schema(cls) -> Dict:
...
         """
-        attr_definitions = []
-        schema = []
+        schema = {
+            'index_name': cls.Meta.index_name,
+            'key_schema': [],
+            'projection': {
+                PROJECTION_TYPE: cls.Meta.projection.projection_type,
+            },
+        }
```
jpinner-lyft, line 161: This now matches the index schema for the dynamodb create table api:
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateTable.html#DDB-CreateTable-request-LocalSecondaryIndexes

pynamodb/indexes.py
```diff
@@ -189,7 +190,16 @@ class GlobalSecondaryIndex(Index[_M]):
...
+
+    @classmethod
+    def _get_schema(cls) -> Dict:
+        schema = super()._get_schema()
+        if getattr(cls.Meta, 'billing_mode', None) != PAY_PER_REQUEST_BILLING_MODE:
+            schema['provisioned_throughput'] = {
+                READ_CAPACITY_UNITS: cls.Meta.read_capacity_units,
+                WRITE_CAPACITY_UNITS: cls.Meta.write_capacity_units
+            }
+        return schema
```
jpinner-lyft, line 202: Here we add the GSI specific "ProvisionedThroughput" setting:
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateTable.html#DDB-CreateTable-request-GlobalSecondaryIndexes

pynamodb/models.py
```diff
@@ -886,41 +879,24 @@ def _get_schema(cls) -> Dict[str, Any]:
...
+                      for index_schema in chain(schema['global_secondary_indexes'], schema['local_secondary_indexes'])
+                      for key_schema in index_schema['key_schema']}
+        attr_keys = {attr.get('attribute_name') for attr in schema['attribute_definitions']}
+        for attr_name in attr_names:
+            if attr_name not in attr_keys:
+                attr_cls = cls.get_attributes()[cls._dynamo_to_python_attr(attr_name)]
+                schema['attribute_definitions'].append({
+                    'attribute_name': attr_cls.attr_name,
+                    'attribute_type': attr_cls.attr_type
+                })
```
jpinner-lyft, line 898: This is now where we iterate through the attributes used in the index and add them to the attribute definitions.

pynamodb/models.py
```diff
@@ -886,41 +879,24 @@ def _get_schema(cls) -> Dict[str, Any]:
...
-                idx['projection'][NON_KEY_ATTRIBUTES] = index.Meta.projection.non_key_attributes
-            if isinstance(index, GlobalSecondaryIndex):
-                index_data['global_secondary_indexes'].append(idx)
+                schema['global_secondary_indexes'].append(index_schema)
             else:
-                index_data['local_secondary_indexes'].append(idx)
-        return index_data
+                schema['local_secondary_indexes'].append(index_schema)
+        attr_names = {key_schema[ATTR_NAME]
+                      for index_schema in chain(schema['global_secondary_indexes'], schema['local_secondary_indexes'])
```
ikonst, line 889:
```suggestion
                      for index_schema in(*schema['global_secondary_indexes'], *schema['local_secondary_indexes'])
```

pynamodb/indexes.py
```diff
@@ -151,27 +152,27 @@ def _get_schema(cls) -> Dict:
...
+            },
+        }
         for attr_name, attr_cls in cls._get_attributes().items():
-            attr_definitions.append({
-                'attribute_name': attr_cls.attr_name,
-                'attribute_type': attr_cls.attr_type
-            })
             if attr_cls.is_hash_key:
-                schema.append({
+                schema['key_schema'].append({
```
ikonst, line 164: Kinda like it better before, when it was less mutate-y.

jpinner-lyft: So before we append to a local list and then set the list on the returned dictionary.
Here we just switch the order so we set the list and then append to it.
I can certainly re-introduce the local variable and then assign if that is preferable.

pynamodb/models.py
jpinner-lyft: is the second one more "pythonic"?

ikonst: 🤷  it's a Python 3.5 thing so we didn't have it before (though `itertools.chain` still has one thing on this syntax, in that you can do `chain(*list_of_lists)`)

pynamodb/indexes.py
ikonst: All good. I guess dict lookups always make me slightly nervous.

ikonst (approved):

PR #997, opened 2021-11-16
Use PEP-487 __set_name__ to simplify the Model and Index metaclasses.

jpinner-lyft:
Set attribute names using the python 3.6+ `__set_name__` function that is called during AttributeContainer class creation.
This simplifies MapAttribute classes slightly since we can check once if an instance is an AttributeContainer class attribute.


jpinner-lyft:
@ikonst @garrettheel This is the first in a series of PRs designed to remove the need for custom model and index metaclasses. Follow-up changes should allow us to move the attribute dictionary creation and discriminator value assignment into the __init_subclass__ method.

garrettheel (approved):

PR #998, opened 2021-11-17
Remove IndexMeta and create attribute dictionary in Index.__init_subclass__

jpinner-lyft:


jpinner-lyft:
@ikonst @garrettheel This is the next in the series of PEP 487 PRs. Here we remove the index meta class entirely.

Technically this is a breaking change -- if needed I can reintroduce IndexMeta and just make it a no-op so that anyone extending it doesn't have to change their code. Thoughts?

pynamodb/indexes.py
```diff
@@ -23,29 +22,20 @@
...
     """
     Base class for secondary indexes
     """
     Meta: Any = None
 
+    @classmethod
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)  # type: ignore  # see https://github.com/python/mypy/issues/4660
+        if cls.Meta is not None:
+            cls.Meta.attributes = {}
```
garrettheel, line 35: Any specific reason to make this public?

garrettheel: missed that this is already public, nevermind

garrettheel (approved):
IMO this is fine to remove without a major version bump. We don't mention it in anything but the auto-generated code docs and any use of it would be fairly unconventional.

PR #1003, opened 2021-12-07
Remove default region

### LINKED ISSUE #211: Use region from aws configuration by default

jalaziz:
Allow the default region to be determined by botocore.

Boto supports loading configuration and credentials from various
providers with a deined order of preference (e.g. environment variables,
config files, metadata service, etc). By hardcoding the defualt region
as us-east-1, we were preventing boto from looking up the correct region
dynamically as one might expect when using the AWS CLI or SDKs.

This change ensures that region configuration is aligned with AWS SDKs
and tables are not unexpectedly created in the wrong region because the
default was not overriden.

This is a breaking change!

Fixes #211


ikonst (approved):

ikonst:
I wonder if we should bump version to 6.0a1 at this point.

jalaziz:
Fixed some failing integration tests that starting failing after I changed my local AWS credentials config.

jalaziz:
@ikonst anything I need to do to get the workflow running and this PR mergeD?

jalaziz:
@ikonst sorry, need to unblock the tests again :(. Hopefully they pass now, had to set an environment variable to set a default region since the tests are not running in an AWS environment.

jalaziz:
Thanks! Tests should be stable now.

Please let me know if there's anything else needed to get this merged 😃

ikonst:
@garrettheel I'd like to merge this. Should we consider this compatibility-breaking and bump to 6.x pre-release before or after merging this?

garrettheel:
> @garrettheel I'd like to merge this. Should we consider this compatibility-breaking and bump to 6.x pre-release before or after merging this?

Yep, this will be part of 6.x and we'll call it out as a breaking change.

Thanks @jalaziz for contributing!

PR #1009, opened 2022-01-03
drop support for python 3.6

jpinner-lyft:
- mypy >= 0.800 is required for python 3.9 and >= 0.930 is required for pypy (https://github.com/python/mypy/pull/11350)
- GenericMeta is no longer necessary on python 3.7+ (https://github.com/python/typing/issues/449)
- disabling mypy tests on pypy even though they pass because they take > 5 minutes to run


jpinner-lyft:
note: i've temporarily ignored mypy failures on the example files -- will fix those in another commit on this PR

garrettheel (approved):

PR #1021, opened 2022-02-14
Add host to settings documentation

delenamalan:


docs/settings.rst
```diff
@@ -70,6 +70,14 @@ proxy before forwarding along. Failure to strip these headers before sending to
 will result in an ``InvalidSignatureException`` due to request signing.
 
 
+host
+------
+
+Default: automatically constructed by boto.
```
ikonst, line 76:
```suggestion
Default: automatically constructed by boto to account for region
```

ikonst: (is this correct?)

delenamalan: I'm not 100% sure. Maybe best to keep it vague?

ikonst: The URLs are e.g. dynamodb.us-east-1.amazonaws.com, so I think that's the case

PR #1038, opened 2022-04-19
aws_access_key_id, aws_secret_access_key will be thread safe

ikeikeikeike:
A `connection.session` is not thread safety yet when the AWS-Credentials are set on Meta class.

Ref: https://github.com/pynamodb/PynamoDB/pull/393

### DIFF
--- a/pynamodb/connection/table.py
+++ b/pynamodb/connection/table.py
@@ -32,19 +32,28 @@ class TableConnection:
         aws_session_token: Optional[str] = None,
     ) -> None:
         self.table_name = table_name
-        self.connection = Connection(region=region,
-                                     host=host,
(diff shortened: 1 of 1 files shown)


bungoume:
@ikonst Could you please review this code?

I am using credentials set to Model Meta class due to my infrastructure configuration.
However, I am having problems with the EC2 Instance role being used unintentionally instead of credentials.

This pull request may solve that problem. I would like to use it as a patch version diff if possible.
Thank you.

ikonst:
Sorry for the delay. This was just addressed in a similar way in #1160. I'm also backporting the fix to the stable release, to be released in pynamodb 5.4.1.

PR #1050, opened 2022-05-26
Exposing control over return_values in Model.update()

chkothe:
### Problem
Currently, invoking `.update()` on a model will _always_ read back the entire document from the DB with no option to override this behavior. If one's object is very large (e.g., contains long lists), one may not wish to receive everything (or anything at all) back after performing an update (e.g., in the common use case of a REST API handler). That's a use case here at Intheon.

DynamoDB has of course a feature for that, the [ReturnValues](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#API_UpdateItem_RequestParameters) request parameter (which can take the values `NONE`, `ALL_NEW`, `ALL_OLD`, `UPDATED_NEW` and `UPDATED_OLD`), and that feature is already exposed in PynamoDB, but so far only in the low-level `TableConnection.update_item()` method. Unfortunately, falling back to that in user code would require the user to duplicate some internal PynamoDB logic and/or to invoke non-public methods of the model, so it's not really an alternative that can be recommended.

### This PR
Currently, `Model.update()` simply hard-codes this value to `ALL_NEW`. This PR proposes to make that a new parameter instead, which can be overridden as in `item.update(actions=[...], read_back=NONE)` to read nothing back from the DB. At this time, the PR is trivial and permits only either `ALL_NEW` (default, and equivalent to the current behavior) or `NONE`, which reads nothing back and consequently returns `None` and leaves the item unaltered.

I've pondered whether it should be called `return_values` instead, but that seems rather confusing to the user, who may assume that the flag affects only the return values of the `update()` method, while the local item might still be updated in-place as usual. Calling it `read_back` makes that clearer since it talks about what information comes back from the DB. That interpretation also holds up when in the future someone decides to add support for other DynamoDB modes like `ALL_OLD`, in which case we'd expect that the local copy will be populated with whatever the DB state was before the update happened (rather than that only the return value has the old data). 

While this PR is very simple, in the future someone could implement support for partial updates of the local record (via `UPDATED_NEW`), and this argument would allow the user to selectively choose one read-back mode or another, which would be an application-dependent choice (e.g., in the presence of concurrency), so there's some future-proofing in there.

A unit test that checks these semantics is provided.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -439,7 +439,7 @@ class Model(AttributeContainer, metaclass=MetaModel):
         if not isinstance(actions, list) or len(actions) == 0:
             raise TypeError("the value of `actions` is expected to be a non-empty list")
         if read_back not in (ALL_NEW, NONE):
-            raise ValueError("expected `read_back` to be `ALL_NEW` or `NONE`, but was: {}".format(return_values))
+            raise ValueError("expected `read_back` to be `ALL_NEW` or `NONE`, but was: {}".format(read_back))
 
         hk_value, rk_value = self._get_hash_range_key_serialized_values()
         version_condition = self._handle_version_attribute(actions=actions)
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -46,7 +46,7 @@ from pynamodb.constants import (
     ATTR_DEFINITIONS, ATTR_NAME, ATTR_TYPE, KEY_SCHEMA,
     KEY_TYPE, ITEM, READ_CAPACITY_UNITS, WRITE_CAPACITY_UNITS,
     RANGE_KEY, ATTRIBUTES, PUT, DELETE, RESPONSES,
-    INDEX_NAME, PROVISIONED_THROUGHPUT, PROJECTION, ALL_NEW,
+    INDEX_NAME, PROVISIONED_THROUGHPUT, PROJECTION, ALL_NEW, NONE,
     GLOBAL_SECONDARY_INDEXES, LOCAL_SECONDARY_INDEXES, KEYS,
     PROJECTION_TYPE, NON_KEY_ATTRIBUTES,
     TABLE_STATUS, ACTIVE, RETURN_VALUES, BATCH_GET_PAGE_LIMIT,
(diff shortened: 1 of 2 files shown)


pynamodb/models.py
```diff
@@ -417,31 +417,43 @@ def delete(self, condition: Optional[Condition] = None, settings: OperationSetti
...
+        :param read_back: what to read back from the DB after the update
+          ALL_NEW: read back entire object after the update (default)
+          NONE: read back nothing, local version is not updated
         :raises ModelInstance.DoesNotExist: if the object to be updated does not exist
         :raises pynamodb.exceptions.UpdateError: if the `condition` is not met
         """
         if not isinstance(actions, list) or len(actions) == 0:
             raise TypeError("the value of `actions` is expected to be a non-empty list")
+        if read_back not in (ALL_NEW, NONE):
+            raise ValueError("expected `read_back` to be `ALL_NEW` or `NONE`, but was: {}".format(read_back))
```
ikonst, line 442:
```suggestion
            raise ValueError(f"expected `read_back` to be `ALL_NEW` or `NONE`, but was: '{read_back}'")
```

tests/test_model.py
```diff
@@ -921,6 +921,28 @@ def test_update(self, mock_time):
             assert item.views is None
             self.assertEqual({'bob'}, item.custom_aliases)
 
+    def test_update_readback(self):
+        self.init_table_meta(SimpleUserModel, SIMPLE_MODEL_TABLE_DATA)
```
ikonst, line 925:
```suggestion
```
No longer needed 😅

tests/test_model.py
```diff
@@ -921,6 +921,28 @@ def test_update(self, mock_time):
...
+            params = {
+                'TableName': 'SimpleModel',
+                'Key': {'user_name': {'S': 'foo'}},
+                'ReturnValues': 'NONE',
+                'UpdateExpression': 'SET #0 = :0',
+                'ExpressionAttributeNames': {'#0': 'email'},
+                'ExpressionAttributeValues': {':0': {'S': 'changed@example.com'}},
+                'ReturnConsumedCapacity': 'TOTAL'
+            }
+            args = req.call_args[0][1]
```
ikonst, line 942:
```suggestion
            args = req.call_args[0].args
```

pynamodb/models.py
```diff
@@ -417,31 +417,43 @@ def delete(self, condition: Optional[Condition] = None, settings: OperationSetti
 
         return self._get_connection().delete_item(hk_value, range_key=rk_value, condition=condition, settings=settings)
 
-    def update(self, actions: List[Action], condition: Optional[Condition] = None, settings: OperationSettings = OperationSettings.default) -> Any:
+    def update(
+        self,
+        actions: List[Action],
+        condition: Optional[Condition] = None,
+        settings: OperationSettings = OperationSettings.default,
+        read_back: str = ALL_NEW,
```
ikonst, line 425: I wonder if we should make it an enum. This library isn't awfully modern, but maybe there's an opportunity to change that (especially as we're establishing a new API here).

ikonst:
Sorry for overlooking this PR for so long. This is a very valid addition!

pynamodb/models.py
marioortizmanero: I'd say an enum is a better choice, too. Let's start to use strong typing whenever possible.

ikonst:
I definitely should merge this before 6.x.

PR #1057, opened 2022-06-25
Fix typing on TransactGet

wasinski:
`TransactGet` can be used to fetch different Models in the same transaction, but because the `_M` was bound to the instance it was not possible to type such code correctly. I think `TransactGet` should be typed the same way as `TransactWrite`.

Additionally, I added typing to `__init__` to prevent _Call to untyped function in typed context_

### DIFF
--- a/pynamodb/transactions.py
+++ b/pynamodb/transactions.py
@@ -31,11 +31,11 @@ class Transaction:
             self._commit()
 
 
-class TransactGet(Generic[_M], Transaction):
+class TransactGet(Transaction):
 
     _results: Optional[List] = None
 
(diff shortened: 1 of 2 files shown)


garrettheel:
Looks good - would you mind adding a test to `test_mypy.py` that illustrates the bug you're fixing?

wasinski:
Sure :ok_hand:

wasinski:
@garrettheel test added. That was my first time writing a test for typing, so it might not be what you wanted, I'll happily improve it if necessary

garrettheel (approved):
@wasinski looks great, thanks very much for fixing!

PR #1079, opened 2022-10-11
Switch back to botocore _make_api_call under the hood

garrettheel:
It's not clear that the reasons for diverging from botocore's `_make_api_call` are still valid 7 years later. The most immediate negative impact of this implementation is that PynamoDB is incompatible with the [opentelemetry botocore instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore).

Our retry implementation is also lacking compared to botocore's these days, so we'll be better off by switching back.

Note that this PR _lessens_ the degree to which we're using internal private methods in botocore, but does not eliminate it entirely.

TODO:
 - [ ] Benchmark to ensure this doesn't cause a noticeable performance regression
 - [x] Deprecate `base_backoff_ms` (not valid when using native botocore retries)
 - [x] Ensure that binary attribute decoding still occurs
 - [x] Ensure that exceptions are wrapped appropriately, check back-compat (currently missing `VerboseClientError` wrapping)
 - [x] Reimplement extra_headers (through the `before-sign.` hook)

Note: I also modernized the connection/base test since it was the most affected by this change.

------------------------------

**Benchmarks**

|  | previous calls/sec | new calls/sec | delta | 
|-|-------------------|---------------|------|
| get_item | 5064 | 3718 | -27% |
| put_item | 4353 | 3267 | -25% |


pynamodb/connection/base.py
```diff
@@ -349,124 +349,7 @@ def send_pre_boto_callback(self, operation_name, req_uuid, table_name):
             log.exception("pre_boto callback threw an exception.")
 
     def _make_api_call(self, operation_name: str, operation_kwargs: Dict, settings: OperationSettings = OperationSettings.default) -> Dict:
```
matthewgrossman, line 351: are we concerned about `settings: OperationSettings` now not being used? I see it only passed into `self._create_prepared_request`

matthewgrossman: ah thats probably `Reimplement extra_headers (through the before-sign. hook)`

tomwans:
+1

pynamodb/connection/base.py
ikonst: The motivation for 'settings' was to allow per-request controls of retries, e.g. you might want to hedged-retry a read but not a write. Honestly didn't end up using this anywhere. With retries delegated to boto, this might not be needed since we'd no longer retry perm-fails indiscriminately.

garrettheel: I removed this in https://github.com/pynamodb/PynamoDB/pull/1079/commits/4f28c9f2a43ea57f45b291580a194121b9363110 since I can't find a good way to retain this without taking more deps on botocore internals, and I think it's unlikely to be used much

garrettheel:
Update on this: the performance regression was larger than I expected, so I don't feel comfortable merging this as is. While I was able to get back a chunk of that by removing additional response parsing (~5%), there is still another 20% or so regression that comes from a lot of little things and is not easily fixed (hooks, passing around context for retries, etc.).

I'll leave this open for now in case something changes later that makes this worth revisiting. At least the benchmark script is likely to come in handy in future

ilanjb:
@garrettheel is this still in the works?

garrettheel:
@ilanjb not right now, the performance regression of 25% is a bit too high. this could be revisited if botocore performance improves and lessens the delta

test.py
```diff
@@ -0,0 +1,49 @@
+import traceback
```
ikonst, line 1: test left behind

ikonst:
> removing additional response parsing

What change is that?

FWIW, I'm feeling like we should merge this for sanity/correctness, and then perhaps plead botocore to optimize? (Are there any obvious bottlenecks we can point them at?)

ikonst:
Fixed conflicts with master. Given the resourcing to maintain this project and how deeper uses of internal botocore APIs made us rush with hot fixes last year, I think the 20% slowdown is a good tradeoff.

garrettheel:
> Fixed conflicts with master. Given the resourcing to maintain this project and how deeper uses of internal botocore APIs made us rush with hot fixes last year, I think the 20% slowdown is a good tradeoff.

That's fair, I'm on board with merging if you like. We should just make sure to update the README/project motivation accordingly and be pretty upfront in the release notes

ikonst (approved):

Issue #422, opened 2017-11-30
Initial DescribeTable before GetItem


levesquejf:
While profiling my application, I realized that there is an initial DescribeTable call sent to DynamoDB before a first GetItem call. Is there a way to remove that DescribeTable call? It adds a lot of latency to the application running on AWS Lambda.

Below is a complete example that can be run locally with DynamoDB Local. The first execution will create the table and add a first item. The second execution will call GetItem four times. Before the first GetItem call, a DescribeTable is added. Package `wrapt` is required.

```bash
$ python test_describe_table.py
  0.05 ms - Calling DescribeTable
429.72 ms - Calling DescribeTable
440.19 ms - Calling CreateTable
571.44 ms - Calling DescribeTable
583.39 ms - Calling DescribeTable
600.91 ms - Calling PutItem

$ python test_describe_table.py
  0.04 ms - Calling DescribeTable
 57.30 ms - Calling GetItem
100.75 ms - Calling GetItem
114.38 ms - Calling GetItem
129.07 ms - Calling GetItem
142.16 ms - Calling PutItem
```

```python
from random import randint
from time import time
from wrapt import wrap_function_wrapper

from pynamodb.models import Model
from pynamodb.attributes import NumberAttribute, UnicodeAttribute
from pynamodb.exceptions import TableDoesNotExist


class TestTable(Model):
    class Meta:
        table_name = 'TestTable'
        host = 'http://localhost:8000'

    key = UnicodeAttribute(hash_key=True, attr_name='k')
    value = NumberAttribute(default=0, attr_name='v')


def patch_dynamodb():
    wrap_function_wrapper(
        'pynamodb.connection.base',
        'Connection.dispatch',
        xray_traced_pynamodb,
    )


def xray_traced_pynamodb(wrapped, instance, args, kwargs):
    print('{:>6.2f} ms - Calling {}'.format(1000 * (time() - start_time), args[0]))
    return wrapped(*args, **kwargs)


if __name__ == '__main__':
    start_time = time()
    patch_dynamodb()
    key = 'test-key'
    try:
        item = TestTable.get(key)
        item = TestTable.get(key)
        item = TestTable.get(key)
        item = TestTable.get(key)
    except TableDoesNotExist:
        TestTable.create_table(read_capacity_units=1, write_capacity_units=1, wait=True)
        item = TestTable(key, value=randint(0, 100))
        item.save()
```

phil-hachey:
[...]

PR #1091, opened 2022-10-27
Do not call DescribeTable for models

ikonst:
When using DynamoDB through a `Model` or `Index` (rather than `(Table)Connection` directly), we will derive the "meta-table" from the model itself rather than make an initial `DescribeTable` call. This has numerous advantages:
- Faster bootstrap (important for lambdas, as pointed out in #422)
- More consistent handling of attribute types: Before this change, if the PynamoDB model definition and the DynamoDB table definition disagreed on a key attribute's type, PynamoDB would use its own idea of the type in some code paths and the underlying type in others. Now it would consistently use its own idea of the type, allowing the erroneous model definition to be spotted sooner.
- Easier testing, since there's no longer a one-off request that only happens once and affects global state.

This approach attempts to change the library as little as possible, by synthesizing a MetaTable from the model.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -568,15 +568,15 @@ class Connection(object):
 
     def add_meta_table(self, meta_table: MetaTable) -> None:
         """
-        Sets the meta table for a given table name
+        Adds information about the table's schema.
         """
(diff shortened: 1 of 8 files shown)


garrettheel (approved):
lgtm!

PR #1096, opened 2022-11-03
Require Attribute default to be immutable or callable

ikonst:
Making it harder to accidentally mutate a default.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -53,6 +53,8 @@ _ACT = TypeVar('_ACT', bound = 'AttributeContainer')
 
 _A = TypeVar('_A', bound='Attribute')
 
+_IMMUTABLE_TYPES = (str, int, float, datetime, timedelta, bytes, bool, tuple, frozenset, type(None))
+
 
 class Attribute(Generic[_T]):
     """
@@ -72,6 +74,10 @@ class Attribute(Generic[_T]):
     ) -> None:
         if default and default_for_new:
             raise ValueError("An attribute cannot have both default and default_for_new parameters")
+        if not callable(default) and not isinstance(default, _IMMUTABLE_TYPES):
+            raise ValueError("default must be immutable type or a callable")
+        if not callable(default_for_new) and not isinstance(default_for_new, _IMMUTABLE_TYPES):
+            raise ValueError("default_for_new must be immutable type or a callable")
         self.default = default
         # This default is only set for new objects (ie: it's not set for re-saved objects)
         self.default_for_new = default_for_new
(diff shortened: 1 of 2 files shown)


tests/test_attributes.py
```diff
@@ -49,7 +49,7 @@ class CustomAttrMap(MapAttribute):
 
 
 class DefaultsMap(MapAttribute):
-    map_field = MapAttribute(default={})
+    map_field = MapAttribute(default=dict)
```
garrettheel, line 52: whoopsie

pynamodb/attributes.py
```diff
@@ -72,6 +74,10 @@ def __init__(
     ) -> None:
         if default and default_for_new:
             raise ValueError("An attribute cannot have both default and default_for_new parameters")
+        if not callable(default) and not isinstance(default, _IMMUTABLE_TYPES):
+            raise ValueError("default must be immutable type or a callable")
```
garrettheel, line 78: Thoughts on how we can help folks get to a fix quickly here? I was thinking we might want to suggest wrapping in a lambda or link to a docs page with some info

pynamodb/attributes.py
```diff
@@ -72,6 +74,10 @@ def __init__(
     ) -> None:
```
garrettheel, line 74: Do you think we could type `default`/`default_for_new` with the immutable types so that anyone using mypy gets earlier feedback?

garrettheel (commented):
Overall looks good to me, couple questions

pynamodb/attributes.py
ikonst: I tried but it gets messy.

At the least you'd need two overloads:
```python
ImmutableT = TypeVar('ImmutableT', str, int, ...)

class Attribute(Generic[T]):
  @overload
  def __init__(self: Attribute[ImmutableT], ..., default: ImmutableT | Callable[..., ImmutableT] | None, ...) -> None: ...
  
  @overload
  def __init__(self, ..., default: Callable[..., T] | None, ...) -> None: ...
```

But that [doesn't seem to work](https://github.com/python/mypy/issues/13993). In fact, even a simpler one where I explicitly single out `str` doesn't seem to work:
```python
class Attribute(Generic[T]):
  @overload
  def __init__(self: Attribute[str], ..., default: str | Callable[..., str] | None, ...) -> None: ...
  
  @overload
  def __init__(self, ..., default: Callable[..., T] | None, ...) -> None: ...
```

See [playground](https://mypy-play.net/?mypy=latest&python=3.10&gist=30f7918d00fbe9c4a86bfacea2d20da3): note how this line passes:
```python
list_attr3 = ListAttribute(default='foobar')
```

pynamodb/attributes.py
ikonst: Good point. Took the time to document `Attribute`'s initializer: 674f91e383ccd1e1a03dde31e96fd52fc31611f7

garrettheel: nice, they look great

pynamodb/attributes.py
garrettheel: gotcha - it's a shame but I suppose this is fine for now and we can hope for type improvements later

garrettheel (approved):

PR #1101, opened 2022-11-05
Improve first iteration logic

ikonst:
Improving the first iteration logic in `PageIterator` and `ResultIterator`.

Both iterators had a `_first_iteration` state attribute that could be replaced with better logic:
- In `PageIterator`, it can be replaced with `_is_last_page` which
  - is exception-safe (only mutate after we do I/O), and
  - (IMO) more directly relates to the state we're handling
- In `ResultsIterator`, we can do without `_first_iteration` (but we should initialize `_index` and `_count`)

Inspired by #1059.

### DIFF
--- a/pynamodb/pagination.py
+++ b/pynamodb/pagination.py
@@ -1,5 +1,5 @@
 import time
-from typing import Any, Callable, Dict, Iterable, Iterator, TypeVar, Optional
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, TypeVar
 
 from pynamodb.constants import (CAMEL_COUNT, ITEMS, LAST_EVALUATED_KEY, SCANNED_COUNT,
                                 CONSUMED_CAPACITY, TOTAL, CAPACITY_UNITS)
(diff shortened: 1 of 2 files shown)


pynamodb/pagination.py
```diff
@@ -170,15 +169,17 @@ def __init__(
         settings: OperationSettings = OperationSettings.default,
     ) -> None:
         self.page_iter: PageIterator = PageIterator(operation, args, kwargs, rate_limit, settings)
-        self._first_iteration = True
         self._map_fn = map_fn
         self._limit = limit
         self._total_count = 0
+        self._items: List = []
```
anandswaminathan, line 175: @ikonst Why is this required. I don't see any calls to `self._items` ahead of https://github.com/pynamodb/PynamoDB/pull/1101/files#diff-05a16c49172d4c74ad60089cd28c1ba68f5da9699bbb94ce1ccee0f9e60d9dfdR182

ikonst: You are right. And for all practical reasons, `self._items[X]` will raise just the same (for an empty list).

pynamodb/pagination.py
```diff
@@ -1,5 +1,5 @@
 import time
-from typing import Any, Callable, Dict, Iterable, Iterator, TypeVar, Optional
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, TypeVar
```
anandswaminathan, line 2: Remove "List"

anandswaminathan (approved):
One small comment

pynamodb/pagination.py
ikonst: Thanks. We really ought to upgrade to repo to common tooling (black, reorder-python-imports, pyupgrade) but I'm dreading the diff size 😅

PR #1102, opened 2022-11-05
Make Conditions and Actions equatable

ikonst:


pynamodb/expressions/condition.py
```diff
@@ -14,6 +14,19 @@ def __init__(self, operator: str, *values) -> None:
...
         self.values = values
 
+    def __eq__(self, other: Any) -> bool:
+        return (
+            type(self) is type(other)
+            and self.operator == other.operator
+            and len(self.values) == len(other.values)
+            and all(
+                (
+                    type(v1) is type(v2)
```
garrettheel, line 24: Is this redundant given that `_equals_to` also checks?

pynamodb/expressions/operand.py
```diff
@@ -248,6 +260,13 @@ def __init__(self, attribute_or_path: Union['Attribute', str, List[str]]) -> Non
             raise ValueError("path cannot be empty")
         super(Path, self).__init__(get_path_segments(path))
 
+    def _equals_to(self, other: Any) -> bool:
+        return (
+            super()._equals_to(other)
+            and self.attr_type == other.attr_type
+            and self.attribute is other.attribute
```
garrettheel, line 267: `is` might be worthy of a comment for the `Attribute` case here

garrettheel (approved):

pynamodb/expressions/operand.py
ikonst: to clarify that I'm doing `is` because it's a type?

ikonst: ah right, right

PR #1104, opened 2022-11-08
Use type var _T on Attribute.set expression builder

conjmurph:


examples/model.py
```diff
@@ -190,9 +190,10 @@ class Meta:
 ]))
 
 # Update list attribute
+thread_item.notes.append(["new note"])
 print(thread_item.update(actions=[
     Thread.notes.set(
-        Thread.notes.append(["new note"])
+        thread_item.notes
     )
```
ikonst, line 197: It's confusing, but the code you've changed is how you add an item to a list atomically in DynamoDB.

In "raw" DynamoDB, that'd be
```
SET Thread.notes = list_append(Thread.notes, "new note")
```
See [docs](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#DDB-UpdateItem-request-UpdateExpression)

ikonst (changes requested):

pynamodb/attributes.py
```diff
@@ -253,7 +253,10 @@ def append(self, other: Iterable) -> '_ListAppend':
     def prepend(self, other: Iterable) -> '_ListAppend':
         return Path(self).prepend(other)
 
-    def set(self, value: Any) -> 'SetAction':
+    def set(
```
ikonst, line 256: Let's add some negative tests to typing_tests, i.e. where you pass stuff that shouldn't work, and have to `# type: ignore[some-code]` it.

We're using the [warn_unused_ignores technique](https://typing.readthedocs.io/en/latest/source/quality.html#testing-using-assert-type-and-warn-unused-ignores) here: adding a `# type: ignore` will only pass mypy check if it's actually suppressing what otherwise would be an error.

ikonst (approved):

PR #1107, opened 2022-11-10
create_table should include polymorphic indexes

ikonst:
With polymorphism, an index's key attribute might be defined only in a derived model (e.g. if it pertains only to a certain kind of model — for other models, it'd be a sparse index), and so it makes sense to add that index only to the derived model, e.g.
```python
class BaseModel(Model):
    ...
    cls = DescriminatorAttribute()

class DerivedIndex(GlobalSecondaryIndex):
    class Meta:
        index_name = 'derived'
        ...
    ham = UnicodeAttribute(hash_key=True)

class DerivedModel(BaseModel, discriminator='spam'):
    ham  = UnicodeAttribute()
    index = DerivedIndex()
```

A common pattern in a development environment is to use `Model.create_table`. For a polymorphic model, given you have to pick _one_ model to invoke `create_table` on, it only serves right that you invoke it on the base model.

Before this change, the base model's "schema"  didn't include all the derived models' indexes. With this change, it will.

### DIFF
--- a/pynamodb/indexes.py
+++ b/pynamodb/indexes.py
@@ -5,16 +5,17 @@ from inspect import getmembers
 from typing import Any, Dict, Generic, List, Optional, Type, TypeVar
 from typing import TYPE_CHECKING
 
+from pynamodb._schema import IndexSchema, GlobalSecondaryIndexSchema
+from pynamodb._schema import ModelSchema
 from pynamodb.constants import (
-    INCLUDE, ALL, KEYS_ONLY, ATTR_NAME, ATTR_TYPE, KEY_TYPE, KEY_SCHEMA,
-    ATTR_DEFINITIONS, PROJECTION_TYPE, NON_KEY_ATTRIBUTES,
+    INCLUDE, ALL, KEYS_ONLY, ATTR_NAME, ATTR_TYPE, KEY_TYPE,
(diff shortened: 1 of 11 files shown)


pynamodb/indexes.py
```diff
@@ -147,17 +150,18 @@ def _get_schema(cls) -> Dict:
...
                 PROJECTION_TYPE: cls.Meta.projection.projection_type,
             },
+            'attribute_definitions': {},
         }
+
         for attr_cls in cls.Meta.attributes.values():
-            if attr_cls.is_hash_key:
-                schema['key_schema'].append({
+            if attr_cls.is_hash_key or attr_cls.is_range_key:
+                schema['attribute_definitions'][attr_cls.attr_name, attr_cls.attr_type] = {
```
ikonst, line 158: Now indexes will return "attribute definitions" even though they only roll up into the table's.

tests/test_model.py
```diff
@@ -481,6 +481,7 @@ def assert_dict_lists_equal(self, list1, list2):
                 if d2_item == d1_item:
                     found = True
             if not found:
+                assert list1 == list2
```
garrettheel, line 484: Can we just drop this function entirely with pytest? Probably worth doing a sweep as a separate PR...

garrettheel (approved):
Glad to see we've squashed the last subtle discriminator bug ;)

PR #1110, opened 2022-11-12
Split BinaryAttribute into BinaryDataAttribute and LegacyBinaryAttribute

ikonst:
For many versions we had a known issue where Binary(Set)Attribute were double-base64-encoded in the underlying DynamoDB table item. Clearly this is inefficient and we wanted PynamoDB to be efficient by default, but similarly didn't want to risk breaking applications during an upgrade.

With that in mind, we decided to split `BinaryAttribute` into `LegacyBinary(Set)Attribute` (the old behavior, for compatibility with existing data) and `BinaryData(Set)Attribute` (new correct behavior), the rationale being that a symbol rename will cause upgraded services to fail to load (or fail tests) and thus prompt an informed decision. The alternative was to keep `BinaryAttribute` with the broken behavior (and introduce a "V2" class for new behavior), which would make PynamoDB less elegant for green-field uses.

P.S. the word "data" after "binary" doesn't really serve a purpose, but simply an attempt to pick a name that's different yet ordinary

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -1,6 +1,7 @@
 """
 PynamoDB attributes
 """
+import base64
 import calendar
 import collections.abc
 import json
@@ -443,8 +444,10 @@ class AttributeContainer(metaclass=AttributeContainerMeta):
     def _coerce_attribute_type(attr_type: str, attribute_value: Dict[str, Any]):
         # coerce attribute types to disambiguate json string and array types
         if attr_type == BINARY and STRING in attribute_value:
-            attribute_value[BINARY] = attribute_value.pop(STRING)
-        if attr_type in {BINARY_SET, NUMBER_SET, STRING_SET} and LIST in attribute_value:
+            attribute_value[BINARY] = base64.b64decode(attribute_value.pop(STRING))
+        elif attr_type == BINARY_SET and LIST in attribute_value:
+            attribute_value[BINARY_SET] = [base64.b64decode(v[STRING]) for v in attribute_value.pop(LIST)]
+        elif attr_type in {NUMBER_SET, STRING_SET} and LIST in attribute_value:
             json_type = NUMBER if attr_type == NUMBER_SET else STRING
             if all(next(iter(v)) == json_type for v in attribute_value[LIST]):
                 attribute_value[attr_type] = [value[json_type] for value in attribute_value.pop(LIST)]
@@ -516,9 +519,13 @@ class DiscriminatorAttribute(Attribute[type]):
         return self._discriminator_map[value]
 
 
-class BinaryAttribute(Attribute[bytes]):
+class LegacyBinaryAttribute(Attribute[bytes]):
     """
-    A binary attribute
+    A legacy binary attribute whose underlying data is double base64-encoded.
+    Use for migrating :code:`BinaryAttribute` from PynamoDB 5 and lower.
+
+    .. warning::
+        Do not use for new models.
     """
     attr_type = BINARY
 
@@ -535,9 +542,52 @@ class BinaryAttribute(Attribute[bytes]):
         return b64decode(value)
 
 
-class BinarySetAttribute(Attribute[Set[bytes]]):
+class BinaryDataAttribute(Attribute[bytes]):
+    """
+    A binary attribute
+    """
+    attr_type = BINARY
+
+    def serialize(self, value):
+        """
+        Returns a binary string
+        """
+        return value
+
+    def deserialize(self, value):
+        """
+        Returns a decoded byte string from a binary value
+        """
+        return value
+
+
+class BinaryDataSetAttribute(Attribute[Set[bytes]]):
     """
     A binary set
+    """
+    attr_type = BINARY_SET
+    null = True
+
+    def serialize(self, value):
+        """
+        Returns a list of strings. Encodes empty sets as "None".
+        """
+        return list(value) or None
+
+    def deserialize(self, value):
+        """
+        Returns a set of decoded byte strings from base64 encoded values.
+        """
+        return set(value)
+
+
+class LegacyBinarySetAttribute(Attribute[Set[bytes]]):
+    """
+    A legacy binary set attribute whose underlying data is double base64-encoded.
+    Use for migrating :code:`BinarySetAttribute` from PynamoDB 5 and lower.
+
+    .. warning::
+        Do not use for new models.
     """
     attr_type = BINARY_SET
     null = True
@@ -1258,8 +1308,8 @@ class ListAttribute(Generic[_T], Attribute[List[_T]]):
 
 
 DESERIALIZE_CLASS_MAP: Dict[str, Attribute] = {
-    BINARY: BinaryAttribute(),
-    BINARY_SET: BinarySetAttribute(),
+    BINARY: BinaryDataAttribute(),
+    BINARY_SET: BinaryDataSetAttribute(),
     BOOLEAN: BooleanAttribute(),
     LIST: ListAttribute(),
     MAP: MapAttribute(),
@@ -1278,5 +1328,5 @@ SERIALIZE_CLASS_MAP = {
     float: NumberAttribute(),
     int: NumberAttribute(),
     str: UnicodeAttribute(),
-    bytes: BinaryAttribute(),
+    bytes: BinaryDataAttribute(),
 }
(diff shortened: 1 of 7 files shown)


pynamodb/attributes.py
```diff
@@ -535,13 +542,56 @@ def deserialize(self, value):
...
+        return set(value)
+
+
+class LegacyBinarySetAttribute(Attribute[Set[bytes]]):
+    """
+    A legacy binary set attribute whose underlying data is double base64-encoded.
+    Use for migrating :code:`BinarySetAttribute` from PynamoDB 5 and lower.
+
+    .. warning::
+        Do not use for new models.
```
garrettheel, line 590:
```suggestion
        Do not use for new models. Use :code:`BinaryDataSetAttribute` instead.
```

pynamodb/attributes.py
```diff
@@ -1258,8 +1308,8 @@ def _get_serialize_class(self, value):
 
 
 DESERIALIZE_CLASS_MAP: Dict[str, Attribute] = {
-    BINARY: BinaryAttribute(),
-    BINARY_SET: BinarySetAttribute(),
+    BINARY: BinaryDataAttribute(),
```
garrettheel, line 1311: This will be a problem if anyone had a `BinaryAttribute` stored in a `DynamicMapAttribute` right?

ikonst: Ah, right, I meant to go back to that, since that change looked iffy. It also depends on whether `_handle_binary_attributes` handles anything recursively and looking at it, I don't think it does... 🙀

pynamodb/attributes.py
ikonst: Except I'd use :class:...

ikonst:
Looking at `_handle_binary_attributes`, it doesn't seem like it recurses over lists and maps at all, does it? I wonder if our "emulating boto" is faithful to the original in this case, but regardless, if that's the case, then double base64 was only ever a problem for top-level binary attributes, which is the better scenario.

ikonst:
@garrettheel Actually, it's worse 🙀 
```python
from pynamodb.attributes import BinaryAttribute
from pynamodb.attributes import MapAttribute
from pynamodb.attributes import UnicodeAttribute


class MyMap(MapAttribute):
    nested_data = BinaryAttribute()


class MyModel(Model):
    class Meta:
        table_name = 'double_base64'
        region = 'us-east-1'
        aws_access_key_id = 'dummy'
        aws_secret_access_key = 'dummy'

    hk = UnicodeAttribute(hash_key=True)
    data = BinaryAttribute()
    map = MyMap()


my_model = MyModel('123')
my_model.data = b'test'
my_model.map = MyMap(nested_data=b'test')

# 10 serialization round-trips
for _ in range(10):
    my_model.save()
    my_model = MyModel.get('123')
```
then
```
>>> my_model.data
b'test'
>>> my_model.map.nested_data
my_model.map.nested_data=b'Vm0wd2VHUXhUWGROVldSWVYwZDRWRll3Wkc5WFZsbDNXa1JTV0ZKdGVEQmFWVll3VmpBeFdHVkdXbFpXZWxaeVZtMHhTMUl5VGtsaVJtUlhUVEZLVFZac1ZtRldNVnBXVFZWV2FHVnFRVGs9'
```

So
- top-level BinaryAttributes are always double-encoded top-level-only (compensates for `_handle_binary_attributes`)
- nested BinaryAttributes are repeatedly double-encoded(!) to the point of being useless

How about:
- keep `BinaryAttribute` named as is
- add a `BinaryAttribute(..., legacy_encoding: Optional[bool] = None)` arg that
  - for top-level attributes, **must be True or False**
  - for nested attributes, **must be None or False**

During the transitional period (likely at least for the duration of PynamoDB 6.0), these rules will apply. Later we might remove this parameter, and default to the non-legacy behavior.

garrettheel:
> @garrettheel Actually, it's worse 🙀
> 
> ```python
> from pynamodb.attributes import BinaryAttribute
> from pynamodb.attributes import MapAttribute
> from pynamodb.attributes import UnicodeAttribute
> from pynamodb.models import Model
> 
> 
> class MyMap(MapAttribute):
>     nested_data = BinaryAttribute()
> 
> 
> class MyModel(Model):
>     class Meta:
>         table_name = 'double_base64'
>         region = 'us-east-1'
>         host = 'http://localhost:8000'
>         aws_access_key_id = 'dummy'
>         aws_secret_access_key = 'dummy'
> 
>     hk = UnicodeAttribute(hash_key=True)
>     data = BinaryAttribute()
>     list = ListAttribute(of=BinaryAttribute)
>     map = MyMap()
>     raw_map = MapAttribute()
> 
> 
> my_model = MyModel('123')
> my_model.data = b'test'
> my_model.list = [b'test']
> my_model.map = MyMap(nested_data=b'test')
> my_model.raw_map = MapAttribute(nested_data=b'test')
> 
> # 10 serialization round-trips
> for _ in range(10):
>     my_model.save()
>     my_model = MyModel.get('123')
> ```
> 
> then
> 
> ```
> >>> my_model.data
> b'test'
> >>> my_model.list
> [b'Vm0wd2VHUXhUWGROVldSWVYwZDRWRll3Wkc5WFZsbDNXa1JTV0ZKdGVEQmFWVll3VmpBeFdHVkdXbFpXZWxaeVZtMHhTMUl5VGtsaVJtUlhUVEZLVFZac1ZtRldNVnBXVFZWV2FHVnFRVGs9']
> >>> my_model.map.nested_data
> b'Vm0wd2VHUXhUWGROVldSWVYwZDRWRll3Wkc5WFZsbDNXa1JTV0ZKdGVEQmFWVll3VmpBeFdHVkdXbFpXZWxaeVZtMHhTMUl5VGtsaVJtUlhUVEZLVFZac1ZtRldNVnBXVFZWV2FHVnFRVGs9'
> >>> my_model.raw_map.nested_data
> b'Vm0wd2VHUXhUWGROVldSWVYwZDRWRll3Wkc5WFZsbDNXa1JTV0ZKdGVEQmFWVll3VmpBeFdHVkdXbFpXZWxaeVZtMHhTMUl5VGtsaVJtUlhUVEZLVFZac1ZtRldNVnBXVFZWV2FHVnFRVGs9'
> ```
> 
> So
> 
> * top-level BinaryAttributes are double encoded (compensates for `_handle_binary_attributes`) but round-trip correctly
> * nested BinaryAttributes are additionally encoded on every round-trip(!), to the point of being useless
> 
> **Proposal:**
> 
> * keep `BinaryAttribute` named as is
> * add a `BinaryAttribute(..., legacy_encoding: Optional[bool] = None)` arg that
>   
>   * for top-level attributes, **must be True or False**
>   * for nested attributes, **must be None or False**
> 
> During the transitional period (likely at least for the duration of PynamoDB 6), these rules will apply. Later we might remove this arg, and default to the non-legacy behavior.

Yikes. I'm inclined to think that there are very limited (if any) people using nested binary attributes since we haven't seen reports of this on the issue tracker.

The required `legacy_encoding` attribute seems a reasonable solution to me, but shall we just keep the `True,False` options for nested attributes too? We can call out in the docs that it's very unlikely you'd ever want to set `legacy_encoding=True` on a nested attribute. I suggest this just to keep things simpler by having semantics remain the same for all usage of the attribute (and then we also don't have to use a sentinel to distinguish between unset and None).

ikonst:
> don't have to use a sentinel to distinguish between unset and None

I don't see why we'd need a sentinel attribute. Let's review again:
- Typing-wise, it's `None | False | True`.
- Logic-wise:
  - `None` is illegal for top-level, since we need you to be explicit.
  - `True` is illegal for nested, since it suggests you misunderstood. (IIUC, it's not _very unlikely_, it's zero.)

garrettheel:
> > don't have to use a sentinel to distinguish between unset and None
> 
> I don't see why we'd need a sentinel attribute. Let's review again:
> 
> * Typing-wise, it's `None | False | True`.
> * Logic-wise:
>   
>   * `None` is illegal for top-level, since we need you to be explicit.
>   * `True` is illegal for nested, since it suggests you misunderstood. (IIUC, it's not _very unlikely_, it's zero.)

I'm suggesting that it's possible someone could have been applying their own decoding to work-around this bug and that the automatic change may break them (still very unlikely, but at least theoretically possible). Conceptually I think it's easier to think about this parameter working the same way for all `BinaryAttribute`s

ikonst:
BTW, I've spot-checked a few versions and the problem reproduces on 3.x and 4.x as well.

ikonst:
Closed in favor of #1112.

PR #1112, opened 2022-11-16
Fix double base64 of binary attributes

ikonst:
For many versions (verified in versions 3-5) we had a known issue where
1. top-level Binary(Set)Attributes were double-base64-encoded in the underlying DynamoDB table item
2. nested Binary(Set)Attributes were unusable: every serialization would add another round of Base64 encoding(!), so nested binary attributes could not have been used in practice

Clearly (1) is inefficient and (2) is unusable, and we wanted PynamoDB to be efficient (and not broken) by default, but similarly didn't want to risk breaking systems during an upgrade to PynamoDB 6. With that in mind, we're introducing a `legacy_encoding` required parameter to `BinaryAttribute` and `BinarySetAttribute`, the rationale being a new required parameter will prompt an informed decision. The parameter will exist throughout the lifetime of PynamoDB 6.x and perhaps removed in a future version (in favor of `legacy_encoding=False` becoming the default behavior).

Since nested binary attributes were previously unusable, we're safely defaulting to `legacy_encoding=False` for nested attributes (and forcing it to be `False` when binary attributes are declared in non-raw maps).

Along with this, a slight breaking change in introduced: A **set** in a "raw" list or map have previously serialized as a list (**L**) but will now result in a string/number/binary set:
- for all strings, string set (**SS**)
- for all numbers, number set (**NS**)
- for all bytes, binary set (**BS**)
- else, raises an error on serialization

Additions to docs:
- [Upgrading Binary(Set)Attribute](https://pynamodb--1112.org.readthedocs.build/en/1112/upgrading_binary.html)
- [BinaryAttribute (initializer)](https://pynamodb--1112.org.readthedocs.build/en/1112/api.html#pynamodb.attributes.BinaryAttribute)
- [BinarySetAttribute (initializer)](https://pynamodb--1112.org.readthedocs.build/en/1112/api.html#pynamodb.attributes.BinarySetAttribute)

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -18,7 +18,6 @@ This is a major release and contains breaking changes. Please read the notes bel
 * Index count, query, and scan methods are now instance methods.
 
 
-
 v5.3.0
 ----------
 * No longer call ``DescribeTable`` API before first operation
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -6,8 +6,17 @@ Unreleased
 
 This is a major release and contains breaking changes. Please read the notes below carefully.
 
+* :py:class:`~pynamodb.attributes.BinaryAttribute` and :py:class:`~pynamodb.attributes.BinarySetAttribute` have undergone breaking changes:
+
+  * The attributes' internal encoding has changed. To avoid this change going unnoticed, :code:`legacy_encoding` have been made required: see :doc:`upgrading_binary` for details.
+    If your codebase uses :py:class:`~pynamodb.attributes.BinaryAttribute` or :py:class:`~pynamodb.attributes.BinarySetAttribute`,
+    go over the attribute declarations and mark them accordingly.
+  * When using binary attributes, the return value of :py:func:`~pynamodb.model.Model.serialize` will no longer be JSON-serializable
+    since it will contain :code:`bytes` objects.
+
 * Python 3.6 is no longer supported.
 * Index count, query, and scan methods are now instance methods.
+
 
 
 v5.3.0
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -11,8 +11,8 @@ This is a major release and contains breaking changes. Please read the notes bel
   * The attributes' internal encoding has changed. To prevent this change going unnoticed, :code:`legacy_encoding` have been made required: see :doc:`upgrading_binary` for details.
     If your codebase uses :py:class:`~pynamodb.attributes.BinaryAttribute` or :py:class:`~pynamodb.attributes.BinarySetAttribute`,
     go over the attribute declarations and mark them accordingly.
-  * When using binary attributes, the return value of :py:func:`~pynamodb.model.Model.serialize` will no longer be JSON-serializable
-    since it will contain :code:`bytes` objects.
+  * When using binary attributes, the return value of :py:func:`~pynamodb.models.Model.serialize` will no longer be JSON-serializable
+    since it will contain :code:`bytes` objects. Note that both `:py:func:`~pynamodb.models.Model.to_dict` and `:py:func:`~pynamodb.model.Model.to_json` are also affected.
 
 * Python 3.6 is no longer supported.
 * Index count, query, and scan methods are now instance methods.
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -8,7 +8,7 @@ This is a major release and contains breaking changes. Please read the notes bel
 
 * :py:class:`~pynamodb.attributes.BinaryAttribute` and :py:class:`~pynamodb.attributes.BinarySetAttribute` have undergone breaking changes:
 
-  * The attributes' internal encoding has changed. To avoid this change going unnoticed, :code:`legacy_encoding` have been made required: see :doc:`upgrading_binary` for details.
+  * The attributes' internal encoding has changed. To prevent this change going unnoticed, :code:`legacy_encoding` have been made required: see :doc:`upgrading_binary` for details.
     If your codebase uses :py:class:`~pynamodb.attributes.BinaryAttribute` or :py:class:`~pynamodb.attributes.BinarySetAttribute`,
     go over the attribute declarations and mark them accordingly.
   * When using binary attributes, the return value of :py:func:`~pynamodb.model.Model.serialize` will no longer be JSON-serializable
--- a/docs/upgrading_binary.rst
+++ b/docs/upgrading_binary.rst
@@ -18,14 +18,14 @@ which were addressed in PynamoDB 6:
 - Top-level binary attributes (i.e. within a :py:class:`~pynamodb.models.Model`) were being Base64-encoded
   twice. For elements in BinarySetAttribute, each element was being encoded twice.
 
-  This behavior was an oversight that carried no particular advantage, and the disadvantage of larger item sizes.
+  This behavior was an oversight and resulted in larger item sizes and non-standard semantics.
 
   .. admonition:: For example...
 
-     The 4 bytes :code:`b'\\xCA\\xFE\\xF0\\x0D'` should've been sent over the wire as :code:`b'yv7wDQ=='` (single round
+     The 4 bytes :code:`CA FE F0 0D` should've been sent over the wire as :code:`yv7wDQ==` (single round
      of Base64 encoding). Server-side, they would've been decoded and stored as 4 bytes. Instead, they were put through an extra
-     round of Base64 encoding, thus sending :code:`b'eXY3d0RRPT0='` over the wire. Server-side, they decoded into :code:`b'yv7wDQ=='`
-     and stored as 8 bytes.
+     round of Base64 encoding, thus sending :code:`eXY3d0RRPT0=` over the wire. Server-side, they decoded into :code:`yv7wDQ==`
+     (in bytes, :code:`79 76 37 77 44 51 3D 3D`) and stored as 8 bytes.
 
 - Nested binary attributes (i.e. within a :py:class:`~pynamodb.attributes.MapAttribute` and :py:class:`~pynamodb.attributes.ListAttribute`)
   were being wrapped in an additional layer of Base64 encoding on every serialization roundtrip.
--- a/docs/upgrading_binary.rst
+++ b/docs/upgrading_binary.rst
@@ -1,0 +1,191 @@
+Upgrading Binary(Set)Attribute
+==============================
+
+.. warning::
+
+    The behavior of :py:class:`~pynamodb.attributes.BinaryAttribute` and
+    :py:class:`~pynamodb.attributes.BinarySetAttribute` has changed in backwards-incompatible ways
+    as of the 6.0 release of PynamoDB.
+
+    To prevent data corruption, use :code:`legacy_encoding=True` for **existing** binary attributes.
+
+Context
+#######
+
+PynamoDB version 5 (and lower) had two bugs in the way they handled binary attributes,
+which were addressed in PynamoDB 6:
+
+- Top-level binary attributes (i.e. within a :py:class:`~pynamodb.models.Model`) were being Base64-encoded
+  twice. For elements in BinarySetAttribute, each element was being encoded twice.
+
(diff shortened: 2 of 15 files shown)


pynamodb/attributes.py
```diff
@@ -443,8 +442,10 @@ def _update_attribute_types(cls, attribute_values: Dict[str, Dict[str, Any]]):
     def _coerce_attribute_type(attr_type: str, attribute_value: Dict[str, Any]):
         # coerce attribute types to disambiguate json string and array types
         if attr_type == BINARY and STRING in attribute_value:
-            attribute_value[BINARY] = attribute_value.pop(STRING)
-        if attr_type in {BINARY_SET, NUMBER_SET, STRING_SET} and LIST in attribute_value:
+            attribute_value[BINARY] = base64.b64decode(attribute_value.pop(STRING))
+        elif attr_type == BINARY_SET and LIST in attribute_value:
+            attribute_value[BINARY_SET] = [base64.b64decode(v[STRING]) for v in attribute_value.pop(LIST)]
```
ikonst, line 447: **Why do we base64-decode here when we didn't use to?** Since `BinaryAttribute(legacy_encoding=False).deserialize` does not base64-decode (assuming `_convert_binary` in connection.py already did the job), we have to do it here for the sake of `Model.from_dict`.

pynamodb/connection/base.py
```diff
@@ -1384,10 +1386,14 @@ def _reverse_dict(d):
...
-        value = attr[BINARY_SET]
-        if value and len(value):
-            attr[BINARY_SET] = {b64decode(v.encode(DEFAULT_ENCODING)) for v in value}
+        attr[BINARY_SET] = [b64decode(v.encode()) for v in attr[BINARY_SET]]
+    elif MAP in attr:
+        for sub_attr in attr[MAP].values():
+            _convert_binary(sub_attr)
+    elif LIST in attr:
+        for sub_attr in attr[LIST]:
+            _convert_binary(sub_attr)
```
ikonst, line 1399: Maps and lists are now being recursed into. This matches botocore behavior.

ikonst:
BTW, now that `Attribute.(de)serialize` and subsequently `Model.(de)serialize` are not (de)serializing a JSON-serializable dict (for DynamoDB API) but a dict for botocore (with non-JSON-serializable bytes), there’s an unfortunate gap forming:
- `Model.serialize` ’s output cannot be JSON-serialized (w/o custom encoder)
- `Model.to_dict` ’s output does not always roundtrip — `bytes` in raw maps will be encoded as base64 strings, but then `Model.from_dict` will decode them as  `str` containing base64

Not sure what to do about that.

garrettheel:
> Along with this, a slight breaking change in introduced: A set in a "raw" list or map have previously serialized as a list (L) but will now result in a string/number/binary set

Is this breaking in the sense that we fail to deserialise these, or just that they're serialised differently when read back and written? Maybe we can add a test for this as well (apologies if one was already there and I didn't notice)

pynamodb/connection/base.py
garrettheel: Worth noting that this is going to have some performance impact (I actually did notice this as part of the delta in https://github.com/pynamodb/PynamoDB/pull/1079). We might want to run benchmarks to get a sense for it and include it in the release notes.

If the difference is meaningful then we may also want to add an optimisation for short-circuiting this when binary attributes aren't in use

garrettheel:
> BTW, now that `Attribute.(de)serialize` and subsequently `Model.(de)serialize` are not (de)serializing a JSON-serializable dict (for DynamoDB API) but a dict for botocore (with non-JSON-serializable bytes), there’s an unfortunate gap forming:
> 
> * `Model.serialize` ’s output cannot be JSON-serialized (w/o custom encoder)
> * `Model.to_dict` ’s output does not always roundtrip — `bytes` in raw maps will be encoded as base64 strings, but then `Model.from_dict` will decode them as  `str` containing base64
> 
> As a result, we don't have a JSON-serializable representation that roundtrips. Not sure what to do about that.

I wonder if we should talk more about the options here. Seems like we're drifting towards binary attributes becoming a second-class citizen that contains so many footguns that we might as well not support them natively (which also feels wrong, since they are part of the DynamoDB API).

I'd be curious to see how much worse each of these points are becoming as of this PR and what it would take to ensure that `from_dict`, `to_dict`, `serialize`, etc. work better with binary. We can maybe do that elsewhere to avoid cluttering this PR too much

ikonst:
>> Along with this, a slight breaking change in introduced: A set in a "raw" list or map have previously serialized as a list (L) but will now result in a string/number/binary set

> Is this breaking in the sense that we fail to deserialise these, or just that they're serialised differently when read back and written? Maybe we can add a test for this as well (apologies if one was already there and I didn't notice)

- `TestMapAttribute.test_attribute_children`: raw maps now round-trip sets (e.g. BS)
- `TestMapAttribute.test_serialize_invalid_set`: when assigning a set to a map, they must be heterogenous, and of one of the supported types (e.g no more `{42, "ham"}`)

The previous behavior is low on my priority to preserve, since
1. it doesn't roundtrip
    ```python
    m.raw_map.my_stuff = {42, "ham"}
    m.save()
    m.refresh()
    assert type(m.raw_map.my_stuff) is list
    ```
2. now we fail loudly so change is apparent
3. old behavior doesn't offer advantages
4. old behavior doesn't give a way to use sets in raw maps

ikonst:
> which also feels wrong, since they are part of the DynamoDB API

Agree. There are many ORMs out there, a DynamoDB specific better cover what DynamoDB supports.

On the topic of JSON-izing, here's how Amazon goes about it. In the AWS console, if you add a binary attribute and then switch from "Form" to "JSON view", there's a nested option to toggle between "JSON" and "DynamoDB JSON" and that option is disabled once there are binary attributes or sets:
image

Compatibility aside, I'd want:
- one method which returns a "JSON" and raises if sets or binaries are present
- another method which returns a "DynamoDB JSON" (with binaries as base64 `str`s, such that `json` can serialize them)

docs/upgrading_binary.rst
```diff
@@ -0,0 +1,190 @@
...
+     round of Base64 encoding, thus sending :code:`eXY3d0RRPT0=` over the wire. Server-side, they decoded into :code:`yv7wDQ==`
+     (in bytes, :code:`79 76 37 77 44 51 3D 3D`) and stored as 8 bytes.
+
+- Nested binary attributes (i.e. within a :py:class:`~pynamodb.attributes.MapAttribute` and :py:class:`~pynamodb.attributes.ListAttribute`)
+  were being wrapped in an additional layer of Base64 encoding on every serialization roundtrip.
+
+  Not only it prevented them from being deserialized correctly, but also the model would also grow
+  in size exponentially until it hit the DynamoDB item limit of 400KB. For this reason we conclude
+  that :code:`BinaryAttribute` and :code:`BinarySetAttribute` were not used in practice within maps and lists
+  before PynamoDB 6.0 and thus will not require any legacy encoding.
```
garrettheel, line 36: > not used in practice within maps and lists... and thus will not require any legacy encoding

Do we need to differentiate between raw and typed maps here? Given that the example below does mention specifying `legacy_encoding` on a map attr.

docs/upgrading_binary.rst
```diff
@@ -0,0 +1,190 @@
...
+        {b'\xF0', b'\x0D'},
+     ]
+
+
+.. _migrating:
+
+Migrating
+#########
+
+Migrating existing systems off the legacy encoding is not necessary at this time. For large tables,
```
garrettheel, line 130: Some folks might want to do this if item size is a particular concern, should we remain a bit more neutral on whether this should be done?

docs/upgrading_binary.rst
```diff
@@ -0,0 +1,190 @@
...
+there might be significant cost and engineering complexity involved. If you choose to do so,
+follow the typical steps for data migration:
```
garrettheel, line 132: Might not hurt to include a brief warning to ensure a table backup exists before running any migrations?

garrettheel:
Couple of remaining thoughts:

1. Should we document that `serialize()` now retains values as bytes, since that method is public and can break anyone who tries to json encode the output? I know we don't have release notes in here yet but it might be helpful to add some while we're discussing everything

2. Regarding below, I wonder whether there are any changes we should implement now to better warn users about these kinds of issues. For example we could split out `to_dynamodb_json` and `to_simple_json` (or similar) and mark `to_json` as deprecated so that we can remove it in a future release. I'm OK to defer this decision until later if you want, since the release is already getting quite large.

> If designing from scratch, I'd expect
> 
> * one method which returns a "JSON" and raises if sets or binaries are present
> * another method which returns a "DynamoDB JSON" (with binaries as base64 strs, such that json can serialize them)

docs/upgrading_binary.rst
ikonst: Good point. I'll use your phrasing here.

docs/upgrading_binary.rst
ikonst: b4600a6d3b90429ddb77a7060d993b8602f40561

ikonst:
I like `to_dynamodb_json` and `to_simple_json`. One caveat is that current API is `to_dict` and `to_json`.

I find that `to_json` is almost never what you want, since it's common that you'd want to use the pre-serialized dict as part of something bigger (e.g. as part of something you're caching in redis, as a list of items...). Would it be OK to name it `to_dynamodb_json` but have it return a dict?

Also, initially I thought:
```python
def to_dynamodb_json() -> str:
  return json.dumps(self.serialize(), cls=EncoderWhichBaes64sBytes)
```
but it's no good for the "use it in something larger" case.

ikonst:
Here's a thought. Given a JSON encoder:
```python
import base64
import json
from typing import Any


class CustomEncoder(json.JSONEncoder):
    def default(self, obj: Any) -> Any:
        if isinstance(obj, bytes):
            return base64.b64encode(obj).decode()
        return super().default(obj)
```
you can do:
```python
class MyModel(Model):
   ...
   picture = BinaryAttribute(legacy_encoding=False)

data = {"items": [my_model.serialize()]}
data_str = json.dumps(data)
```

Now, how do you deserialize that? You can do
```python
data = json.loads(data_str)
items = [
   MyModel.from_raw_data(item)
   for item in data['items']
]
```
but that's broken since `BinaryAttribute(legacy_encoding=False)` will just pass through the `str value.

What if we change
```diff
 class BinaryAttribute
     def deserialize(self, value):
-        if self.legacy_encoding:
+        if self.legacy_encoding or isinstance(value, str):
             return b64decode(value)
         return value
```
and then perhaps get rid of `_handle_binary_attributes`, which you suspect can slow things down?

docs/release_notes.rst
```diff
@@ -6,6 +6,14 @@ Unreleased
 
 This is a major release and contains breaking changes. Please read the notes below carefully.
 
+* :py:class:`~pynamodb.attributes.BinaryAttribute` and :py:class:`~pynamodb.attributes.BinarySetAttribute` have undergone breaking changes:
+
+  * The attributes' internal encoding has changed. To prevent this change going unnoticed, :code:`legacy_encoding` have been made required: see :doc:`upgrading_binary` for details.
+    If your codebase uses :py:class:`~pynamodb.attributes.BinaryAttribute` or :py:class:`~pynamodb.attributes.BinarySetAttribute`,
+    go over the attribute declarations and mark them accordingly.
+  * When using binary attributes, the return value of :py:func:`~pynamodb.model.Model.serialize` will no longer be JSON-serializable
+    since it will contain :code:`bytes` objects.
```
garrettheel, line 15:
```suggestion
  * When using binary attributes, the return value of :py:func:`~pynamodb.models.Model.serialize` will no longer be JSON-serializable
    since it will contain :code:`bytes` objects. Note that both `:py:func:`~pynamodb.models.Model.to_dict` and `:py:func:`~pynamodb.model.Model.to_json` are also affected.
```

Should we also include a suggestion to remediate?

garrettheel (approved):

docs/release_notes.rst
ikonst: I'd like to implement those remediations first, won't release 6.0 without it.

PR #1113, opened 2022-11-16
Fix exception messages

pauliokas:
Pull request #786 broke messages for all errors that extend `PynamoDBException`, because it would overwrite the default message with `None`. This change returns the previous behaviour.


ikonst (approved):
Thank you!

PR #1114, opened 2022-11-17
Connection: call describe_table as needed

ikonst:
In #1095 we stopped calling DescribeTable for models since they can provide their own "meta-table". However, some users are using Connection directly and this was a breaking change for them, so we'll return to calling DescribeTable as needed in those cases.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -578,10 +578,9 @@ class Connection(object):
         """
         Returns information about the table's schema.
         """
-        try:
-            return self._tables[table_name]
-        except KeyError:
-            raise TableError(f"Meta-table for '{table_name}' not initialized") from None
+        if table_name not in self._tables:
+            self.describe_table(table_name)
+        return self._tables[table_name]
 
     def create_table(
         self,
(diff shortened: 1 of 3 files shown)


ikonst:
👋 @genhernandez

pynamodb/connection/base.py
```diff
@@ -578,10 +578,9 @@ def get_meta_table(self, table_name: str) -> MetaTable:
         """
         Returns information about the table's schema.
         """
-        try:
-            return self._tables[table_name]
-        except KeyError:
-            raise TableError(f"Meta-table for '{table_name}' not initialized") from None
+        if table_name not in self._tables:
+            self.describe_table(table_name)
```
anandswaminathan, line 582: @ikonst Don't you want to do add_meta_table here after the describe call ?

anandswaminathan: Oh ignore. This is self.describe_table. 

Looks good,

anandswaminathan (approved):

pynamodb/connection/base.py
ikonst: Yeah, `describe_table` populates the `_tables` with what it learns.

PR #1126, opened 2022-12-06
Replace to_dict with to_simple_dict, to_dynamodb_dict

ikonst:
Since #1112, `Model.serialize` return value is not necessarily JSON-serializable. We're adding `to_dynamodb_dict` as a replacement (basically, `bytes` are base-64 encoded), and `to_simple_dict` to provide for a common ask of a "simple" JSON representation similar to what the AWS Console defaults to (previously called `to_dict` in earlier versions of the 6.x branch).

We're making those methods of `AttributeContainer` rather than `Model` so they'd be equally applicable to `MapAttribute`.


pynamodb/attributes.py
```diff
@@ -478,6 +482,44 @@ def _instantiate(cls: Type[_ACT], attribute_values: Dict[str, Dict[str, Any]]) -
...
+        Returns the contents of this instance as a normal JSON-serializable mapping,
+        using the same format as the "normal" JSON mapping in the AWS Console.
+
+        :param force: If :code:`True`, force the conversion even if the model contains Binary or Set attributes
+          (as they cannot be unambiguously represented in JSON, so this assumes the same model will be used
+          when parsing). If :code:`False`, a :code:`ValueError` will be raised if such attributes are set.
+        """
+        return {k: attr_value_to_normal_dict(v, force) for k, v in self._container_serialize(null_check=False).items()}
+
+    def from_normal_dict(self, d: Dict[str, Any]) -> None:
```
garrettheel, line 515: to nit a little bit on the name, I wonder if the term "normal" might be a little ambiguous.. but I also don't necessarily have a better suggestion that is not horribly verbose. "simple" is maybe the only term that comes to mind, since it implies that there may be something unsupported

pynamodb/attributes.py
```diff
@@ -478,6 +482,44 @@ def _instantiate(cls: Type[_ACT], attribute_values: Dict[str, Dict[str, Any]]) -
         AttributeContainer._container_deserialize(instance, attribute_values)
         return instance
 
+    def to_dynamodb_dict(self) -> Dict[str, Any]:
+        """
+        Returns the contents of this instance as a JSON-serializable mapping,
```
garrettheel, line 487: What do you think about providing a really simple example of the output here to show the structure? Seems like a lot of people might trip up over the diff between dynamodb/normal. Would it also help to type as `dict[str, dict[str, Any]]`?

garrettheel (approved):
Couple comments on naming/docs but nothing blocking

pynamodb/attributes.py
ikonst: I used "simple" before. Can switch back.

ikonst: 0dc40c1795028763617785e455ecdb9a3669d8c2

garrettheel:
:shipit:

PR #1156, opened 2023-01-26
Expose transaction cancellation reasons

### LINKED ISSUE #891: CancellationReasons are not included in the VerboseClientError

ikonst:
This will allow the caller to know e.g. which of the transaction items ran into a conflict, which can sometimes allow app logic to better handle the error without needing to make more requests.

This is a backport of #1144 to 5.x branch (adjusted to Python 3.6 compatibility, hence no dataclass).

Fixes #891.

### DIFF
--- a/pynamodb/connection/base.py
+++ b/pynamodb/connection/base.py
@@ -48,7 +48,8 @@ from pynamodb.constants import (
 from pynamodb.exceptions import (
     TableError, QueryError, PutError, DeleteError, UpdateError, GetError, ScanError, TableDoesNotExist,
     VerboseClientError,
-    TransactGetError, TransactWriteError)
+    TransactGetError, TransactWriteError, CancellationReason,
+)
 from pynamodb.expressions.condition import Condition
 from pynamodb.expressions.operand import Path
 from pynamodb.expressions.projection import create_projection_expression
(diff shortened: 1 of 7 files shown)


anandswaminathan (approved):

pynamodb/connection/base.py
```diff
@@ -465,7 +466,20 @@ def _make_api_call(self, operation_name: str, operation_kwargs: Dict, settings:
...
+                    raise VerboseClientError(
+                        botocore_expected_format,
+                        operation_name,
+                        verbose_properties,
+                        cancellation_reasons=(
+                            (
+                                CancellationReason(
+                                    code=d['Code'],
+                                    message=d.get('Message'),
+                                ) if d['Code'] != 'None' else None
```
hoIIer, line 478: Why return `None` here? Shouldn't we instead not return anything at all?

The following code will break with `AttributeError: 'NoneType' object has no attribute 'code'`:
```
        except pydb_exc.TransactWriteError as e:
            retry_err_types = (
                'ProvisionedThroughputExceededException',
                'ThrottlingError',
                'TransactionConflict',
            )
            for error in e.cancellation_reasons:
                if error.code in retry_err_types:
                    raise RuntimeError(f'transaction failed {error.code}')
```
Just seems like it'd be cleaner to not return anything vs returning `None`

ikonst: The cancellation reasons are 1:1 to the transaction items so you can know what item was the reason.

Would you rather the reason itself not to be none but for code to be none?

hoIIer: Yeah to me that would feel cleaner but I guess it's subjective. Just negates the extra "if error and error.code"

ikonst: The thing is, if it was "cleaner", it would leave  users with no way of associating reasons with individual transaction items. Maybe a different interface would convey this better, e.g. mapping of item to reason.

hoIIer: yeah that would make sense to me

PR #1163, opened 2023-02-19
Update DATETIME_FORMAT to clarify it must be UTC

### LINKED ISSUE #1139: Incostistent log and documentation.
https://github.com/pynamodb/PynamoDB/blob/b3330a43ee8f9c203a32834f080dde42ab4cf8c7/pynamodb/attributes.py#L886

expected format is %Y-%m-%dT%H:%M:%S.%f+0000 (seen here https://github.com/pynamodb/PynamoDB/blob/b3330a43ee8f9c203a32834f080dde42ab4cf8c7/pynamodb/attributes.py#L882-L886 )
but logs and documentation show that required format is %Y-%m-%dT%H:%M:%S.%fZ

ikonst:
The DATETIME_FORMAT is
1. used as a format string for `strftime` but only for UTC datetimes
2. interpolated into error messages

While the `%z` didn't matter for (1) as it was always `"+0000"`, for (2) it suggested that `UTCDateTimeAttribute` can read serialized values with different timezone offsets, which is not the case.

Fixes #1139.


ikonst:
@anandswaminathan the serialization code is `value.astimezone(timezone.utc).strftime(DATETIME_FORMAT)` so I think `%z` would always become `"+0000"`

anandswaminathan (approved):

PR #1172, opened 2023-04-18
Add 'add_version_condition' arg

### LINKED ISSUE #1168: Model.Update should not inject version condition
# The problem

The `VersionAttribute` design pattern consists of
- **incrementing the version with every change** — to trip others so they don't overwrite your change
- **conditioning on the version** — to trip yourself so you don't overwrite others' change

There are two methods that can modify an existing model:
- `Model.save`
- `Model.update`

We've observed it to be a common pattern to change just a few fields and then call `Model.save`. In this case, a version condition is a good precaution to avoid overwriting more attributes than you meant to. Once users start seeing high % of conditional failures ("conflicts"), a more nuanced approach is called for, using `Model.update`. Often "last write wins" is good enough as long as you don't write unrelated attributes. Other times, a more nuanced condition can be used (e.g. update field X only if it had value Y before).

Unfortunately when we introduced `VersionAttribute`, we’ve made `Model.update` [append a condition](https://github.com/pynamodb/PynamoDB/blob/b5600ea8c44dca523f94a0d1c8491d7ee0a8b3f5/pynamodb/models.py#L425-L427) on the version attribute. This is probably counter-productive to what I described above.

# The proposal

In PynamoDB 6.0, we can allow changing this behavior.

Options:
1. Stop appending the condition in `Model.update`. This risks silently breaking existing apps.
2. Make it optional and default to append. This drags the bad behavior as a default into green-field apps.
3. Make it optional and default to not appending. This risks silently breaking existing apps.
4. Ask user to decide, provide no default. 📌 

I'm inclined towards option (4). Adding a parameter to `Model.update` is disruptive, and also risks adding a hidden crash, since not every `Model.update` call is necessarily exercised by tests. Instead, we can add a parameter to `VersionAttribute`, since it's executed at class load / module load, and likely prevent an app from starting up until its maintainer makes a conscious decision.

We should name it something like `legacy_update_behavior: bool` as to discourage setting it to `True` in green-field apps:
```diff
-version = VersionAttribute()
+version = VersionAttribute(legacy_update_behavior=False)
```

In future versions we could switch to Option 1 by removing it.

ikonst:
Add a flag for controlling whether `Model.save`, `Model.update` and `Model.delete` add a condition that the persisted version is the same as the in-memory one, defaulting to `True` (the "safe" behavior)

Fixes #1168.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -772,7 +772,36 @@ class NumberSetAttribute(Attribute[Set[float]]):
 
 class VersionAttribute(NumberAttribute):
     """
-    A version attribute
+    A number attribute that denotes a version / revision counter of the model.
+
+    This attribute has special semantics within the model:
+
+    * :meth:`~pynamodb.models.Model.save` and :meth:`~pynamodb.models.Model.update` would increment the counter
+      every time the model is persisted. This allows concurrent updates not to overwrite each other, at the expense
+      of the latter update failing.
+    * :meth:`~pynamodb.models.Model.save`, :meth:`~pynamodb.models.Model.update`
+      and :meth:`~pynamodb.models.Model.delete` would avoid acting on a model if the counter in the database
+      is not what they expect to find, at the expense of failing. For `update` and `delete`, this behavior is optional
+      since a more granular approach can be preferred.
+
+    For example, suppose a room should be first reserved and then rented, and users can cancel
+    their reservation. Naive code would check to ensure the room is reserved, and then persist it with `rented=True`.
+    If the user has meanwhile cancelled their reservation, the version check would fail, allowing the rent code
+    to realize that its check was under outdated assumptions. (A retry would allow the 'rent' function to indicate
+    a high-level error to the user, e.g. "You must reserve before renting".)
+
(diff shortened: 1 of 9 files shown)


garrettheel (approved):
Looks like a reasonable way to allow for granular overrides. Should we just add this to `save` for consistency as well? Given that there may sometimes be a use-case to totally disregard the version check there too

pynamodb/attributes.py
```diff
@@ -772,7 +772,36 @@ def deserialize(self, value):
 
 class VersionAttribute(NumberAttribute):
     """
-    A version attribute
+    A number attribute that denotes a version / revision counter of the model.
```
ikonst, line 775: TODO simply say it's an optimistic locking mechanism and move the prose to a separate page

ikonst:
@garrettheel What you reviewed before was totally broken -- when `False` it didn't add condition **nor** incremented version 😅 

This time
- fixed the thing
- updated tests
- removed the prose from VersionAttribute, instead linking to an "optimistic locking" page that we already had, and adding a section to it about skipping the condition

PR #1177, opened 2023-04-21
Add 'add_version_condition' arg

ikonst:
Add a flag for controlling whether `Model.save`, `Model.update` and `Model.delete` add a condition that the persisted version is the same as the in-memory one, defaulting to `True` (the "safe" behavior)

This is a backport to 5.x.


ikonst:
@garrettheel since it's additive see no reason not to merge into 5.x?

garrettheel (approved):

PR #1189, opened 2023-07-12
Fix updating set to empty

### LINKED ISSUE #962: Setting <UnicodeSetAttribute>.set([]) (empty list) throws "Nonetype is not iterable" error.
During an update operation a UnicodeSet field may have it's value changed back to empty e.g. []

The following code fails:
```
(Pdb) actions = []
(Pdb) actions.append(attr.set([]))
(Pdb) record.update(actions)
```
`*** TypeError: 'NoneType' object is not iterable`

The following two examples work:
```
(Pdb) actions = []
(Pdb) actions.append(attr.set(None))
(Pdb) record.update(actions)
```
```
(Pdb) actions = []
(Pdb) actions.append(attr.set(['foo', 'bar']))
(Pdb) record.update(actions)
```

Is this a bug?

ikonst:
Fixes #962.

Empty sets are not supported by DynamoDB. Currently we're representing empty sets by unsetting the attribute (hence it's indistinguishable from unsetting), i.e.
```python
m1.my_set = set()
m1.save()
m2.my_set = None
m2.save()
```
then
```python
m1.refresh()
m2.refresh()
assert m1.my_set == m2.my_set == None
```

Another way to empty the set should be through an update expression:
```python
m1.update([MyModel.my_set.set([]))
```

We're not changing this semantic here, but we're making it so that `m1.update([MyModel.my_set.set([]))` doesn't crash in botocore.

When preparing an update expression, we've been sending update expressions like `{':0': {'NS': None}}` which crashed botocore with "TypeError: 'NoneType' object is not iterable". The PR addresses it by using a `REMOVE` action when assigning an empty value to a set.

### DIFF
--- a/pynamodb/expressions/operand.py
+++ b/pynamodb/expressions/operand.py
@@ -299,7 +299,9 @@ class Path(_NumericOperand, _ListAppendOperand, _ConditionOperand):
         return _IfNotExists(self, self._to_operand(other))
 
     def set(self, value: Any) -> Union[SetAction, RemoveAction]:
-        # Returns an update action that sets this attribute to the given value
+        # Returns an update action that sets this attribute to the given value.
+        # For attributes that may not be empty (e.g. sets), this may result
+        # in a remove action.
         operand = self._to_operand(value)
         if isinstance(operand, Value) and next(iter(operand.value.values())) is None:
             return RemoveAction(self)
--- a/pynamodb/expressions/operand.py
+++ b/pynamodb/expressions/operand.py
@@ -298,9 +298,12 @@ class Path(_NumericOperand, _ListAppendOperand, _ConditionOperand):
     def __or__(self, other):
         return _IfNotExists(self, self._to_operand(other))
 
-    def set(self, value: Any) -> SetAction:
+    def set(self, value: Any) -> Union[SetAction, RemoveAction]:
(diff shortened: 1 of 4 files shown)


ikonst:
@jmphilli are you still dynamo-ing?

anandswaminathan:
@ikonst Reviewing

pynamodb/expressions/operand.py
```diff
@@ -298,9 +298,12 @@ def __getitem__(self, item: Union[int, str]) -> 'Path':
     def __or__(self, other):
         return _IfNotExists(self, self._to_operand(other))
 
-    def set(self, value: Any) -> SetAction:
+    def set(self, value: Any) -> Union[SetAction, RemoveAction]:
         # Returns an update action that sets this attribute to the given value
```
anandswaminathan, line 302: Comment: "Sets/removes" attribute

anandswaminathan (approved):

pynamodb/expressions/operand.py
ikonst, line 302:
```suggestion
        # Returns an update action that sets this attribute to the given value.
        # For attributes that may not be empty (e.g. sets), this may result
        # in a remove action.
```

PR #1201, opened 2023-10-25
feat: Model Attribute Transformation

neilramsay:
PynamoDB sometimes need to connect to a DynamoDB Table that is not managed by PynamoDB, and may have different naming conventions, such as Camel Case (which is used by JSON).

PynamoDB can be configured to override individual attribute names, however in practice every attribute needs to be overridden to map between naming conventions.

This commit introduces a Model Attribute Transformation, which effectively converts python snake case to another naming convention without overriding every attribute.
This is an opt-in feature, and does not cause any breaking change to existing functionality.


neilramsay:
Please note:
- I have not built the documentation, so there may need to be some formatting updates
- I'm not 100% clear on how to format the type hints for long function definitions (and have avoided using my inbuilt 'black' formatter).

PR #1204, opened 2023-11-04
Fixed filter_condition examples in the docs

Ali-Toosi:
The current examples don't work with DynamoDB because they are using `email` in the range key condition.
It also doesn't seem like we can combine query terms for range conditions so I changed the wording to specify this is about filter conditions.

No code changes made.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -136,11 +136,18 @@ Now, suppose that you want to search the table for users with a last name
     for user in UserModel.query('Smith', UserModel.first_name.startswith('J')):
         print(user.first_name)
 
-You can combine query terms:
+You can combine query terms in filter conditions using OR:
 
 ::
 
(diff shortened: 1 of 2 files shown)


docs/quickstart.rst
```diff
@@ -136,11 +136,18 @@ Now, suppose that you want to search the table for users with a last name
     for user in UserModel.query('Smith', UserModel.first_name.startswith('J')):
         print(user.first_name)
 
-You can combine query terms:
+You can combine query terms in filter conditions using OR:
```
Aareon, line 139: I would clarify by saying
> You can combine query terms in filter conditions using logical operators (ex, `|`, `&`, `==`, etc.)

Ali-Toosi: Thanks for the review.
I have given a separate example for `&` just below this one. Not sure how `==` would work here for combining queries 🤔 
Let me know if there are other logical operators that are supported here. I'll either add them in-line or add a separate example if they make more sense that way.

PR #1208, opened 2023-12-13
Enable tests for Python 3.11 and 3.12

kmkhr:


vanlenhat:
I use Python 3.11.7 and I think is a bug: https://discord.com/channels/1131108036010836000/1212064770211975209/1217053517731201024

jesusch:
guess 3.7 (EOL) should be dropped?

Issue #1225, opened 2024-02-10
CancellationReason missing Item


knap1930:
When sending "ReturnValuesOnConditionCheckFailure": "ALL_OLD" for a transaction, there is no way to access the item that failed the condition check. here is an example response copied from https://github.com/getmoto/moto/issues/5204

```
{'CancellationReasons': [{'Code': 'ConditionalCheckFailed',
                          'Item': {'customer': {'S': 'mark'},
                                   'lock': {'M': {'acquired_at': {'N': '123'}}},
                                   'mealtime': {'S': 'breakfast'}},
                          'Message': 'The conditional request failed'}],
 'Error': {'Code': 'TransactionCanceledException',
           'Message': 'Transaction cancelled, please refer cancellation '
                      'reasons for specific reasons [ConditionalCheckFailed]'},
 'Message': 'Transaction cancelled, please refer cancellation reasons for '
            'specific reasons [ConditionalCheckFailed]',
 'ResponseMetadata': {'HTTPHeaders': {'connection': 'keep-alive',
                                      'content-length': '393',
                                      'content-type': 'application/x-amz-json-1.0',
                                      'date': 'Wed, 08 Jun 2022 18:55:51 GMT',
                                      'server': 'Server',
                                      'x-amz-crc32': '947576560',
                                      'x-amzn-requestid': 'BDBL292TOK52OSPP0GTJTTS2CVVV4KQNSO5AEMVJF66Q9ASUAAJG'},
                      'HTTPStatusCode': 400,
                      'RequestId': 'BDBL292TOK52OSPP0GTJTTS2CVVV4KQNSO5AEMVJF66Q9ASUAAJG',
                      'RetryAttempts': 0}}
```

PR #1226, opened 2024-02-10
return Item in cancellation_reasons when transaction fails and return_values=ALL_OLD

knap1930:
When sending "ReturnValuesOnConditionCheckFailure": "ALL_OLD" the item is returned within the CancellationReasons list.

https://github.com/pynamodb/PynamoDB/issues/1225

### DIFF
--- a/pynamodb/exceptions.py
+++ b/pynamodb/exceptions.py
@@ -134,6 +134,7 @@ class CancellationReason:
     """
     code: str
     message: Optional[str] = None
+    item: Optional[Dict[str, Any]] = None
 
 
 class TransactWriteError(PynamoDBException):
--- a/pynamodb/exceptions.py
+++ b/pynamodb/exceptions.py
@@ -134,7 +134,7 @@ class CancellationReason:
     """
     code: str
     message: Optional[str] = None
-    item: Optional[Dict[str, Any]] = None
+    item: Optional[Dict[str, Dict[str, Any]]] = None
 
 
 class TransactWriteError(PynamoDBException):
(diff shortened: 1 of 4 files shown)


pynamodb/exceptions.py
```diff
@@ -134,6 +134,7 @@ class CancellationReason:
     """
     code: str
     message: Optional[str] = None
+    item: Optional[Dict[str, Dict[str, Any]]] = None
```
ikonst, line 137: Maybe we should call it `raw_item` given that typically in PynamoDB we also attempt to map the item?

knap1930: makes sense

tests/integration/test_transaction_integration.py
```diff
@@ -168,6 +169,24 @@ def test_transact_write__error__transaction_cancelled__condition_check_failure(c
...
+    with pytest.raises(TransactWriteError) as exc_info:
+        with TransactWrite(connection=connection) as transaction:
+            transaction.save(User(1), condition=(User.user_id.does_not_exist()), return_values=ALL_OLD)
+    assert exc_info.value.cause_response_code == TRANSACTION_CANCELLED
+    assert 'ConditionalCheckFailed' in exc_info.value.cause_response_message
+    assert exc_info.value.cancellation_reasons == [
+        CancellationReason(code='ConditionalCheckFailed', message='The conditional request failed', raw_item=User(1).to_dynamodb_dict()),
+    ]
+    assert isinstance(exc_info.value.cause, botocore.exceptions.ClientError)
+    assert User.Meta.table_name in exc_info.value.cause.MSG_TEMPLATE
```
ikonst, line 187:
```suggestion
```
I think we already test it elsewhere?

ikonst (approved):

PR #1267, opened 2025-05-13
chore(docs): remove default region reference in docs

ryancausey:
This removes the reference to the default region being us-east-1 if not specified in the Model's Meta class. This is not longer true as of version 6.0.0.


ikonst (approved):

ikonst:
Looks like tests are currently broken. Perhaps another botocore change we need to hack around.

ryancausey:
Is there anything I need to do in order to fix the tests here, or is that going to be part of a different change?

ikonst:
Let's try and fix this separately just for clarity.

PR #1270, opened 2025-05-19
feat: add wait argument to `Model.delete_table`

ryancausey:
This adds a `wait` argument that will make the `delete_table` call block until the DynamoDB table is fully deleted if set to `True`. It defaults to `False` to not break the existing behavior.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -3438,3 +3438,21 @@ def test_delete(add_version_condition: bool) -> None:
         }
         args = req.call_args[0][1]
         assert args == expected
+
+def test_delete_with_wait(mocker):
+    """Test that the wait argument works as expected on the delete."""
+    mock_exists = mocker.patch.object(target=Model, attribute="exists", autospec=True)
+    mock__get_connection = mocker.patch.object(
+        target=Model, attribute="_get_connection", autospec=True
+    )
+
+    # Make the first two exists calls show the table as still existing, then have the
+    # last call show it has been deleted.
+    mock_exists.side_effect = (True, True, False)
+
+    result = Model.delete_table(wait=True)
+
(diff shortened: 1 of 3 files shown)


ryancausey:
@ikonst do I need to open an issue for this PR or is this good enough to be reviewed on its own? I was trying to integrate PynamoDB models into FactoryBoy and mimic the DjangoORM behavoir of dropping and re-creating the db on each test, but found the `wait` argument missing from the `delete_table` call.

ikonst:
No need for an issue ;) I'll get to it soon!

tests/test_model.py
```diff
@@ -3438,3 +3438,18 @@ def test_delete(add_version_condition: bool) -> None:
         }
         args = req.call_args[0][1]
         assert args == expected
+
+@patch.object(target=Model, attribute="exists", autospec=True)
+@patch.object(target=Model, attribute="_get_connection", autospec=True)
+def test_delete_with_wait(mock__get_connection, mock_exists):
```
ikonst, line 3444: I find [mocker](https://pytest-mock.readthedocs.io/en/latest/usage.html) to be more readable than `@patch` decorators, since you don't need to match arguments in reverse.

```suggestion
def test_delete_with_wait(mocker):
    mock_exists = mocker.patch.object(target=Model, attribute="exists", autospec=True)
    mock__get_connection = mocker.patch.object(target=Model, attribute="_get_connection", autospec=True)
```

ryancausey: TIL

ikonst:
Can you also update docs/release_notes.rst?

ryancausey:
@ikonst release notes added.

ikonst (approved):

Issue #1287, opened 2025-09-09
PynamoDB's Model.get operation throws an incorrect exception when getting an object with only single optional attribute_to_get


domderen:
Hey, 

I just ran into an issue while using Model.get API. I called it with a single attribute value in `attributes_to_get` parameter. This attribute in my model is an optional value, that might not exist on my DynamoDB object. So when the underlying get_item call is made, it returns a response where `Item` field is an empty dict (`{}`). 

So then [this](https://github.com/pynamodb/PynamoDB/blob/2f8e8bb60474babd6ad558a91b0ad89a2c36d515/pynamodb/models.py#L553) if case is executed, which evaulates to false, and because of that, a `DoesNotExist` exception is thrown. 

I would expect this API to return an empty object, or at least throw a different kind of exception. Though in my case, such a query is completely valid, so I would prefer an empty object response. I worked around it by requesting the `attributes_to_get` parameter with two values, my hash key attribute, and my optional attribute. That way I always get some value to deserialize into class instance.

What do you think? Do you think this is something that should be modified in PynamoDB?

A small sample to present what I mean

```python
from pynamodb.attributes import (
    UnicodeAttribute,
)
from pynamodb.models import Model


class SomeModel(Model):

    id = UnicodeAttribute(hash_key=True)
    optional_attr = UnicodeAttribute(null=True, default=None)

def get_optional_attr():
    try:
        # this throws an exception even if instance with hash_key 1 exists, but it doesn't contain a value for optional_attr.
        return SomeModel.get("1", attr_name=["optional_attr"]).optional_attr
    except Exception as e:
        print(e)
        return None

def get_optional_attr_with_default():
    # This works fine, returning a None for optional_attr, but it also obtains the "id" attribute from DynamoDB.
    return SomeModel.get("1", attr_name=["id", "optional_attr"]).optional_attr
```

ikonst:
Agree, I'd rather have something like:
```diff
         if data:
             item_data = data.get(ITEM)
-            if item_data:
+            if item_data is not None:
                 return cls.from_raw_data(item_data)
```
or maybe more precise
```diff
         if data:
-            item_data = data.get(ITEM)
-            if item_data:
-                return cls.from_raw_data(item_data)
+            if ITEM in data:
+                return cls.from_raw_data(item_data[ITEM])
```

domderen:
Yeah either of those would be great. I can create a PR if you want?

ikonst:
Sure!

[...]

PR #1289, opened 2025-09-16
Fix get method in models.py to not throw an exception when returning an empty item.

domderen:
Fix for issue https://github.com/pynamodb/PynamoDB/issues/1287#issuecomment-3281642349 When getting a model value with `attributes_to_get` parameter, set in a way where it could return an empty object, the get method is erronously throwing an exception even though the object exists. This fixes the issue.

### DIFF
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -549,9 +549,8 @@ class Model(AttributeContainer, metaclass=MetaModel):
             attributes_to_get=attributes_to_get,
         )
         if data:
-            item_data = data.get(ITEM)
-            if item_data:
-                return cls.from_raw_data(item_data)
+            if ITEM in data:
+                return cls.from_raw_data(data.get(ITEM))
         raise cls.DoesNotExist()
 
     @classmethod
--- a/pynamodb/models.py
+++ b/pynamodb/models.py
@@ -548,9 +548,8 @@ class Model(AttributeContainer, metaclass=MetaModel):
             consistent_read=consistent_read,
             attributes_to_get=attributes_to_get,
         )
-        if data:
-            if ITEM in data:
-                return cls.from_raw_data(data.get(ITEM))
+        if data and ITEM in data:
+            return cls.from_raw_data(data.get(ITEM))
         raise cls.DoesNotExist()
 
     @classmethod
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1766,7 +1766,7 @@ class ModelTestCase(TestCase):
                 # This simulates DynamoDB returning an empty Item when the requested
                 # attributes don't exist or are empty, but the item itself exists
                 return {
-                    'Item': {}  # Empty item dictionary
+                    'Item': {}
                 }
             return MODEL_TABLE_DATA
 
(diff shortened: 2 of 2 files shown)


pynamodb/models.py
```diff
@@ -549,9 +549,8 @@ def get(
             attributes_to_get=attributes_to_get,
         )
         if data:
-            item_data = data.get(ITEM)
-            if item_data:
-                return cls.from_raw_data(item_data)
+            if ITEM in data:
```
ikonst: probably `if data and ITEM in data:`

ikonst:
Are you going to add a test?

domderen:
Hey, let me know whatt you think :)

tests/test_model.py
```diff
@@ -1752,6 +1752,68 @@ def fake_dynamodb(*args):
...
+        
+        def fake_dynamodb(*args):
+            kwargs = args[1]
+            if kwargs == {'TableName': UserModel.Meta.table_name}:
+                return MODEL_TABLE_DATA
+            elif 'Key' in kwargs and 'ProjectionExpression' in kwargs:
+                # This simulates DynamoDB returning an empty Item when the requested
+                # attributes don't exist or are empty, but the item itself exists
+                return {
+                    'Item': {}  # Empty item dictionary
```
ikonst: Your comments literally dub the code
```suggestion
                    'Item': {}
```

tests/test_model.py
```diff
@@ -1752,6 +1752,68 @@ def fake_dynamodb(*args):
...
+                }
+            return MODEL_TABLE_DATA
+
+        fake_db = MagicMock()
+        fake_db.side_effect = fake_dynamodb
+
+        with patch(PATCH_METHOD, new=fake_db) as req:
+            # This should not raise DoesNotExist even though the Item is empty
+            # because the item exists in DynamoDB, just the requested attributes are empty
```
ikonst:
```suggestion
```
You're repeating the test's docstring

tests/test_model.py
```diff
@@ -1752,6 +1752,68 @@ def fake_dynamodb(*args):
...
+                )
+                # The item should be returned with empty attributes
+                self.assertIsNotNone(item)
+                # Verify that the get method was called with the right parameters
+                self.assertTrue(req.called)
+                call_args = req.call_args[0][1]
+                self.assertIn('Key', call_args)
+                self.assertIn('ProjectionExpression', call_args)
+            except DoesNotExist:
+                self.fail("get method should not raise DoesNotExist when DynamoDB returns empty Item but item exists")
```
ikonst: No need to raise/except since an unexpected exception will fail the test anyway.

tests/test_model.py
```diff
@@ -1752,6 +1752,68 @@ def fake_dynamodb(*args):
...
+                    'bar',
+                    attributes_to_get=['nonexistent_attribute']
+                )
+                # The item should be returned with empty attributes
+                self.assertIsNotNone(item)
+                # Verify that the get method was called with the right parameters
+                self.assertTrue(req.called)
+                call_args = req.call_args[0][1]
+                self.assertIn('Key', call_args)
+                self.assertIn('ProjectionExpression', call_args)
```
ikonst: I don't think we need to assert on implementation details. This just makes for a brittle test.

tests/test_model.py
```diff
@@ -1752,6 +1752,68 @@ def fake_dynamodb(*args):
...
+            # This should not raise DoesNotExist even though the Item is empty
+            # because the item exists in DynamoDB, just the requested attributes are empty
+            try:
+                item = UserModel.get(
+                    'foo',
+                    'bar',
+                    attributes_to_get=['nonexistent_attribute']
+                )
+                # The item should be returned with empty attributes
+                self.assertIsNotNone(item)
```
ikonst: Let's make a stronger assertion here than just `item is not None`.
```suggestion
                assert item is not None
                assert item.user_name is None
```

tests/test_model.py
```diff
@@ -1752,6 +1752,68 @@ def fake_dynamodb(*args):
...
+                # Verify that the get method was called with the right parameters
+                self.assertTrue(req.called)
+                call_args = req.call_args[0][1]
+                self.assertIn('Key', call_args)
+                self.assertIn('ProjectionExpression', call_args)
+            except DoesNotExist:
+                self.fail("get method should not raise DoesNotExist when DynamoDB returns empty Item but item exists")
+    
+    def test_get_with_no_item_still_raises_does_not_exist(self):
```
ikonst: Do we have no test for this already?

domderen:
Hey, I updated the comments, and I removed the second test as indeed there were other places that already check if this exception is thrown when `Item` is not in the response. 

In regards to better asserts, I'm now checking if the property we are trying to get indeed does not exist, even though the object is of correct type. Is this something more in line with your thinking?

domderen:
Hey, any update on that?

PR #1294, opened 2025-11-27
feat: attr_type argument to Attribute.is_type

KO1231:
This PR adds `attr_type` to make it possible to check the type of an `Attribute` in DynamoDB within a [Conditional Expression](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#:~:text=Checking%20for%20attribute%20type).

It defaults to `None` to not break the existing behavior.

### DIFF
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -233,9 +233,8 @@ class Attribute(Generic[_T]):
     def does_not_exist(self) -> 'NotExists':
         return Path(self).does_not_exist()
 
-    def is_type(self):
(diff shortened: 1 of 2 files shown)


KO1231:
In the [AWS documentation example](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#:~:text=Checking%20for%20attribute%20type), it is used in the Conditional Expression when performing a delete-item operation, and it can also be used for null checks.

When checking to see if data is null, in addition to using attribute_not_exists, we must also verify whether the attribute is set to null by this.

PR #1302, opened 2026-04-23
fix: allow Enum values as attribute defaults

### LINKED ISSUE #1301: Enum not considered immutable for defaults.
In PynamoDB 4, I was allowed to do:

```
@unique
class ApplicationType(Enum):
    AWS = 'aws'
    AZURE = 'azure'

class ApplicationModel(Generic[M], Model):
    type = StringEnumAttribute[ApplicationType](attr_name=_m('type'), default=ApplicationType.AWS)
```

However, in PynamoDB 6, I get this error:
```
E   ValueError: An attribute's 'default' must be immutable (str, int, float, datetime, timedelta, bytes, bool, tuple, frozenset, NoneType) or a callable (see https://pynamodb.readthedocs.io/en/latest/api.html#pynamodb.attributes.Attribute)
```
Why can't I use this string (which is immutable) from an enum (which is also immutable)?  I realize that I can use `default = lambda: ApplicationType.AWS` to work around this, but that's less clear.

GonzalezSanchez:
Problem
In PynamoDB 6.0, using an Enum value as an attribute default raises a ValueError:

```python
class ApplicationType(Enum):
AWS = 'aws'
AZURE = 'azure'

Raises: ValueError: An attribute's 'default' must be immutable (str, int, ...) or a callable
type = StringEnumAttribute(default=ApplicationType.AWS)
```

The only workaround was to wrap it in a lambda:

```python
type = StringEnumAttribute(default=lambda: ApplicationType.AWS)
```

Fix
Enum instances are immutable by design — this is a Python guarantee. Added Enum to _IMMUTABLE_TYPES so all Enum subclass instances are accepted as defaults.

Changes
pynamodb/attributes.py: import Enum and add it to _IMMUTABLE_TYPES
tests/test_attributes.py: add test_default_enum to cover this case
Closes #1301

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -3,6 +3,13 @@
 Release Notes
 =============
 
+v6.1.1
+------
+
+Fixes:
+
+* Allow :py:class:`~enum.Enum` instances as attribute defaults (:pr:`1302`)
+
 v6.1.0
 ------
 
--- a/pynamodb/attributes.py
+++ b/pynamodb/attributes.py
@@ -7,6 +7,7 @@ import collections.abc
 import json
 import time
 import warnings
+from enum import Enum
 from base64 import b64encode, b64decode
 from copy import deepcopy
 from datetime import datetime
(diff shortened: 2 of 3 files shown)


ikonst (approved):

ikonst (commented):
Consider updating release_notes.rst

GonzalezSanchez:
Done! Added a release note under v6.1.1.

Issue #1293, opened 2025-11-26
Support for Multi Attribute Composite Key for GSIs


keremeyuboglu:
Related AWS announcement:

https://aws.amazon.com/about-aws/whats-new/2025/11/amazon-dynamodb-multi-attribute-composite-keys-global-secondary-indexes/

Here is the example doc page:

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.DesignPattern.MultiAttributeKeys.html

I wanted to create this issue to garner potential support as this looks like a great improvement on Dynamo itself. I don't know how it would be implemented in the current state of the code without breaking changes though.

PR #1303, opened 2026-05-01
feat: Support Composite Global Secondary Indexes

keremeyuboglu:
Adds support for composite Global Secondary Index keys, including multiple hash key and range key attributes declared on a GSI. The implementation preserves declaration order when building index schemas and query key conditions. 
See: #1293 

For the query API, I created a separate hash_keys parameter for composite partition keys instead of changing the existing hash_key parameter to also accept dictionaries, but unsure if it would have been better to merge this new logic onto the hash_key parameter instead.


keremeyuboglu:
I'd rather have a feedback on this as I would rather not use a deviated custom version of this library instead of the main one

